PR #34, opened 2020-05-31
[Feature] Improve Logger (closes #22)

### LINKED ISSUE #22: Improve Logger
Consider the following aspects:

1. The logger should not log anything in `APP_STORE` builds.
2. Remove unused code from `Logger.swift`.
3. Improve error logging: Make errors stand out so that they can easily be spotted during development.

Schlabbi:
The following changes have been made:

- The logger no longer provides functions like `logError` where the Log Level can then be overridden manually. Thus things like `logError(error, level: .info)` are no longer possible.
- As a replacement the logger provides wrapper functions for the various log levels. Example: `appLogger.info("App Started")`
- All log statements in the app have been migrated to these new logging functions
- The logger will no longer log to the console when the app is not in `DEBUG` mode.
- As a basis for the new logger the dependency [SwiftyBeaver](https://github.com/SwiftyBeaver/SwiftyBeaver) was added (in case this is not wished, SwiftyBeaver can be easily replaced using this PR as a basis)
- The logs are now also written into a file. This log data can be retrieved by calling `appLogger.getLoggedData()` (this has been manually tested since I cannot get past the Onboarding due to missing entitlements)
- SwiftyBeaver highlights the different log levels in the console by default. Thus errors can be more easily spotted during development.


src/xcode/ENA/ENA/Source/Workers/Logger.swift
```diff
@@ -16,27 +16,75 @@
...
+	
+	private let sb = SwiftyBeaver.self
+	
+	private let fileDest = FileDestination()
+	private let consoleDest = ConsoleDestination()
 
-	func getLoggedData() -> Data? {
-		Data()
+	private var logDirectory: URL? {
+		return try? FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
```
flashspys, line 31: I suggest to use `.cachesDirectory` instead of `.documentDirectory`

flashspys (changes requested):
Log should not appear in the documents folder. Afaik they will be synced via iCloud(?) which is really not needed.

src/xcode/ENA/ENA/Source/Workers/Logger.swift
Schlabbi: Yes, the log files are better placed in the caches directory. Just pushed a new commit to fix this.

src/xcode/ENA/ENA/Source/Workers/Logger.swift
```diff
@@ -16,27 +16,75 @@
...
+	func warning(message: String, _ file: String = #file, _ line: Int = #line, _ function: String = #function) {
+		log(level: .warning, message: message, file: file, line: line, function: function)
+	}
+	
+	func error(message: String, _ file: String = #file, _ line: Int = #line, _ function: String = #function) {
+		log(level: .error, message: message, file: file, line: line, function: function)
+	}
+	
+	private func log(level: SwiftyBeaver.Level, message: String, file: String, line: Int, function: String) {
+		#if DEBUG
```
inf2381, line 71: Could you please "reverse" the logic here to enable logs for all configurations that are not APP_STORE? We will add a new configuration "COMMUNITY" soon. 😊

Schlabbi: Just changed it. Please let me know if this is what you had in mind.

ChristianKienle:
I think we should approve this PR and then talk about the dependency that was introduced along with it. If we decide to go with SwiftyBeaver then we would have to get the internal OK for that. If we decide to go with os_log then we would simply have to remove SwiftyBeaver but the vast majority of this PR would still make sense.

All good? :)

ChristianKienle (approved):

Schlabbi:
That's exactly what I had in mind when I built it. If SwiftyBeaver cannot be in the app just replace it. That shouldn't be a lot of work.

flashspys (approved):

haosap:
Hi @ChristianKienle ,
Before merging it, I'll ask OS colleague to review the SwiftyBeaver. After it's done, we can merge it. Thanks.

src/xcode/ENA/ENA/Source/Workers/Logger.swift
inf2381: Yes. Thank you!

inf2381:
> Hi @ChristianKienle , Hi @Schlabbi
> Before merging it, I'll ask OS colleague to review the SwiftyBeaver. After it's done, we can merge it. Thanks.

I also checked with them 😄 . SwiftyBeaver is approved. Could @Schlabbi could you please add the new dependency to THIRD_PARTY_NOTICES? After that, we can approve it. Really appreciate your efforts!

Schlabbi:
Sure, just added SwiftyBeaver to the third party notice file. Also I resolved the current merge conflicts and migrated the new code to the new logging system.

haosap:
Hi @Schlabbi ,
Good news is the [SwiftyBeaver](https://github.com/SwiftyBeaver/SwiftyBeaver) has been approved by the OS team. But I also have noticed that, the https://swiftybeaver.com/ offers the online logging service. That means, theoretically,  it's not only a logging lib, but also has online system backed. Is it correct? Thanks.

Schlabbi:
Yes, it also has a cloud service which can be configured. However, in the current configuration only logging to the console and the log file is enabled.

haosap:
Hi @Schlabbi,
Any 3rd party we use is try to avoid network connectivity due to "Datenschutz". I suggest replace   SwiftyBeaver with other logging frameworks. Thanks.

Schlabbi:
Yeah that makes sense. I will look into other frameworks or just use os_log then.

Schlabbi:
So I just removed SwiftyBeaver and implemented a new logging system from scratch. No new dependencies were used here.

Here I did not use os_log on purpose, as it is difficult to retrieve those logs later.

The new logging system does log to a file and to the console (whenever the app is not running in APP_STORE mode).

Please let me know what you think about this.

AdnanZiSAP:
> So I just removed SwiftyBeaver and implemented a new logging system from scratch. No new dependencies were used here.
> 
> Here I did not use os_log on purpose, as it is difficult to retrieve those logs later.
> 
> The new logging system does log to a file and to the console (whenever the app is not running in APP_STORE mode).
> 
> Please let me know what you think about this.

I would suggest to avoid creating file logger by hand, as it's hard to control how much the file can grow. 

Also, when choosing a 3rd party logger, it's imperative for the logger to be able to control the file size and file rotation IMHO.

Schlabbi:
> > So I just removed SwiftyBeaver and implemented a new logging system from scratch. No new dependencies were used here.
> > Here I did not use os_log on purpose, as it is difficult to retrieve those logs later.
> > The new logging system does log to a file and to the console (whenever the app is not running in APP_STORE mode).
> > Please let me know what you think about this.
> 
> I would suggest to avoid creating file logger by hand, as it's hard to control how much the file can grow.
> 
> Also, when choosing a 3rd party logger, it's imperative for the logger to be able to control the file size and file rotation IMHO.

File size could be a problem, yes. Log rotation by date can be easily implemented on top if this, if it is needed. As I understand, the log retrieval is also just meant for development purposes. Thus file logging could also be disabled in the APP_STORE mode which eliminates this problem from occuring in production.

ChristianKienle:
There are already 18 comments in this PR. Not sure if one more will make any difference but I will try:

From my point of view a file logger does more harm than good. If we are using the standard logging facilities from iOS then you can easily get the logs from your device onto your Mac by using *Console.app*. While developing this app I personally never had the urge to manually retrieve the log file from the device and look at it. For production we will not have a log file anyways.

When I look at our current logging implementation I basically see the following issues:

- We always log. Even for `APP_STORE` builds.
- We have unused code in our `Logger.swift` file.
- We log too much. There is just too much noise. I think we should either reduce the amount of logging we do (only log errors) or make errors stand out by using emojis for example:

```
❌❌❌ ERROR ❌❌❌
HTTP request failed because a timeout was reached.
```

This PR addresses some of those (in part personal) concerns.

@AdnanZiSAP maybe we should quickly discuss this tomorrow and also look at the latest changes done by @Schlabbi. Does this make sense to both of you?

Schlabbi:
Sure, that makes sense. The reason why I built the file logger in the first place is because there was a call in the old logger to retrieve the log data. This one had no function at the time as it would return empty data. The logging to file enabled me to implement this function correctly. Using os_log this function would be probably impossible to implement as retrieving the log data from within the app is (as I know) not possible.

But if the file logger needs to go just tell me and I will remove it or replace it with os_log.

ChristianKienle:
> Sure, that makes sense. The reason why I built the file logger in the first place is because there was a call in the old logger to retrieve the log data. This one had no function at the time as it would return empty data. The logging to file enabled me to implement this function correctly. Using os_log this function would be probably impossible to implement as retrieving the log data from within the app is (as I know) not possible.
> 
> But if the file logger needs to go just tell me and I will remove it or replace it with os_log.

I think removing the file logging capabilities is fine with all of us.

Schlabbi:
Sure, but then I would also take out this debug menu from the Settings where you can send the logs via email. I do not see any other option to support this any longer when we do not log to a file.

How about you talk tomorrow first and I can still take out the file logging then. Thats not even 5 minutes of work. Including a migration to os_log. Given that you still want to go with some of my changes I can implement this then 😄

Schlabbi:
Okay so I just implemented a migration to os_log and removed the FileLogger. As described above this makes the option to send logs via mail in the settings impossible to implement. I thus also removed that option.

cyborgthefirst:
I do not want to trigger the third rewrite here, but there would be the option to got with the excellent swift-log (https://github.com/apple/swift-log/) directly from Apple, more specifically the Swift Server Work Group. It supports all the requirements mentioned in the issue and offers an extensible API should the need for additional functionality arise.

inf2381:
Closed by #380

PR #38, opened 2020-05-31
[Technical] Test for DynamicTableViewController

dasdom:
Note: This doesn't add tests for all methods in this class.
This class is not that easy to test because it is only used as subclasses
and those use storyboards. I've added a fake storyboard to simulate the
loading of those classes in the tests. This is how I would test code like
this. If anyone has a better idea to test this, let me know.

- Added tests for registering the cells
- Added tests for number of sections and number of rows


dasdom:
Should I add more tests for this class before the pull request is going to be reviewed?

ChristianKienle (approved):
Looks good to me. Thank you very much for writing tests for this massive UI component. 😅 Let's wait for the second approval…

PR #48, opened 2020-05-31
[feature] Fix submission flow

thomasaugsten:


src/xcode/ENA/ENA/Source/Client/Client.swift
```diff
@@ -104,6 +104,7 @@ enum SubmissionError: Error {
 	case invalidTan
 	case serverError(Int)
 	case requestCouldNotBeBuilt
+	case simpleErrror(String)
```
fbernutz, line 107: There is a small typo in here. I think it should be `simpleError`.

melloskitten (approved):

johannesrohwer (approved):

PR #52, opened 2020-05-31
[Bug] Fix Readability of Content when using Dynamic Type

fbernutz:
Set numberOfLines in `ExposureDetection` and `RiskLegend` to 0 to let the content scale and make it fully readable when using Dynamic Type. 

To test this use Xcode's environment overrides settings, enable Dynamic Type and set a large font size or change your device setting to use a large font size. 

# `ExposureDetection`
| Before | After |
|---|---|
| Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 37 25 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 36 23 |

# `RiskLegend`
| Before | After |
|---|---|
| Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 37 45 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 35 19 |

### DIFF
--- a/src/xcode/ENA/ENA/Resources/Storyboards/ExposureDetection.storyboard
+++ b/src/xcode/ENA/ENA/Resources/Storyboards/ExposureDetection.storyboard
@@ -216,7 +216,7 @@
                                                 <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" spacing="1" translatesAutoresizingMaskIntoConstraints="NO" id="0Sp-0h-6O3">
                                                     <rect key="frame" x="20" y="16" width="334" height="45.5"/>
                                                     <subviews>
-                                                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Title" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontForContentSizeCategory="YES" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="ZsC-uh-ys5" customClass="DynamicTypeLabel" customModule="ENA" customModuleProvider="target">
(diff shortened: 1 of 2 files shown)


ChristianKienle:
@fbernutz Thank you very much for this PR. A11y is very important in general and in particular for this app. To me it is unclear what the best practice is w/r to this specific case though. Do you have any guidance on that? I mean to me your PR makes perfect sense but if you look at Apple's own apps it seems that some labels have their number of lines set to `0` and others don't. If you look at the list of your notes in *Notes.app* you may notice that the number of lines seems to be fixed at `1`.

Do you think that this is just an oversight by Apple? If yes we should definitely merge your PR and double check all other labels. Would love to hear what you have to say about this...

fbernutz:
> @fbernutz Thank you very much for this PR. A11y is very important in general and in particular for this app. To me it is unclear what the best practice is w/r to this specific case though. Do you have any guidance on that? I mean to me your PR makes perfect sense but if you look at Apple's own apps it seems that some labels have their number of lines set to `0` and others don't. If you look at the list of your notes in _Notes.app_ you may notice that the number of lines seems to be fixed at `1`.
> 
> Do you think that this is just an oversight by Apple? If yes we should definitely merge your PR and double check all other labels. Would love to hear what you have to say about this...

I think it depends on whether the content can be fully visible in a detail screen. 
In the _Notes.app_ it's furthermore content which was added by the user. So, the user might still know what was typed in earlier, but of course if not, it's possible to open the detail and see the full content. 
In our case of the cwa app, there is no option to see the full content in a detail page and it's not content the user created, so people with large fonts have simply _no option_ to read this at the moment. I think we should double check every label and set `numberOfLines = 0` more often.

dasdom:
Looking at Apple when it comes to Accessibility is always a good idea. But ignoring Apple for the moment and just looking at the screenshots in this PR, the 'after' screenshots are way more accessible. "Keine Risiko..." might not be clear to someone just checking the app, but "Keine Risikoprüfung" is crystal clear.

ChristianKienle (approved):
This PR makes a whole lot of sense. Thank you very much for making the Corona Warn App more accessible!

haosap (approved):

PR #54, opened 2020-05-31
[Bug] Use System Fonts to automatically adjust font size in `AppInformation`

fbernutz:
The screen with App Informations don't adjust font size when using dynamic type. I changed the font to `Body` and set `Automatically Adjust Font` in storyboard which is the easiest way to activate dynamic type.

To test this use Xcode's environment overrides settings, enable Dynamic Type and set a large font size or change your device setting to use a large font size. 

# `AppInformation`
| Before | After |
|---|---|
| Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 43 56 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-31 at 15 44 20 |


ChristianKienle (approved):

ChristianKienle:
Thank you very much for making the Corona Warn App more accessible.

Issue #51, opened 2020-05-31
Dark Mode Background Color in ExposureNotificationSettingVC


larsaugustin:
## Describe the bug
In the `ExposureNotificationSettingViewController`, the `TracingHistoryTableViewCell` and `DescriptionTableViewCell` use a background color on the label that's different from the background color of the cell. While both colors appear the same in light mode, they differ in dark mode.

## Expected behaviour
The same dynamic color should be used for both backgrounds. 

## Steps to reproduce the issue
The different colors can be seen when having dark mode turned on and opening the exposure notification setting view controller.
![dark mode color bug](https://user-images.githubusercontent.com/37324823/83353553-b0ad9380-a353-11ea-8456-0731db04c2be.png)

## Possible Fix
Use the same dynamic colors

larsaugustin:
I can also fix this issue!

steins-code:
done

PR #56, opened 2020-05-31
[Feature] add restricted state handling

steins-code:
The user disables the contact tracing in the OS settings causing the ENManager to fail. ENStatus will show restricted state. The user needs to enable contact tracing in the OS settings again. This UI change shall guide the user how and why to enable settings again. This design is an intermediate solution and might be changed in the future.


steins-code:
fixes #51

inf2381 (approved):

ChristianKienle (approved):

PR #69, opened 2020-05-31
Feature/background task scheduling

i336616:
Updated the handling of background tasks

What was needed:
- The previous implementation of ExposureDetectionTransaction used notifications to indicate when the detectExposures() had returned. This caused an issue for the BGTask handling, as it's a requirement of the task execution that it calls setTaskComplete to ensure that if the task expires, the system doesn't end up killing the app.
  - This we accomplished using an optional taskCompletion block passed through from ExposureDetectionTransaction's start() execution path

```
func start(taskCompletion: (() -> Void)? = nil)
private func continueWith(remoteDaysAndHours: Client.DaysAndHours, taskCompletion: (() -> Void)? = nil)
private func detectExposures(writer: AppleFilesWriter, configuration: ENExposureConfiguration, taskCompletion: (() -> Void)? = nil)
private func _detectExposures(diagnosisKeyURLs: [URL], configuration: ENExposureConfiguration, completion: @escaping () -> Void, taskCompletion: (() -> Void)? = nil)
```

Also included:
- Basic test background task to ensure the core functionality is working
- File logging for logging the background tasks activity

Summary of updates:
- Added optional taskCompletion block to ExposureDetectionTransaction start methods()
  - This is to allow the ExposureDetectionTransaction to report back to the bgTask it was called from, allowing it to setTaskComplete(_:), as is required by the API
- Added default handler for LocalNotificationAction
  - Uses ignore as the default action, if not specified
- Updated ENATaskExecutionDelegate methods
  - We handle task completion in the taskDelegate, so it's no longer necessary in the taskScheduler
- Updated implementation of taskScheduler in AppDelegate
  - backgroundTaskIdentifier is now an var in enum ENATaskIdentifier
  - AppDelegate is the taskDelegate
  - Added the ENATaskExecutionDelegate protocol methods to the AppDelegate
- Changed .exposureNotification to .detectExposures for clarity
  - Removes confusion as to which ENATaskIdentifier is which
- Appended ".exposure-notification" to the detect-exposures task
  - This ensures the task gets the BackgroundTask & ExposureNotification framework's background task prioritization


src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -20,13 +20,16 @@ import ExposureNotification
...
 	case fetchTestResults = "fetch-test-results"
 
 	var backgroundTaskScheduleInterval: TimeInterval {
 		switch self {
-		case .exposureNotification: return 15 * 60 // 2 * 60 * 60 // set to trigger every 2 hours
-		case .fetchTestResults: return 5 * 60 // 30 * 60     // set to trigger every 30 min
+		// set to trigger every 2 hours
+		case .detectExposures: return 2 * 60 * 60
+		// set to trigger every 30 min
+		case .fetchTestResults: return 2 * 60 * 60
```
thomasscholz, line 32: Comment says 30 minutes; but two hours a returned

i336616: It's the comment that's wrong, will fix that now. Good catch!

i336616 (commented):
Happy with these changes 👍 

edit: oops, there's a parsing error, will check that

haosap (approved):

ChristianKienle (approved):

PR #75, opened 2020-05-31
[Technical] Unit tests for SQLiteKeyValueStore (closes #21)

### LINKED ISSUE #21: Write Tests for SQLiteKeyValueStore

SebastianWild:
This PR adds unit tests to `SQLiteKeyValueStore`, as well as some doc comments.  

Some behaviors of `SQLiteKeyValueStore` have also been modified:
- inserting empty `Data` into the store now results in getting empty `Data` out instead of `nil`
- accessing the value for a key that does not exist now returns `nil` instead of empty `Data`
- it is now possible to overwrite a value with `nil`

### DIFF
--- a/src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
+++ b/src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
@@ -18,6 +18,7 @@
 import FMDB
 import Foundation
 
+/// Basic SQLite Key/Value store with Keys as `TEXT` and Values stored as `BLOB`
 class SQLiteKeyValueStore {
 	private let db: FMDatabase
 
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -45,15 +46,16 @@ class SQLiteKeyValueStore {
...
 		openDbIfNeeded()
 
 		do {
 			let query = "SELECT value FROM kv WHERE key = ?;"
 			let result = try db.executeQuery(query, values: [key])
-			var resultData = Data()
+			var resultData: Data?
 			while result.next() {
-				guard let data = result.data(forColumn: "value") else {
+				guard let data = result.dataNoCopy(forColumn: "value") else {
```
SebastianWild, line 58: I'd like to call out attention to this change here - everything seems to work fine but this change had to be made to support getting empty `Data` out of the DB. Without this, behavior is as follows.

Say we do this `kvStore["hello"] = Data()`, we should arguable expect `kvStore["hello"]?.isEmpty == true`
However with the current code this is not the behavior, we get this:
`result.columnIsNull("value")` -> `false`
`result.data(forColumn: "value")` -> `nil` but if we do instead
`result.dataNoCopy(forColumn: "value")` -> empty `Data`

To me this seemed the easiest fix, but it doesn't quite sit right - even though the tests pass.

src/xcode/ENA/ENA/Source/Workers/__tests__/SQLiteKeyValueStoreTests.swift
```diff
@@ -0,0 +1,137 @@
...
+		("developerDistributionBaseURLOverride", "testing4".data(using: .utf8)!),
+		("developerVerificationBaseURLOverride", "testing5".data(using: .utf8)!)
+		// swiftlint:enable force_unwrapping
+	]
+
+	override func setUp() {
+		super.setUp()
+		// Old DB is deinited and hence connection closed at every setUp() call
+		// swiftlint:disable:next force_unwrapping
+		kvStore = SQLiteKeyValueStore(with: URL(string: "file::memory")!)
```
ChristianKienle, line 42:
```swift
URL(staticString: "file:memory")
```
allows you to get rid of the `!`.

ChristianKienle (approved):

llanghammer-sap (approved):

PR #79, opened 2020-06-01
[Bug] Onboarding "Continue" non-functional

SebastianWild:
## Summary

It is an error to call `ENManager.activate()` more than once without disabling it first.
Current onboarding & error handling logic did not cover this case (doesn't handle `ENErrorCodeAPIMisuse`), and so the button became non-functional.

The fix was to handle this error - and in the case of onboarding simply go to the next screen.

## Steps to Reproduce

1. Delete app & clear permissions cache (so you get prompted again)
2. In the third onboarding page, grant permissions as normal
3. Observe that we go to the next page
4. Hit the back button in the nav bar
5. Observe that the continue button does not work anymore.

## Alternative Fixes Considered

We could simply guard against the `enabled` precondition on `ExposureManager` like so:

```swift
guard !exposureManager.preconditions().enabled else {
	completion?()
        return
}
```

## Future Considerations

A scenario this PR does not cover is this flow:
1. User hits "Continue"
2. User denies Exposure Notifications in the system prompt
3. User is directed to next onboarding step
4. User, having changed their mind, goes back and hits "Continue"
5. The app just goes forward an onboarding step.

A nice UX thing here might be to direct the user to Settings.app

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
@@ -22,6 +22,8 @@ enum ExposureNotificationError: Error {
 	case exposureNotificationRequired
 	case exposureNotificationAuthorization
 	case exposureNotificationUnavailable
+	/// Typically occurs when `activate()` is called more than once.
+	case apiMisuse
 }
 
 struct ExposureManagerState {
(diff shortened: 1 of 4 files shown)


haosap:
Hi @SebastianWild ,
Yes. you're right. Current code doesn't cover all error handlings. ExposureManager should be revised carefully again.

i336616 (approved):
The flow is now:
- user taps Weiter
- user taps Don't Enable
- user taps Züruck
- user taps Weiter
=> user is no longer prompted to enable/disable exposure notifications and progresses to the next screen

Fix works as described. We should readdress this though, and prompt the user for exposure notification permissions again if they tap back to this screen, regardless of their previous choice.

SebastianWild:
> Fix works as described. We should readdress this though, and prompt the user for exposure notification permissions again if they tap back to this screen, regardless of their previous choice.

I will do another pass and see what we can do 👍

ChristianKienle (approved):

PR #85, opened 2020-06-01
[Feature] SQLCipher Integration and SecureStore Update (closes #125)

### LINKED ISSUE #125: [BSI][20200601] Sensitive Data Stored in Unencrypted Database on Filesystem
Rating: Low
 
Description:
The iOS app stores sensitive data in an unencrypted sqlite Database in the
app's sandbox. An attacker with root privileges, or code execution in the
scope of the CWA app can read this information. This includes data, such as
the TeleTAN or registration token. A code comment mentions that this database
"still needs to be encrypted" [1]. If this measure is implemented, the issue
is mitigated.

However, it is questionable why the TeleTAN needs to be stored at all. As it
should only be used once to create a ``sessionToken``, there should be no need
to store the TeleTAN on disk.

[1]: cwa-app-ios/src/xcode/ENA/ENA/Source/Services/ExposureSubmissionService.swift:286

Proof of Concept:
An excerpt of the binary content of the ``secureStore.sqlite`` in ``Library/Application Support``:
```
secureStore.sqlite
SQLite format 3
{tablekvkv
CREATE TABLE kv (
    key TEXT UNIQUE,
    value BLOB
indexsqlite_autoindex_kv_1kv
teleTan"[REDACTED]"
isOnboardedtrue0
[...]
```

llanghammer-sap:
**Summary**

This PR integrates [SQLCipher](https://www.zetetic.net/sqlcipher) into the project for encrypting the Seucre the Secure Store. 
The integration is done via the Source Integration described [here](https://www.zetetic.net/sqlcipher/ios-tutorial/).

The SecureStore was updated to use the SQLCipher resources as I wasn't able to connect FMDB with the SQLCipher sources.

**How to Test**
Storing and reading values from the SecureStore should work as exspected, for example during the onboarding process when the corresponding value for onboarding and term agreement are stored.


ChristianKienle:
This PR has to be carefully reviewed. I would also like to consider alternative approaches. What you did is to basically roll your own SQLite wrapper.

From my experience writing any SQL wrapper is kind of tricky. Even though SQLite has a nice and easy C API there are tricky edge cases. I personally would like to avoid rolling our own SQLite wrapper/using the C API directly if possible. I know that you looked into ways how to combine FMDB with SQLCipher and probably concluded that it is not possible. I would like to also take a look at it - just to double check. This I why I added the on-hold label for now.

llanghammer-sap:
I completly agree with you. Better to be save than sorry. If you find a way to get them working together. I will adjust the Code to use FMDB again.

haosap:
Some tips:
https://github.com/ccgus/fmdb/blob/master/FMDB.podspec
From the podspec file, we might find the way to integrate SQLCipher with FMDB.

ChristianKienle:
We just had a call where I shared a minimal sample with @llanghammer-sap that shows how to combine FMDB with SQLCipher + Swift PM.

Sn0wfreezeDev (commented):
I read through the implementation and to me it looks fine. 
I commented on the things that could get some enhancements.

One thing, where I would like to have some clarifications is that I don't see the reason why the whole  sqlite3 should be added to the project. iOS has a library that can be linked called `libsqlite3`. I have used it in the past and it worked reliably. 
Is the added sqlite version newer or does the iOS library not have all needed features?

llanghammer-sap:
> One thing, where I would like to have some clarifications is that I don't see the reason why the whole sqlite3 should be added to the project

@Sn0wfreezeDev The whole sqlite3 has been added as part of the [SQLCipher Source Integration](https://www.zetetic.net/sqlcipher/ios-tutorial/). So the sqlite3.c and sqlite3.h files are part of the integration process and are needed for SQLCipher.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -26,13 +26,36 @@ class SQLiteKeyValueStore {
...
+			key = generatedKey
+			if savetoKeychain(key: "secureStoreDatabaseKey", data: Data(key.utf8)) == noErr {
+				logError(message: "Unable to save Key to Keychain")
+				return
+			}
+		}
+
+		let dbhandle = OpaquePointer(db.sqliteHandle)
+
+		guard sqlite3_key(dbhandle, key, Int32(key.count)) == SQLITE_OK else {
```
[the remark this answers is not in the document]
llanghammer-sap: There are 2 version: 
1. Inital Setup <-- Only the empty file is created and no values can be saved. On the next time the sqlite3_key call is successfull the key will be set and the database will be created. 
2. Database already created <-- Access to the Database will fail as the key is not set. 

So as far as I could test no data will be saved in a non encrypted database. But it is a good point, as this failing should be adressed when trying to connect to the db later on. 

If anyone has other findings feel free to correct me so we can deliver a good app.

llanghammer-sap:
@ChristianKienle a review from your side would be awesome.

ChristianKienle:
> @ChristianKienle a review from your side would be awesome.

On it. :)

ChristianKienle:
> iOS has a library that can be linked called libsqlite3. I have used it in the past and it worked reliably. Is the added sqlite version newer or does the iOS library not have all needed features?

@llanghammer-sap has already explained the reasons behind including `sqlite3.{c,h}`. Those files have been created by using the [official SQLCipher + Xcode tutorial](https://www.zetetic.net/sqlcipher/ios-tutorial/). We have to encrypt certain files to protect your privacy. The included `sqlite3.{c,h}` files are compatible with the official SQLite implementation but add transparent full database encryption even while the device is unlocked.

Hope this gives more background information w/r to why those files are there. :)

ChristianKienle (changes requested):
I think we should set `-DNDEBUG` only for `Release` and `ReleaseAppStore`.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -23,26 +23,56 @@ class SQLiteKeyValueStore {
...
+			logError(message: "Database could not be opened")
+			return
+		}
+		initDatabase()
 	}
 
 	deinit {
 		db.close()
 	}
+	private func initDatabase(){
```
ChristianKienle, line 39:
```suggestion
	private func initDatabase() {
```

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -163,3 +191,55 @@ class SQLiteKeyValueStore {
...
+	}
+
+	func generateDatabaseKey() -> String? {
+		var bytes = [UInt8](repeating: 0, count: 32)
+		let result = SecRandomCopyBytes(kSecRandomDefault, bytes.count, &bytes)
+		guard result == errSecSuccess else {
+			logError(message: "Error creating random bytes.")
+			return nil
+		}
+		return "x'"+Data(bytes).hexEncodedString()+"'"
```
ChristianKienle, line 230:
```suggestion
		return "x'\(Data(bytes).hexEncodedString())'"
```

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -163,3 +191,55 @@ class SQLiteKeyValueStore {
...
+}
+
+/// Extensions for Hexencoding when generating key
+extension Data {
+	struct HexEncodingOptions: OptionSet {
+		let rawValue: Int
+		static let upperCase = HexEncodingOptions(rawValue: 1 << 0)
+	}
+
+	func hexEncodedString(options: HexEncodingOptions = []) -> String {
```
ChristianKienle, line 241: May I suggest to simply remove `HexEncodingOptions` completely? As far I can tell `hexEncodedString` is only ever called without any options. This is just code that is not used by anyone. Or do you have some use case in mind?

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -23,26 +23,56 @@ class SQLiteKeyValueStore {
...
+		initDatabase()
 	}
 
 	deinit {
 		db.close()
 	}
+	private func initDatabase(){
+		var key: String
+		if let keyData = loadFromKeychain(key: "secureStoreDatabaseKey") {
+			key = String(decoding: keyData, as: UTF8.self)
```
ChristianKienle, line 42: Why not `String(data:encoding:)`?

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -23,26 +23,56 @@ class SQLiteKeyValueStore {
...
+		} else {
+			guard let generatedKey = generateDatabaseKey() else {
+				db.close()
+				return
+			}
+			key = generatedKey
+			if savetoKeychain(key: "secureStoreDatabaseKey", data: Data(key.utf8)) == noErr {
```
ChristianKienle, line 49: Why not `key.data(using: .utf8)`? I have never used the `.utf8` property to get the data to be honest... Did I miss something? :D

ChristianKienle (changes requested):
Sorry for the review taking so long. I found some things + have questions. 
In general I would like to suggest to not close the database after every read/write operation. Just leave it open. But this is only my feeling. I suggest we have a quick call tomorrow so that this can be merged.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
llanghammer-sap: Added with https://github.com/corona-warn-app/cwa-app-ios/pull/85/commits/511354c0e70698651c96f1835de360bb9c654504

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
llanghammer-sap: Updated with https://github.com/corona-warn-app/cwa-app-ios/pull/85/commits/511354c0e70698651c96f1835de360bb9c654504

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
llanghammer-sap: That is true. It is not going to be needed, so removed with https://github.com/corona-warn-app/cwa-app-ios/pull/85/commits/511354c0e70698651c96f1835de360bb9c654504

ChristianKienle (approved):

haosap (approved):

PR #90, opened 2020-06-01
[Feature] Implemented Risk Legend According to Latest Design Specifications

eisvogel:
This PR implements the risk legend according to the latest design specifications:

IMG_D748AEFF2B85-1

Additionally it introduces the following:
* `ENALabel` which implements the typography configuration according to the design guides.
* Minor bug fix to the `DynamicTableViewController`
* Pending refactorings and improvements for the `DynamicTableViewController`. Changes will be refactored into the actual controller in due course, once submission team has merged their current progress to avoid merge conflicts.


src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -258,7 +258,7 @@ extension DynamicTableViewController {
 
 	func tableView(_: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
 		if dynamicTableViewModel.section(at: indexPath).isHidden(for: self) {
-			return 0
+			return .leastNonzeroMagnitude
```
ChristianKienle, line 261: Why that change?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import Foundation
+import UIKit
+
+class DynamicTableViewSpaceCell: UITableViewCell {
+	private lazy var heightConstraint: NSLayoutConstraint = self.contentView.heightAnchor.constraint(equalToConstant: 0)
+
+	var height: CGFloat {
+		set {
+			if newValue == UITableView.automaticDimension {
```
ChristianKienle, line 28: I am always a bit careful then comparing non-discrete numerical types like `CGFloat`s. What do you think about checking if `newValue` is roughly equal to `UITableView.automaticDimension`?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
```diff
@@ -0,0 +1,46 @@
...
+				} else {
+					heightConstraint.constant = newValue
+				}
+				heightConstraint.isActive = true
+			}
+		}
+		get { heightConstraint.isActive ? heightConstraint.constant : UITableView.automaticDimension }
+	}
+
+	override func prepareForReuse() {
```
ChristianKienle, line 42: I think you should call super here.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
```diff
@@ -0,0 +1,46 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+import Foundation
+import UIKit
+
+class DynamicTableViewSpaceCell: UITableViewCell {
```
ChristianKienle, line 23: To be honest, using a cell to add some spacing between cells seems a bit overkill. But maybe it is just me.

Using a cell purely for layout related things means that it will always be some kind of special case. This may not be an issue now but imagine the impact a spacing cell has on anything related to `IndexPath`s: Selection handling, gesture handling (swipe to delete, …). Also the space cell is only used once as far I can tell. Have you considered simply increasing the padding/margin of the body cell? Wouldn't that have the same effect?

ChristianKienle (changes requested):
I have added a few comments that may require a couple of changes.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
EisvogelSAP: Returning 0 at this point may result in the default height rather than 0.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
EisvogelSAP: This should be fine, as `UITableView.automaticDimension` will either be used as constant or it will not. Therefore a direct comparison should not be of concern for this matter. This is how UIKit handles the property internally as well.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
EisvogelSAP: Since the point of the `DynamicTableViewController` is to allow a declarative design approach this cell may have some benefit. Alternatively one has to split semantically linked sections into multiple to have the desired spacing between cells.

One has two be careful about the index path with this controller anyway, for the exact same reason of the declarative design approach.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewSpaceCell.swift
EisvogelSAP: Done.

ChristianKienle (approved):

llanghammer-sap (approved):

PR #91, opened 2020-06-01
[Bug ] Fix leaking keyboard observer with possible memory leak.

eisvogel:
This PR fixes a leaking keyboard observer with possible memory leak.


thomasaugsten (approved):

ChristianKienle (approved):
Good catch!

PR #92, opened 2020-06-01
[Feature] enhanced onboarding with checking for bluetooth 

steins-code:
Add alert to notify users of disabled bluetooth during onboarding. The user has the option to pass the onboarding without enabling bluetooth by tapping on "Later" or the user can directly go to settings to enable bluetooth. The alert pops up only when bluetooth is disabled.


inf2381 (approved):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -198,7 +210,7 @@ final class OnboardingInfoViewController: UIViewController {
 	// MARK: Exposure notifications
 
 	private func askExposureNotificationsPermissions(completion: (() -> Void)?) {
-		if TestEnvironment.shared.isUITesting {
+		guard !TestEnvironment.shared.isUITesting && !exposureManagerActivated else {
```
inf2381, line 213: I don't like `TestEnvironment.shared.isUITesting` inside our code but that's not part of this PR

i336616 (approved):
👍

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
i336616: Do we have a preferred way to check for UI testing state at the moment?

PR #93, opened 2020-06-01
[Bug] Navigation Bar image rendering in dark mode (closes #89)

### LINKED ISSUE #89: Navigation bar image does not adapt to dark mode (HomeView)
## Describe the bug

The asset "navi_bar_icon.pdf" in "Assets.xcassets" is missing an dark appearance, so it looks off when 'dark mode' is selected as the 'userInterfaceStyle'. Also, the navigationItem is never redrawn after init creation, so even if there is an asset for the dark mode, when the app is running and the user initiates an appearance change, the navigationItem will not adapt to the new environment style.

## Expected behaviour
Expected handling | Current handling
------------ | -------------
 | 

## Possible fix
* Creating an asset file (vector) with a white text color to display correctly when dark mode is enabled.
* Redraw the navigationItem when the appearance is changed (dark-to-light or vice versa).

I already created a vector file with the same specs as the current asset file but with a white text color and called the redrawing of the navigationItem inside the super method 'traitCollectionDidChange' from UIViewController. If it's possible to merge my commit, I will create a pull-request for this issue, so just let me know.

tens0rfl0w:
This pull request fixes issue #89.
* Added a second vector file to the asset collection 'navi_bar_icon' containing the same logo/image, but with white text color to adapt to dark mode.
* Fixed the metadata for the mentioned asset collection. (I also added the static propertie ` "properties" : {
    "template-rendering-intent" : "original"
  }` so the propertie tag in the source code for rendering the image as 'orginal' can be removed.
* I moved the initial creation of the 'navigationItem.leftBarButtonItem' to the super method 'viewDidLayoutSubviews()' so it gets properly redrawn when the appearance of the UI changes (from light to dark or vice versa).


haosap (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -97,6 +97,13 @@ final class HomeViewController: UIViewController {
 		super.viewWillDisappear(animated)
 		NotificationCenter.default.removeObserver(summaryNotificationObserver as Any, name: .didDetectExposureDetectionSummary, object: nil)
 	}
+	
+	override func viewDidLayoutSubviews() {
+		let image = UIImage(named: "navi_bar_icon")
+		let leftItem = UIBarButtonItem(image: image, style: .plain, target: nil, action: nil)
+		leftItem.isEnabled = false
+		self.navigationItem.leftBarButtonItem = leftItem
+	}
```
haosap, line 106: why not do it traitCollectionDidChange?

EisvogelSAP: Actually this should be handle by using appearances. There is no code required to change an image based on the current interface style.

tens0rfl0w: Also tried that implementation, but I tested both methods for calls that are unrelated to appearance changes (dark mode), and from what I've seen 'traitCollectionDidChange' was called a little bit more often than 'viewDidLayoutSubviews'. But sure both is possible without any difference. Actually Apple mentions both methods (and many more) as possibilities to place code like that.

So If you think that 'traitCollectionDidChange' is a better place for that kind of stuff, I will change that!

EisvogelSAP: Why does it have to be called in any of those two methods, anyway? Shouldn't that be a one-off configuration in `viewDidLoad`. Am I missing something?

tens0rfl0w: > Why does it have to be called in any of those two methods, anyway? Shouldn't that be a one-off configuration in `viewDidLoad`. Am I missing something?
@EisvogelSAP 
Thanks for your comment! As per my understanding `viewDidLoad()` is only called once when the controller view is loaded into the memory. It is possible to initialize the navigationItem in this method, but if the appearance changes when the app ist running, the navigationItem can't be re-loaded, so it would stuck on the ‚old‘ appearance. So practically you have to restart the app every time you change the appearance, to see the change happen on the navigationItem, because of the restricted one-time calling.

haosap: Hi @tens0rfl0w,
Which method got called more offer, I think, it depends on what the experiment did. If you have manipulate the view hierarchy very often,  the `viewDidLayoutSubviews` will got called more often than `traitCollectionDidChange`. For our project, we did a lot of view manipulation on the home screen. So I think `traitCollectionDidChange` is better. Thanks.

haosap (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
tens0rfl0w: @haosap Fixed in the newest commit, please have a look!

haosap:
Hi @tens0rfl0w ,
Please resolve the conflict.

tens0rfl0w:
> Hi @tens0rfl0w ,
> Please resolve the conflict.
@haosap 
Fixed!

PR #95, opened 2020-06-01
[Feature] ENAButton styles and tap states

eisvogel:
This PR implements predefined button styles for the `ENAButton` including tap states.

The following button states are supported:
image

The respective properties are exposed to interface builder, but can also be used in code (`isTransparent` and `isInverted`). The emphasized color can be set using the `color` property of the button.

The button automatically applies the default font being System, 17pt, Semibold and adjust with dynamic type. When tapped the button is highlighted with a black view with 20% opacity.

To use the button in interface builder a standard `UIButton` with a custom class can be used. There is no need to set any of the properties manually.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
+++ b/src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
@@ -23,10 +23,7 @@ class ENAButton: DynamicTypeButton {
 	@IBInspectable var color: UIColor?
 
 	@IBInspectable var isTransparent: Bool = false { didSet { applyStyle() } }
-	@IBInspectable var isLight: Bool = false { didSet { applyStyle() } }
 	@IBInspectable var isInverted: Bool = false { didSet { applyStyle() } }
-
-	private var style: Style = .regular
 
 	override var isEnabled: Bool { didSet { applyStyle() } }
 	override var isHighlighted: Bool { didSet { applyHighlight() } }
@@ -66,15 +63,15 @@ class ENAButton: DynamicTypeButton {
 
 		clipsToBounds = true
 
-		contentEdgeInsets = UIEdgeInsets(top: 14, left: 14, bottom: 14, right: 14)
+		contentEdgeInsets = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
 		heightAnchor.constraint(greaterThanOrEqualToConstant: 50).isActive = true
-		titleLabel?.numberOfLines = 0 // TODO
 
 		highlightView?.removeFromSuperview()
 		highlightView = UIView(frame: bounds)
 		highlightView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
 		addSubview(highlightView)
 
+		titleLabel?.font = .preferredFont(forTextStyle: .body)
 		cornerRadius = 8
 		dynamicTypeSize = 17
 		dynamicTypeWeight = "semibold"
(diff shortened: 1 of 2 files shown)


AdnanZiSAP:
Shouldn't all screens be adjusted to use these new properties? Like, Reset screen has a red button, for example. What's the idea how these changes should be applied?

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -15,55 +15,148 @@
...
 		heightAnchor.constraint(greaterThanOrEqualToConstant: 50).isActive = true
+		titleLabel?.numberOfLines = 0 // TODO
+
+		highlightView?.removeFromSuperview()
+		highlightView = UIView(frame: bounds)
+		highlightView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+		addSubview(highlightView)
 
-		titleLabel?.adjustsFontForContentSizeCategory = true
+		cornerRadius = 8
```
steins-code, line 78: I would prefer standard corner radius of six points
https://developer.apple.com/design/human-interface-guidelines/watchos/interface-elements/buttons/

EisvogelSAP:
Yes, all screens should be adjusted in due course. But to avoid major merge conflicts I would prefer to have this one merged first. Then gradually apply the properties to every screen. What do you think?

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
EisvogelSAP: Actually this should have been 10 according to the design specs. The page you linked is for watchOS only. iOS itself does not know buttons as we use them here at all.

AdnanZiSAP:
> Yes, all screens should be adjusted in due course. But to avoid major merge conflicts I would prefer to have this one merged first. Then gradually apply the properties to every screen. What do you think?

I tend to disagree on that. I _would_ agree if there would be _one_ PR coming right after this one where all screens will be adapted. IMO, opening more PRs for this, potentially even more than 5, at this stage, it's too late for that. Also, there's a story to make this button adapt nicely to larger fonts. Ideally we should open max 2 PRs for each, one for implementation and one for adoption. Or do everything in one PR.

EisvogelSAP:
Agree, will follow up on this once merged. I'll try to update all screens in on PR.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -15,55 +15,148 @@
...
+import Foundation
 import UIKit
 
-/// A Button which has the same behavior of UIButton, but with different tint color.
 @IBDesignable
-class ENAButton: UIButton {
-	override init(frame: CGRect) {
-		super.init(frame: frame)
-		customizeButton()
+class ENAButton: DynamicTypeButton {
```
AdnanZiSAP, line 22: This button was crashing IB. I've just checked it, and it seems it doesn't anymore, but please double check this.

EisvogelSAP: It needs a clean build for it work properly on the first attempt, yes. After that it should be resolved. This is due to the way named UIColors are handled.

EisvogelSAP: If you encounter another issue, please post your IB log file with the corresponding stack trace.

AdnanZiSAP (approved):

steins-code (approved):

PR #97, opened 2020-06-01
[Bug] Disable swiftlint force_cast on DynamicTableViewControllerTests

i336616:
Added disable force_cast to DynamicTableViewControllerTests to allow builds on CI


eisvogel (changes requested):
This should rather be a test assertion with `if let` rather than suppressing a forced downcast

i336616:
> This should rather be a test assertion with `if let` rather than suppressing a forced downcast

👍

i336616:
> > This should rather be a test assertion with `if let` rather than suppressing a forced downcast
> 
> 👍

Done, replaced with guard & assert instead

thomasaugsten (approved):

eisvogel (approved):

PR #100, opened 2020-06-01
[bug] TAN entry: continue button inadvertently active on pulling down VC

mn033:
When entering a TAN, pulling down the modal view controller window slowly results in the "continue" button to be activated due to a call to "viewWillDisappear".

The button gets activated and can be pressed by the user, upon releasing the view (bouncing back to the top) - even though not all TAN fields are filled.

The solution would be, to activate the button on viewDidDisappear, when it is completely offscreen. 

Current | Fixed
-- | --
current | fixed


thomasaugsten (approved):

thomasaugsten:
good catch, thanks

ChristianKienle (approved):
Thank you very much!

Issue #103, opened 2020-06-01
UICollectionViewCells do not have the correct background color (HomeView)


Bohne9:
## Describe the bug
Some UICollectionViewCells on the HomeView do not have the correct background color in the dark mode.
Currently (in the dark mode) the background color is set to black which does not contrast with the background color of the UICollectionView.

## Expected behaviour
<img width="398" alt="Screenshot 2020-06-01 at 18 00 40" src="https://user-images.githubusercontent.com/24317813/83428045-ff315f80-a431-11ea-9cb9-e076a32582c3.png">

## Possible Fix
A possible fix would be to set the background color of the container view in the cells to **UIColor.secondarySystemGroupedBackground**. That causes the cell to have a white background in light mode and the gray you can see in the image above in dark mode.

I already started fixing the cells background color.

PR #105, opened 2020-06-01
Changed the background color of the home screen cells container view …

jonahschueller:
…to UIColor.secondarySystemGroupedBackground.


Bohne9:
Refers to issue: https://github.com/corona-warn-app/cwa-app-ios/issues/103

ChristianKienle:
Looks good to me. Certainly better than what we have at the moment. That being said dark mode is currently under refinement and we will give more love to dark mode starting tomorrow evening. I will merge this but leave the issue open...

ChristianKienle (approved):

PR #108, opened 2020-06-01
[Technical] Add tests for DynamicTableViewModel

dasdom:
This commit adds tests for all methods in DynamicTableViewModel.
I didn't want to change the code to make the tests easier. As a result
I couldn't compare DynamicSections directly. So I compare the
number of cells and the content of the contained cell.


ChristianKienle (approved):
Thank you very much Dominik.

PR #116, opened 2020-06-01
[Feature] Notify user on status change (closes #770)

### LINKED ISSUE #770: [UI] Load risk view before rendering
## Avoid duplicates

* [x] This issue has not already been raised before
* [x] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation
If you launch the app, the risk view cell initially renders as grey and with the status „unknown“ and then loads the actual risk and changes the appearance 
## Suggested Enhancement
Load the actual risk status before rendering the risk view cell, so the first render is the correct one.
## Expected Benefits
Prettier UI is always a good thing😄

---
Internal Tracking ID: EXPOSUREAPP-1946

i336616:
iOS: A notification should be created with text, "Es gibt Neuigkeiten von Ihrer Corona-Warn-App", if the risk level changes(low <-> high)

When the app is backgrounded, a background task is scheduled to occur no earlier than 2 hours time. When the task event occurs, it will call the detect exposures API. If the new maximumRiskScore value returned is greater than the previously persisted value, a notification will be presented to the user, indicating that an update is ready for them to view in the app.


ChristianKienle (approved):

src/xcode/ENA/ENA/Source/Models/Exposure/RiskLevel.swift
```diff
@@ -18,8 +18,8 @@
 import ExposureNotification
 import Foundation
 
-enum RiskLevel {
-	case unknown
+enum RiskLevel: Int, Comparable {
+	case unknown = 0
 	case inactive
 	case low
 	case high
```
haosap, line 25: This is changed by @SebastianWild .

haosap (approved):

haosap:
Hi @i336616 , please resolve the conflict. Especially the one RiskLevel from @SebastianWild

PR #147, opened 2020-06-02
[Feature] If risk level changes, a local notification should be pushed (closes #209)

i336616:
Background task logic check for determining if the risk level has increased since the previous time a detectExposures() API call was made. 
Present a User Notification, if so.

[issue here](https://sapjira.wdf.sap.corp/projects/EXPOSUREAPP/issues/EXPOSUREAPP-209)

## Checklist

* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate.swift
@@ -223,6 +223,18 @@ extension AppDelegate: ENATaskExecutionDelegate {
 				return
 			}
 
+			if let previousSummary = self.store.previousSummary {
+				let previousRiskLevel = RiskLevel(riskScore: previousSummary.maximumRiskScore)
+				let newRiskLevel = RiskLevel(riskScore: newSummary.maximumRiskScore)
+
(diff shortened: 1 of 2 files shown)


EisvogelSAP:
Do we really want a notification for the user if the risk level changes from unknown to inactive or low? I think the user is only interested in high risk notifications.

If we really want to got this far in terms of notifications, I guess it wouldn't hurt to also inform the user of a decreased risk, say when high drop to low, or when we go back unknown for whatever reason.

What do you think?

i336616:
Agreed, I’ll amend and update the PR

Sent from my iPhone

On 2 Jun 2020, at 17:46, Marc-Peter Eisinger <notifications@github.com> wrote:

﻿

Do we really want a notification for the user if the risk level changes from unknown to inactive or low? I think the user is only interested in high risk notifications.

If we really want to got this far in terms of notifications, I guess it wouldn't hurt to also inform the user of a decreased risk, say when high drop to low, or when we go back unknown for whatever reason.

What do you think?

PR #154, opened 2020-06-02
removed singleton

ChristianKienle:


src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -25,11 +25,12 @@ protocol CoronaWarnAppDelegate: AnyObject {
 	var client: Client { get }
 	var downloadedPackagesStore: DownloadedPackagesStore { get }
 	var store: Store { get }
+	var taskScheduler: ENATaskScheduler { get }
 }
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
-	let taskScheduler = ENATaskScheduler.shared
+	let taskScheduler = ENATaskScheduler()
```
haosap, line 33: The non-singleton class wraps a singleton class BGTaskScheduler?

i336616: @ChristianKienle  We've concerns about the possibility of registering the same background task multiple times. Given that the ENATaskScheduler is a shell wrapper around the BGTaskScheduler (which itself is implemented as a singleton), we felt it would be fine to implement our class in the same way.

ChristianKienle: I think this is solved by my latest commit...

PR #160, opened 2020-06-03
[Technical] Don't allow http calls anymore (closes #123)

### LINKED ISSUE #123: [BSI][20200601] App Transport Security Partially Disabled
Rating: Informational
 
Description:
App Transport Security (ATS) is a security feature introduced in iOS
9 to ensure that the app's network connections employ only industry-
standard protocols and ciphers without known weaknesses. ATS is 
enabled by default for any iOS app. In this case, ATS is explicitly
disabled for the distribution mock server domain. The configuration
allows connections in plaintext over HTTP for the domain and all its
subdomains.
This exception is currently defined for the mock server and 
presumably only required for testing. However, this configuration
should not be present in the production variant of the app or used
for production servers in the CWA ecosystem.
 
Proof of Concept:
The configuration can be found in the app's Info.plist file starting
on line 38:

```
<key>NSAppTransportSecurity</key>
<dict>
	<key>NSAllowsArbitraryLoads</key>
	<true/>
	<key>NSExceptionDomains</key>
	<dict>
		<key>distribution-mock-cwa-[...]</key>
		<dict>
			<key>NSExceptionAllowsInsecureHTTPLoads</key>
			<true/>
			<key>NSIncludesSubdomains</key>
			<true/>
		</dict>
	</dict>
</dict>
```

inf2381:


EisvogelSAP:
Why not remove the key all together to use the system default?

ChristianKienle:
@inf2381 I think we should remove the key from the plist.

ChristianKienle (changes requested):
please remove the key completely.

ChristianKienle (approved):

PR #162, opened 2020-06-03
Improve logging performance and don't log in production

tapwork:
- Prefer print over NSLog since print is less expensive
- Add debug level, file, function and line to log output
- Do not log on production

Logging with NSLog is more expensive than with print. Also please log only in debug sessions not in production.
See: https://developer.apple.com/library/archive/technotes/tn2347/_index.html or https://stackoverflow.com/questions/25951195/swift-print-vs-println-vs-nslog

This PR also adds more context like the file, function and line: see the following example: 
Screenshot 2020-06-03 at 09 15 03

### DIFF
--- a/src/xcode/ENA/ENA/Source/Workers/Logger.swift
+++ b/src/xcode/ENA/ENA/Source/Workers/Logger.swift
@@ -19,12 +19,14 @@ import Foundation
 
 let appLogger = Logger()
 
-func log(message: String, level _: LogLevel = .info, file _: String = #file, line _: UInt = #line, function _: String = #function) {
-	NSLog("%@", message)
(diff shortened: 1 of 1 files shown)


tkowark:
Dear @tapwork , as the team is working with almost full capacity on implementing all the backlog items, please excuse some delays in the PR review. We hope that there might be more time for that after the weekend.

Would be great if you could keep it up to date with changes to the base branch in the meantime. For further PRs, we kindly ask to only provide ones for issues that the team marked for community contributions and assigned to you. Thanks for your understanding.

ChristianKienle (approved):

ChristianKienle:
Will be merged as part of #380. Thank you very much for contributing! I only made small adjustments in #380.

PR #165, opened 2020-06-03
[Bug] Notifications state on Settings screen (#1008)

AdnanZiSAP:
This is a very simple change that allows showing Notifications state in Settings screen as On/An, when any of two types of notifications are on.


i336616 (approved):
👍

eisvogel (approved):

PR #198, opened 2020-06-04
[Feature] Generate new database key after resets and reinstalls

llanghammer-sap:
## Checklist

* [ ] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)

## Description
After a manual reset or after an app reinstall the database key should be changed. This PR handles the check if it is a first app start (first installation, reinstallation) and the rekeying after a reset with storing the new db key in the keychain. 

## How to test
Start the App and everything should work as exspected. Afterwards you can manually reset the app and see if everything is working as exspected


inf2381 (approved):

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -151,12 +175,27 @@ class SQLiteKeyValueStore {
 			try db.executeUpdate(deleteStmt, values: [])
 			try db.executeUpdate("VACUUM", values: [])
 			log(message: "Flushed SecureStore", level: .info)
+			rekeyDatabase()
```
ChristianKienle, line 178: Why do we not just wipe everything on reset? Even the URL overrides? I mean reset should just outright reset everything... shouldn't it? Resetting to be means to just close the database, if open and then delete the database files... This would also remove the need for a rekey...

ChristianKienle (changes requested):
see comment about just deleting the db files.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -59,12 +67,29 @@ class SQLiteKeyValueStore {
...
 		let sqlStmt = """
+		PRAGMA auto_vacuum=2;
+
 		CREATE TABLE IF NOT EXISTS kv (
 			key TEXT UNIQUE,
 			value BLOB
 		);
 		"""
-		db.executeStatements(sqlStmt)
+		if !db.executeStatements(sqlStmt) {
```
llanghammer-sap, line 77: Just want to point to this change to get some Feedback on it. When the database can't be accessed after the setup process with setting the encryption key. We have to assume that the Database file is not in a valid format or damaged. Therefore we remove it and reset the store.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
```diff
@@ -59,12 +67,29 @@ class SQLiteKeyValueStore {
...
+			removeDatabase()
+		}
```
llanghammer-sap, line 79: Just want to point to this change to get some Feedback on it. When the database can't be accessed after the setup process with setting the encryption key. We have to assume that the Database file is not in a valid format or damaged. Therefore we remove it and reset the store.

src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift
llanghammer-sap: That is true. I added a deleteDatabase() Function to just remove the database files. So no more rekey is needed

ChristianKienle: I will approve this - however I think you will leave data behind. (transaction log files).
They are encrypted but will cause issues when you re-create the database again.
Let's have a cal about this tomorrow - eh today if you don't mind. I can show you how to tell SQLite to delete the file for you.

ChristianKienle (approved):

PR #203, opened 2020-06-04
[Feature] Light/Dark Mode ExposureNotificationSettingViewController

eisvogel:
This PR updates the `ExposureNotificationSettingViewController` to the latest color scheme.

Furthermore some design discrepancies are fixed.

Please limit your review to these view only:
image
image


haosap (approved):

haosap:
Hi @EisvogelSAP ,
approved. I think the biggest change is, some text view was replaced by label view.

EisvogelSAP:
Some typography and the circular view were also adjusted to align with our specs.

src/xcode/ENA/ENA/Source/Views/ENSetting/DescriptionTableViewCell.swift
```diff
@@ -20,20 +20,18 @@ import UIKit
 
 class DescriptionTableViewCell: UITableViewCell, ConfigurableENSettingCell {
 	@IBOutlet weak var titleLabel: ENALabel!
-	@IBOutlet var textView1: UITextView!
-	@IBOutlet var textView2: UITextView!
-	@IBOutlet weak var textView3: UITextView!
+	@IBOutlet weak var label1: UILabel!
+	@IBOutlet weak var label2: UILabel!
+	@IBOutlet weak var label3: UILabel!
```
AdnanZiSAP, line 25: Just a small things. Swift Format kicks out ```weak``` from outlets when it's ran on a file. That's why most if not all outlets in the codebase are without ```weak``` now. Perhaps you could ran Swift Format on this file or just remove it by hand, for the sake of consistency?

AdnanZiSAP (approved):

PR #206, opened 2020-06-04
Feature/set accessibility labels glossary share

steins-code:
add accessibility labels for invite friends page and risk legend (glossary) and slightly change DynamicTableView ImageHeaderView


i336616:
I would set a default value of nil for accessibilityLabel: String?, to handle cases where it's not implemented (it's failing on DynamicTableViewControllerTests)

i336616 (approved):
Happy to approve this, if everyone else is happy with my commit (accessibilityLabel: String? **= nil**)

haosap (approved):

PR #213, opened 2020-06-04
[Feature] risk cell ui polish

aleksandr-tikhonov:
This PR updates ui staff some of cell on home screen. including adapt `ENALabel` and `ENAButton`


eisvogel (commented):
Looks good, however, something seems to be wrong:
image

johannesrohwer (approved):

aleksandr-tikhonov:
> Looks good, however, something seems to be wrong:
> image

Hi @EisvogelSAP, do you mean "Aktualisiert:  - " ?

EisvogelSAP:
Nope, why is it red although it is showing no contacts?

aleksandr-tikhonov:
> Nope, why is it red although it is showing no contacts?

this text depends on `numberRiskContacts` if it's less 1 it has such title.

EisvogelSAP:
Yes, but why is showing high risk then? When there are not contacts the risk should be low.

aleksandr-tikhonov:
> Yes, but why is showing high risk then? When there are not contacts the risk should be low.

there was debug code, I deleted it

eisvogel (approved):

PR #232, opened 2020-06-05
Improved typo

braydofficial:
## Description
Just changed typo from "end to end" to the correct usage "end-to-end".


SebastianWolf-SAP (approved):
Good catch, thanks for fixing. One thing though: If you find such things, probably also have a look at our cwa-website and cwa-documentation repositories. There are certainly additional similar things which wait for their discovery and a fix. Thanks!

PR #237, opened 2020-06-05
[Bug] Fixed all asset properties

eisvogel:
This PR intends to fix all asset properties which were previously lost, thereby correcting asset appearance within the app.

* Fixed all asset properties.
* Set rendering option.
* Removed superfluous dark mode appearances.
* Purged duplicated and unreferenced assets.


melloskitten (approved):
Looks great, thank you!

aleksandr-tikhonov (approved):

PR #245, opened 2020-06-05
Fix/config url

ChristianKienle:


inf2381 (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient+Configuration.swift
```diff
@@ -122,7 +122,7 @@ extension HTTPClient {
 				.appending(
 					"version",
 					apiVersion,
-					"parameters",
+					"configuration",
 					"country",
 					country
```
ChristopherSchmitz, line 127: @ChristianKienle  I guess we need to add "app_config"

ChristopherSchmitz (approved):

PR #267, opened 2020-06-05
[Bug] Silence auto layout warning on home screen.

eisvogel:
This PR introduces a temporary fix which silences the auto layout warning on the home screen.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
+++ b/src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
@@ -65,7 +65,10 @@ class ENAButton: DynamicTypeButton {
 		cornerRadius = 8
 
 		contentEdgeInsets = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
-		heightAnchor.constraint(greaterThanOrEqualToConstant: 50).isActive = true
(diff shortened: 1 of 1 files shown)


haosap (approved):

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -65,7 +65,10 @@ class ENAButton: DynamicTypeButton {
 		cornerRadius = 8
 
 		contentEdgeInsets = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
-		heightAnchor.constraint(greaterThanOrEqualToConstant: 50).isActive = true
+		let heightConstraint = heightAnchor.constraint(greaterThanOrEqualToConstant: 50)
```
aleksandr-tikhonov, line 68: Does somebody know why we need to have 50.0 height constraint for all ENAButtons ?

aleksandr-tikhonov: @SebastianWolf-SAP

EisvogelSAP: I guess because it is defined in the designs guides that a button must have a height of 50pt or more. ;-)

i336616 (approved):
👍

PR #292, opened 2020-06-06
[Bug] Onboarding: Screen pushed twice

AdnanZiSAP:
There is a bug with onboarding screen "Falls Sie positiv getested werden..." being pushed twice after the reset. The cause was that after ```apiMisuse``` error, there was a double completion. That is removed now.

Also, while testing this, I managed to double-tap on the next button at some point, also causing the next screen to be pushed twice. I wasn't able to reproduce it again, but since it happened on 11 Pro, I believe it's better to be safe, since we are supporting 6s/SE. To fix this, I've added a prevention - disabled the button after the tap and enabled it again on the callback.


i336616 (approved):
👍

steins-code (approved):

haosap (approved):

PR #298, opened 2020-06-06
[Feature] Risk calculation in store

aleksandr-tikhonov:


src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
```diff
@@ -77,10 +77,10 @@ enum RiskCalculation {
 		}
 
 		// Precondition 1 - Exposure Notifications must be turned on
-		guard preconditions.isGood else {
-			// This overrides all other levels
-			return .success(.inactive)
-		}
+//		guard preconditions.isGood else {
```
aleksandr-tikhonov, line 80: @inf2381 should we comment code here?

inf2381: @aleksandr-tikhonov I think we should revert these changes and fix the root cause in a separate PR

inf2381 (approved):

inf2381:
@haosap Should we merge this PR or not?

haosap:
Hi @inf2381 ,
Please don't do it.
I put it on hold.  The risk calc is done in real time instead of storing it.

i336616:
> Hi @inf2381 ,
> Please don't do it.
> I put it on hold. The risk calc is done in real time instead of storing it.

If this is to close https://jira.itc.sap.com/browse/EXPOSUREAPP-484, we persist the previousSummary & previousSummaryDate instead (ENExposureDetectionSummaryContainer), covered by this PR: https://github.com/corona-warn-app/cwa-app-ios/pull/293

haosap:
Hi @ChristianKienle ,
Shall we close it now?

PR #337, opened 2020-06-08
[Chore] Add UITesting config

inf2381:
- Enabled UI tests in CI again
- Added new build flag + configuration for UI testing to the project

These changes enable us to use a mocked implementations without having to inject everything like QR scanner view from AppDelegate and without compiling the adjustments into the target binary.
Example usage in the code:
```swift
#if UITESTING
    let foo = MockedBar()
#else
    let foo = Bar()
#endif
```


i336616 (approved):
👍

i336616 (approved):
Yeo, looks good!

PR #362, opened 2020-06-08
Automate legal notices

### LINKED ISSUE #204: Automate legal screen based on THIRD-PARTY-NOTICES
"Like I mentioned above I adapted the same approach which is already used for AppInformationHelp with a Model and ModelData for static strings. But as of now whenever a new dependency is added we have to maintain both places manually:
- THIRD-PARTY-NOTICES
- AppInformationLegalModelData.swift

I can think of a more generic approach to display the licensing info by reading the content of THIRD-PARTY-NOTICES directly but this requires restructuring this file (e.g. JSON or PList). Going one step further, I can think of adding this (and maybe a license check) step to CI so that each build gets an updated and accurate content for THIRD-PARTY-NOTICES file."

_Originally posted by @frnkschmtt in https://github.com/corona-warn-app/cwa-app-ios/pull/111#issuecomment-638419059_

A build step in Xcode would also be sufficient instead of adding it to the CI pipeline.

dfelber:
## Description
Add a fastlane action (`bundle exec fastlane ios update_licenses`) to generate the licenses in `AppInformationViewController+LegalModel.swift` and the `THIRD-PARTY-NOTICES` based on a new `licenses.json`.

closes #204.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -41,15 +41,111 @@ Website:   https://github.com/rwbutler/Connectivity
 License:   MIT License
 
 --------------------------------------------------------------------------------
-Apache License 2.0 (swift-protobuf)
+MIT License
 
+Copyright (c) 2017 Thomas Zoechling
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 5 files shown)


EisvogelSAP:
@dfelber @frnkschmtt please align with #371.

inf2381:
> @dfelber @frnkschmtt please align with #371.

Discussion already done in the related issue #204

inf2381 (approved):

ChristianKienle (approved):

PR #364, opened 2020-06-08
Add missing verticalScrollIndicatorInsets

dfelber:
## Description
Also add bottom insets for the scrollIndicators so they don't vanish behind the footerView.


EisvogelSAP:
Resolved by #425

PR #367, opened 2020-06-09
Disable Certificate Pinning for ENACommunity builds  (closes #358)

### LINKED ISSUE #358: Enable Developers to disable Certificate Pinning
#301 will enable certificate pinning on all environments. We should find a way to still be able to disable pinning for development purposes since it make debugging the HTTP layer terribly inconvenient.

At the same time we have to keep security considerations in mind. There should not be any hidden setting or runtime capabilities to disable pinning. We should maybe consider having a separate target for developers. This way pinning becomes disabled/enabled at compile time rather than runtime.

sebastianpicklum:
## Description
Closes #358:
"_We should find a way to still be able to disable pinning for development purposes since it make debugging the HTTP layer terribly inconvenient._"

This PR will disable certificate pinning for ENACommunity builds at compile time using macros. Production builds don't contain any code that would be able to disable certificate pinning.

## Changes
* Added DISABLE_CERTIFICATE_PINNING to the  ENACommunity scheme
* CoronaWarnURLSessionDelegate() is used if DISABLE_CERTIFICATE_PINNING isn't defined

### DIFF
--- a/src/xcode/ENA/ENA.xcodeproj/project.pbxproj
+++ b/src/xcode/ENA/ENA.xcodeproj/project.pbxproj
@@ -2394,6 +2394,7 @@
 					"DEBUG=1",
 					"$(inherited)",
 					"SQLITE_HAS_CODEC=1",
+					"DISABLE_CERTIFICATE_PINNING=1",
 				);
 				INFOPLIST_FILE = ENA/Resources/Info.plist;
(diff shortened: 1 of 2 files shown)


inf2381 (approved):

ChristianKienle (approved):

ChristianKienle:
Thanks! :)

sebastianpicklum:
My pleasure... 

@inf2381 Can you merge this manually or start circleci?

PR #379, opened 2020-06-09
[Feature]Update the PROD url.

haosap:
## Purpose
Change the placeholders of old URLs to the correct ones. In this PR, the urls of production system are updated now.

## Impact
If the user doesn't launch the app via the url schema, the production url will be by default chosen.


inf2381 (approved):

ChristianKienle:
We are getting closer and closer. 🥳

ChristianKienle (approved):

PR #401, opened 2020-06-09
Risk Level Interval Cell Dynamic Color

larsaugustin:
In risk level collection view cell, the detection interval uses the wrong dynamic color. Instead of using the text contrast color, it uses a system gray. That's why the label's color doesn't match other label's colors in dark mode:

|Current|Expected|
|---|---|
|Current|Expected|


eisvogel (approved):

ChristianKienle (approved):

i336616 (approved):
👍

PR #406, opened 2020-06-09
[Fix] Exposure Submission Reference Cycles

johannesrohwer:
## Description

 This PR removes some reference cycles to avoid stale VCs remaining in memory after they have been dismissed.

## Checklist

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionNavigationController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionNavigationController.swift
@@ -161,21 +161,21 @@ class ExposureSubmissionNavigationController: UINavigationController, UINavigati
 		applyDefaultRightBarButtonItem(to: topViewController)
 		applyNavigationBarItem(of: topViewController)
 
-		keyboardWillShowObserver = NotificationCenter.default.addObserver(forName: UIResponder.keyboardWillShowNotification, object: nil, queue: nil) { notification in
-			self.isKeyboardHidden = false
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionNavigationController.swift
```diff
@@ -161,21 +161,21 @@ class ExposureSubmissionNavigationController: UINavigationController, UINavigati
...
+			self?.keyboardWindowFrame = (notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue
+			self?.updateBottomSafeAreaInset(animated: true)
 		}
 
-		keyboardWillChangeFrameObserver = NotificationCenter.default.addObserver(forName: UIResponder.keyboardWillChangeFrameNotification, object: nil, queue: nil) { notification in
-			self.keyboardWindowFrame = (notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue
-			self.updateBottomSafeAreaInset(animated: true)
+		keyboardWillChangeFrameObserver = NotificationCenter.default.addObserver(forName: UIResponder.keyboardWillChangeFrameNotification, object: nil, queue: nil) { [weak self] notification in
+			self?.keyboardWindowFrame = (notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue
+			self?.updateBottomSafeAreaInset(animated: true)
```
EisvogelSAP, line 178: Technically not needed, as the observers will be discarded once the view disappears anyway.
Also you may want to use `guard let self = self else { return }` instead of optional changing.

eisvogel (approved):

thomasaugsten (approved):

PR #408, opened 2020-06-09
Update spinner / loading indicators on exposure submission

melloskitten:
I changed the spinners according to the design feedback, now we have spinners as part of the ENAButton. I have included these in the TANInput view and the TestResult view. 

Note: the spinning behaviour in the QRCodeScan view is supposed to be the same as before, spinning in the middle of the Overview screen. (Checked this with the designers as well.)


src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -29,6 +29,7 @@ class ENAButton: DynamicTypeButton {
 	override var isHighlighted: Bool { didSet { applyHighlight() } }
 
 	private var highlightView: UIView!
+	private var spinner: UIActivityIndicatorView!
```
EisvogelSAP, line 32: Should be called activityIndicator, please

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -80,6 +81,19 @@ class ENAButton: DynamicTypeButton {
 		highlightView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
 		addSubview(highlightView)
 
+		// MARK: - Add spinner for loading state.
+		
+		spinner = UIActivityIndicatorView(style: .medium)
+		spinner.translatesAutoresizingMaskIntoConstraints = false
+		spinner.isUserInteractionEnabled = false
+        spinner.isExclusiveTouch = false
```
EisvogelSAP, line 89: no spaces, please

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -80,6 +81,19 @@ class ENAButton: DynamicTypeButton {
...
+		addSubview(spinner)
+		if let title = titleLabel {
+			spinner.leadingAnchor.constraint(greaterThanOrEqualTo: leadingAnchor, constant: 8).isActive = true
+			title.leadingAnchor.constraint(greaterThanOrEqualTo: spinner.trailingAnchor, constant: 8).isActive = true
```
EisvogelSAP, line 93: Should be `equalTo`. Currently the layout is ambiguous.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
EisvogelSAP, line 89: Not required, when `isUserInteractionEnabled` is false.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -108,6 +122,18 @@ class ENAButton: DynamicTypeButton {
 	private func applyHighlight() {
 		highlightView.isHidden = !isHighlighted
 	}
+
+	func startSpinner() {
```
EisvogelSAP, line 126: Please expose `isLoading` property to like `isTransparent` and `isInverted` for use in InterfaceBuilder.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -108,6 +122,18 @@ class ENAButton: DynamicTypeButton {
...
+		isEnabled = false
+		spinner.startAnimating()
+		applyStyle()
```
EisvogelSAP, line 129: `applyStyle` is redundant at this point, since it is already being called with `isEnabled`.

eisvogel (changes requested):
Looks good, however:
* Layout is ambiguous and needs to be fixed
* `isLoading` property should be used to expose the activity indicator
* the activity indicator should be shown/hidden in `applyStyle` according to `isLoading`
* the color of the indicator should be set to foreground

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -80,6 +81,19 @@ class ENAButton: DynamicTypeButton {
...
+		spinner = UIActivityIndicatorView(style: .medium)
+		spinner.translatesAutoresizingMaskIntoConstraints = false
+		spinner.isUserInteractionEnabled = false
+        spinner.isExclusiveTouch = false
+		addSubview(spinner)
+		if let title = titleLabel {
+			spinner.leadingAnchor.constraint(greaterThanOrEqualTo: leadingAnchor, constant: 8).isActive = true
+			title.leadingAnchor.constraint(greaterThanOrEqualTo: spinner.trailingAnchor, constant: 8).isActive = true
+			spinner.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+		}
```
EisvogelSAP, line 95: default layout should be centered, if the label doesn't exist.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -80,6 +82,20 @@ class ENAButton: DynamicTypeButton {
...
+		activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+		activityIndicator.isUserInteractionEnabled = false
+		addSubview(activityIndicator)
+		if let title = titleLabel {
+			activityIndicator.leadingAnchor.constraint(greaterThanOrEqualTo: leadingAnchor, constant: 8).isActive = true
+			title.leadingAnchor.constraint(equalTo: activityIndicator.trailingAnchor, constant: 8).isActive = true
+			activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+		} else {
+			activityIndicator.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+			activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
```
EisvogelSAP, line 96: Sorry, one more thing. By default the button as a height of 50 or more, but the width can collapse.
So we need 8pt margin to left and right, please.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -102,6 +118,13 @@ class ENAButton: DynamicTypeButton {
 			setTitleColor(style.disabledForegroundColor.withAlphaComponent(0.5), for: .disabled)
 		}
 
+		if isLoading {
+			activityIndicator.color = style.disabledForegroundColor.withAlphaComponent(0.5)
```
EisvogelSAP, line 122: Please check if enabled and set either `foregroundColor` or `backgroundColor`. Otherwise this will result in wrong appearance:
```
isLoading = true
isEnabled = true
```

eisvogel (approved):

AdnanZiSAP:
I'm on this... (review).

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -108,6 +114,52 @@ class ENAButton: DynamicTypeButton {
...
+
+		if let title = titleLabel {
+			title.leadingAnchor.constraint(equalTo: activityIndicator.trailingAnchor, constant: 8).isActive = true
+			activityIndicator.leadingAnchor.constraint(greaterThanOrEqualTo: self.leadingAnchor, constant: 8).isActive = true
+			activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+		} else {
+			activityIndicator.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+			activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+			activityIndicator.leadingAnchor.constraint(greaterThanOrEqualTo: self.leadingAnchor, constant: 8).isActive = true
+			self.trailingAnchor.constraint(greaterThanOrEqualTo: activityIndicator.trailingAnchor, constant: 8).isActive = true
```
AdnanZiSAP, line 140: Are these two constraints really needed if we have ```centerXAnchor``` and ```centerXAnchor``` already?

AdnanZiSAP:
I've merged this into my brach ```feature/ena-button-accessibility```, to check how this plays with buttons resizing, and it looks good when the indicator is presented - the button resizes to show indicator and full, not truncated, text, and the indicator is displayed in the middle of its Y axis, which is spot on. But after the indicator disappears, the space it was holding is left, so the button is still resized, although the text can fit into one line. IMHO the indicator shouldn't still take the space in the view after it's gonne. 

Here's how it looks before it's tapped:

IMG_14ECFC2EC2E1-1

And here's how it looks after the indicator is shown: 

IMG_93184C0F0EB9-1

This space left of the button title is a space left by the indicator. It's a bit hard to catch a screenshot for the few moments the indicator is shown, but I guess that is not necessary here.

AdnanZiSAP (changes requested):
The indicator should be removed after it's not shown anymore.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
EisvogelSAP: Yes, otherwise the button may collapse in width.

EisvogelSAP:
Actually the indicator is removed, if disabled: https://github.com/corona-warn-app/cwa-app-ios/pull/408/files#diff-e01a747b5261c35cd2f9f4707c6261d3R120

Mayer the button needs some trigger to update the label size again?

ChristianKienle (approved):

PR #410, opened 2020-06-09
[Bug] Fix tap target size on home screen

eisvogel:
This PR fixes the tap target size of the info button on the home screen. It also has some minor adjustments/improvements.


johannesrohwer (approved):
looks good to me

i336616 (approved):
👍

PR #419, opened 2020-06-09
Fix/linter

ChristianKienle:


johannesrohwer (commented):
There are many white space changes with lines that contain single tabs- is this intentional?

ChristianKienle:
> There are many white space changes with lines that contain single tabs- is this intentional?

Somehow Xcode does not to seem to honor my settings. :(

inf2381 (approved):

SebastianWild (approved):

PR #427, opened 2020-06-09
[Bug] Privacy protection screen

eisvogel:
This PR addresses the issue of the privacy protection screen appearing, when interrupted by system dialogs such as phone call prompts as show below.

|Before|After|
|:--|:--|
|image|image|


SebastianWild (approved):
Side effect here is that the privacy screen does not show up immediately when triggering the app switcher...but the much better thing is that the privacy screen of course does not show up anymore for action sheets and swiping down the control center 👍

steins-code (approved):

PR #436, opened 2020-06-10
Risk cell chevron icon tint color

jonahschueller:
I changed the tint color of the risk cell. The previous tint color did not contrast enough with the background color of the cell.

The tint color now is ENA Text Contrast Color. It's basically just white but to make the icon automatically adapt to any future color scheme changes.

Before:
</img>

After:
</img>


EisvogelSAP:
Hey @Bohne9, thanks for your changes. Unfortunately this had already been fixed at some point and needs some rework from our side altogether. The chevron color will be an icon with white background in that case. Will put this PR on hold until I can work on this.

ChristianKienle:
@EisvogelSAP I will close this for now (house keeping). Feel free to re-open this in case you need to.

@Bohne9 Thank you very much for this PR. As @EisvogelSAP already stated, this has been fixed in the meantime. I think that the new chevron with a background color is also a bit more legible than the one without a background color…

PR #439, opened 2020-06-10
[Feature] Make Buttons resize with accessibility settings changed (closes EXPOSUREAPP-994)

AdnanZiSAP:
# What this PR does

This PR introduces updated ```ENAButton```, which now handles resizing on font changes. There're also some other updates to mitigate what prevent ```ENAButton``` to do this:

- ```ENACloneButton``` is removed - replaced with ```ENAButton```. This affects "Aktualisieren" button on Home screen. To my eyes looks just like in the mockups, but double-check is welcome.
- In submission flow, there are changes, because the lower button, which is not present all the time, wasn't resizing correctly. It now does, with one exception (below).

## Known Limitations

The view that is the container for two buttons at the bottom of the screen in submission flow, along with buttons in it, doesn't resize properly in cases when both buttons are shown and user moves font slider up and down. This only happens while the slider is moved exactly when that screen is open - all cases when the slider is moved once during the app usage, and left that way, are working correctly - when user in navigated on that screen, no matter about the font size, it always behaves correctly. I decided that at this point of time fixing that edge case is not worth it.

# The looks

No time for that this time sorry. :) You'd have to run the app on your own and verify it's looking as it should. :)


src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -108,6 +106,16 @@ class ENAButton: DynamicTypeButton {
 	private func applyHighlight() {
 		highlightView.isHidden = !isHighlighted
 	}
+
+	private func setHeight() {
+		let heightConstraint = heightAnchor.constraint(greaterThanOrEqualToConstant: 50)
```
aleksandr-tikhonov, line 111: Do we have some option don't use constraints inside the ENAButton? Maybe `contentInsets` or similar ? It's hard to predict how it would work in `UIStackView`.

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
```diff
@@ -108,6 +106,16 @@ class ENAButton: DynamicTypeButton {
...
+		heightConstraint.priority = .defaultHigh
+		heightConstraint.isActive = true
+
+		if let titleLabel = titleLabel {
+			addConstraint(NSLayoutConstraint(item: self, attribute: .height, relatedBy: .equal, toItem: titleLabel, attribute: .height, multiplier: 1, constant: 0))
```
aleksandr-tikhonov, line 116: please use _anchor_ constraint api

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionNavigationController.swift
```diff
@@ -157,6 +157,13 @@ class ExposureSubmissionNavigationController: UINavigationController, UINavigati
 		exposureSubmissionService
 	}
 
+	override func viewDidLayoutSubviews() {
+		super.viewDidLayoutSubviews()
+
+		updateBottomSafeAreaInset()
+		view.layoutIfNeeded()
```
aleksandr-tikhonov, line 164: Is it good idea to use it here? It could lead to stackoverflow issue in this method.

aleksandr-tikhonov (changes requested):

src/xcode/ENA/ENA/Source/Views/BaseElements/ENAButton.swift
AdnanZiSAP: This is the only option to make the button title resize on system font changes that I could find that works. And when I implemented this originally, on Invite Friends screen, I lost hours trying different options. 

When I originally implemented this, I also had it on Reset screen, in a stack view, and it worked. It's still the case, as far as I could see.

thomasaugsten (approved):

ChristianKienle:
@aleksandr-tikhonov could you please revisit your change request here?

aleksandr-tikhonov:
> @aleksandr-tikhonov could you please revisit your change request here?

yep, 15 mins please

aleksandr-tikhonov (approved):

haosap:
Hi @EisvogelSAP ,
is there any reason to put it onhold?

melloskitten:
@EisvogelSAP @AdnanZiSAP Are there any news regarding this PR?

AdnanZiSAP:
> @EisvogelSAP @AdnanZiSAP Are there any news regarding this PR?

AFAIK, It's waiting for the submission flow PR to be merged first. I will assign it to @EisvogelSAP since he took this over from me.

EisvogelSAP:
Jep, will take of this one tomorrow.

EisvogelSAP:
Addressed in #752 and #753.

PR #446, opened 2020-06-10
[Chore] Unit tests for SecureStore

### LINKED ISSUE #411: [Tests] Unit tests for SecureStore

inf2381:
Fixes #411


src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
```diff
@@ -0,0 +1,119 @@
...
+
+	func testResultReceivedTimeStamp_Success() {
+		XCTAssertNil(store.testResultReceivedTimeStamp)
+		store.testResultReceivedTimeStamp = Int64.max
+		XCTAssertEqual(store.testResultReceivedTimeStamp, Int64.max)
+		store.testResultReceivedTimeStamp = Int64.min
+		XCTAssertEqual(store.testResultReceivedTimeStamp, Int64.min)
+	}
+
+	func lastSuccessfulSubmitDiagnosisKeyTimestamp_Success() {
```
llanghammer-sap, line 38: Xcode doesn't seem to recognize functions in Unit Tests without a prefix of 'test'. This applies to all following functions. 

If this was on purpose feel free to resolve this conversation

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
```diff
@@ -0,0 +1,119 @@
...
+		XCTAssertNil(store.lastSuccessfulSubmitDiagnosisKeyTimestamp)
+		store.lastSuccessfulSubmitDiagnosisKeyTimestamp = Int64.max
+		XCTAssertEqual(store.lastSuccessfulSubmitDiagnosisKeyTimestamp, Int64.max)
+		store.lastSuccessfulSubmitDiagnosisKeyTimestamp = Int64.min
+		XCTAssertEqual(store.lastSuccessfulSubmitDiagnosisKeyTimestamp, Int64.min)
+	}
+
+	func numberOfSuccesfulSubmissions_Success() {
+		XCTAssertNil(store.numberOfSuccesfulSubmissions)
```
llanghammer-sap, line 47: Currently returns 0 not nil.

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
```diff
@@ -0,0 +1,119 @@
...
+		store.numberOfSuccesfulSubmissions = Int64.max
+		XCTAssertEqual(store.numberOfSuccesfulSubmissions, Int64.max)
+		store.numberOfSuccesfulSubmissions = Int64.min
+		XCTAssertEqual(store.numberOfSuccesfulSubmissions, Int64.min)
+	}
+
+	func initialSubmitCompleted_Success() {
+		XCTAssertNil(store.initialSubmitCompleted)
```
llanghammer-sap, line 55: Currently returns _false_ and not _nil_

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
```diff
@@ -0,0 +1,119 @@
...
+		store.initialSubmitCompleted = true
+		XCTAssertEqual(store.initialSubmitCompleted, true)
+		store.initialSubmitCompleted = false
+		XCTAssertEqual(store.initialSubmitCompleted, false)
+	}
+
+	func registrationToken_Success() {
+		XCTAssertEqual(store.registrationToken, "")
```
llanghammer-sap, line 63: Currently returns _nil_ and not _""_

llanghammer-sap (changes requested):
Currently only the first test seems to run, as Xcode seems to exspect a prefix of "test" for all functions. I don't think that was on purpose.

There are also so failing test, where we maybe should have a look at the exspected behaviour of the store. I marked all of them with the value they are currently returning.

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
inf2381: Fixed

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
inf2381: Fixed

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
inf2381: Fixed

src/xcode/ENA/ENA/Source/Workers/__tests__/StoreTests.swift
inf2381: Fixed

ChristianKienle (approved):
Super cool tests. We should add a few more tests though. Especially the following should also be covered:

We store the summary + the tracing history inside the store. This means that we can only ever make `Codable`-compatible changes to those structs. Thus we should use a store file - add it to our tests as a fixture and then try to modify it in our tests using the current coding. This test ensures that old "data" can still be read/written using new code.

llanghammer-sap (approved):

ChristianKienle (changes requested):
whops

ChristianKienle (approved):

PR #448, opened 2020-06-10
[Security] Explicitly set min TLS protocol version to prevent replay attack 

### LINKED ISSUE #2: [BSI][0.5-alpha] No Pinning
Rating: Medium
 
Description:
Certificate or public key pinning is when an app communicates with entites with specific TLS Certificates/Public Key. Any other communication is denied by the app, even if the TLS Certificate is signed by a trusted CA. This provides extra protection against Man in the Middle (MitM) attacks.

iOS app does not implement this secure measure, and this increases the possibility of a MitM attack. 

---
Internal Tracking ID: EXPOSUREAPP-2517

sebastianpicklum:
## Description
TLSv12 and above support Message Authentication Codes which prevents replay attacks 
See #2 

This PR set configures URLSessionConfiguration explicitly to use TLSv12 or newer.

See https://tools.ietf.org/html/rfc5246#section-5


ChristianKienle:
@ChristopherSchmitz could you please look into this PR?

ChristopherSchmitz:
@sebastianpicklum thanks for opening the report. On iOS we consider the operation system default as secure and will not introduce a dedicated min TLS version on the app for now.

PR #500, opened 2020-06-11
[Bug] Trait headers on risk legend

eisvogel:
This PR adds missing header trait on risk legend. Furthermore, the global navigation subtitle was changed to header.


i336616 (approved):
👍

haosap (approved):

PR #504, opened 2020-06-11
removed dev menu

ChristianKienle:


src/xcode/ENA/ENA/Source/Developer Menu/DMDeveloperMenu.swift
```diff
@@ -15,10 +15,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
+#if !RELEASE
```
aleksandr-tikhonov, line 18: Why not just `DEBUG`?

ChristianKienle: We still have other configurations like `Community`

haosap (approved):

PR #505, opened 2020-06-11
[Bug] Risk cell hide label

aleksandr-tikhonov:
Hi,

This PR contains logic of hiding or showing interval label. And also it updates button title according to    state.

### DIFF
--- a/src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.strings
+++ b/src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.strings
@@ -516,6 +516,8 @@
 
 "Home_Risk_Period_Update_Title" = "Updated every %@\U00A0hours";
 
+"Home_Risk_Period_Disabled_Button_Title" = "Updated only every %@\U00A0H";
+
 /* Home Unknown card */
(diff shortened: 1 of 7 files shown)


steins-code (approved):

src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -516,6 +516,8 @@
 
 "Home_Risk_Period_Update_Title" = "Updated every %@\U00A0hours";
 
+"Home_Risk_Period_Disabled_Button_Title" = "Updated only every %@\U00A0H";
```
ChristianKienle, line 519: strings dict? :)

ChristianKienle:
reviewing right now

ChristianKienle (approved):

src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.strings
aleksandr-tikhonov: Do you mean for **0** case ? I think, there will not be **0**, we always will have h > 0

PR #510, opened 2020-06-11
fix: Reload Test Result cell unless it is positive

melloskitten:
This fix removes the error where the user would still have a negative result in their homescreen after scanning a new positive qr code, because it would not reload correctly.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
@@ -382,7 +382,7 @@ extension HomeInteractor {
 extension HomeInteractor {
 	func updateTestResults() {
 		// Avoid unnecessary loading.
-		guard testResult == nil || testResult == .pending else { return }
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -382,7 +382,7 @@ extension HomeInteractor {
 extension HomeInteractor {
 	func updateTestResults() {
 		// Avoid unnecessary loading.
-		guard testResult == nil || testResult == .pending else { return }
+		guard testResult == nil || testResult != .positive else { return }
```
johannesrohwer, line 385: Once the state .positive is reached, the app goes to end-of-life state (and should _not_ update the result further). Seems to be correct now!

johannesrohwer (approved):

inf2381 (approved):

ChristopherSchmitz (approved):

ChristopherSchmitz:
@haosap @ChristianKienle this fixes a bug reported by t systems. please merge into release

PR #511, opened 2020-06-11
[Fix] Slow DynamicTableView cell actions

johannesrohwer:
This PR makes sure that the DynamicTableView actions are dispatched to the main dispatch queue asynchronously. Previously, there were sporadic tap events that were visibly slower than others.

## Checklist


eisvogel (changes requested):
This does not makes sense. The delegate method will always be called on the main thread. This adds an unnecessary on queuing actions. Maybe this avoid the problem you are facing, but it is not fixing the root cause of the issue.

What exactly is the problem you are trying to solve here?

johannesrohwer:
> This does not makes sense. The delegate method will always be called on the main thread. This adds an unnecessary on queuing actions. Maybe this avoid the problem you are facing, but it is not fixing the root cause of the issue.
> 
> What exactly is the problem you are trying to solve here?

Without the proposed changes, the actions in the exposure submission selection screen (QR/Tan/Hotline) sometimes trigger very slow (>0.5s). It is often noticeable for the QR option that triggers an alert. I am also not a fan of adding the DispatchQueue, however it seems to solve this problem.

EisvogelSAP:
Can you give a precise example? The submission flow uses the button for navigation. Only the overview controller relies on `DynamicAction`.

johannesrohwer:
> Can you give a precise example? The submission flow uses the button for navigation. Only the overview controller relies on `DynamicAction`.

1. Open the overview controller
2. Click the QR option -> Alert opens -> Decline
3. Repeat Step 2 a few times until you notice a delay. If you are lucky it will already be the first tap.

This  usually takes a few tries. Christopher approached me with this, me and Sandra were able to reproduce it. When I measured the execution time of the showDisclaimer() method, I did not notice an increase in time even when the alert popup was delayed. I checked SO and found other users with a similar problem when using the didSelectRowAt tableview delegate where this was a proposed solution. (I totally agree its a bit hacky)

EisvogelSAP:
I cannot reproduce this on my side. Let's have quick chat about this tomorrow. The only thing I notice is a delay in responsibility of the tap target after dismissing the alert controller which is by design.

Levino:
What's the suggested regression test for the future? Will someone do step 1., 2. and 3. for every PR in the future and raise a hand if they feel that the delay is back?

MauriceArikoglu (changes requested):
I would refrain from wrapping the calling code into a block for `DispatchQueue.main.async` as this can lead to problems if `func execute(_ action: DynamicAction)` was to be called from somewhere else.
Instead look at the different cases [here](https://github.com/corona-warn-app/cwa-app-ios/blob/3dde0eb5b30193b840029a354183dfe499d2066e/src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift#L146) and wrap the execution of problematic cases instead.
(e.g.: `performSegue`, `block(self)`)

MauriceArikoglu:
> This does not makes sense. The delegate method will always be called on the main thread. This adds an unnecessary on queuing actions. Maybe this avoid the problem you are facing, but it is not fixing the root cause of the issue.
> 
> What exactly is the problem you are trying to solve here?

You are right, the delegate is always called on the main thread, although this does not mean calling the code in an asynchronous manner can't fix the problem he is experiencing.
Especially when the UITableView is operating in the same thread atm.

The solution is not at all reliable though.

As I mentioned in my comment above, looking at the possible values for `action` the problem is probably one of these. Executing an unknown block or selector (without sanitization by the way) can have any amount of side effects.

johannesrohwer:
Thanks for all the additional input. I will do some further investigation and reopen the PR if I come up with new findings.

Levino:
My five cents on this, even though I am not part of the project: Please create an issue first. It will be much more sustainable to discuss how to reproduce this issue and see how many people are affected there instead of several PRs. Use it to verify that there is an issue that other people have too, before you start working on the fix.

This is meant as a general advice for this project and I hope you can take it as a friendly attempt to help. More on this to be found on the [internet](https://eli.thegreenplace.net/2019/how-to-send-good-pull-requests-on-github/).

PR #517, opened 2020-06-11
[Bug] Onboarding - Risikoermittlung - logic flaw (EXPOSUREAPP-1475)

AdnanZiSAP:
# What this PR does.

There was a flaw in onboarding logic. When user activates or deactivates exposure notifications, navigates to the next page, and changes her/his mind, then goes back, the previous screen is the same as it was - there's no indication are ENs activated or not. Furthermore, if navigated again to the next page, no alert asking for permissions is being shown, so she/he might assume that, for example, ENs are enabled when they aren't.

This PR fixes that. There are several additions, as following:

- New illustrations added for the off state.
- Strings updated.
- ~~```ExposureManager``` adjusted to return full ```authorizationStatus``` not only true or false. This was needed for some edge cases. One can argue that the old property could be replaced instead of adding a new one, but that would mean updating a lot of code, especially tests. This could be a task for the future.~~
- The alert when user taps "Nicht aktivieren" on the 3rd screen is modified to show two buttons, as in the design (see screenshots below).
- ```Onboarding``` storyboard is updated with the new view to show the state on the third screen (again, see the screenshots).
- ```OnboardingInfo``` model is updated to contain new stuff for aforementioned state view.
- ```OnboardingViewController``` is modified    
  - UI is now updated in ```ViewWillAppear```, to accept changes in state, also when coming back to the screen.
  - ```UpdateUI``` method contains logical additions for state updates. ~~Again, perhaps not the exact place to do those kind of things, but in the current context IMO it's fine.~~ (Simplified the logic in the meantime.)

## The look

IMG_63E957118272-1

IMG_EF607CFC31BE-1

IMG_9A08F96FA6D0-1

IMG_D93EC72E339E-1

IMG_A3B6A60371C5-1


EisvogelSAP:
Test failing, please double check @AdnanZiSAP

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -192,7 +195,8 @@ final class ENAExposureManager: NSObject, ExposureManager {
 		.init(
 			authorized: type(of: manager).authorizationStatus == .authorized,
 			enabled: manager.exposureNotificationEnabled,
-			status: manager.exposureNotificationStatus
+			status: manager.exposureNotificationStatus,
+			authorizationStatus: type(of: manager).authorizationStatus
```
steins-code, line 199: no need to inject it:  just use ENManager.authorizationStatus where you need it.. its a static class variable

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -134,13 +144,20 @@ final class OnboardingInfoViewController: UIViewController {
...
 
-	private func updateUI() {
+	private func updateUI(exposureManagerState: ExposureManagerState) {
 		guard isViewLoaded else { return }
 		guard let onboardingInfo = onboardingInfo else { return }
 
 		titleLabel.text = onboardingInfo.title
 
-		imageView.image = UIImage(named: onboardingInfo.imageName)
+		let exposureNotificationsNotSet = exposureManagerState.status == .unknown && exposureManagerState.authorizationStatus == .unknown
```
steins-code, line 153: see my comment above ENManager.authorizationStatus ... you can have a look into ENStateHandler for details

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -186,6 +186,11 @@
...
 
+"OnboardingInfo_enableLoggingOfContactsPage_stateHeader" = "Status";
```
steins-code, line 189: Please also add some intermediate english translation as our translation folks might not have time to translate it

steins-code (changes requested):
Looks good, had only a few comments :)

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
AdnanZiSAP: Thanks for the catch.

AdnanZiSAP:
Please don't merge it until I sat it's fine.

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
AdnanZiSAP: Done [here](https://github.com/corona-warn-app/cwa-app-ios/pull/517/commits/cf56b4302acaf9feacb5f652f441d17eed6f10ba).

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
AdnanZiSAP: Are we sure we want to access ```ENManager``` directly here and skip ```ENAExposureManager```?

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
AdnanZiSAP: It's similar to your comment above. Please see my reply above.

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
steins-code: Yes

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
steins-code: Yes please use it directly from the ENManager and not from wrapper class

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
AdnanZiSAP: I mean, at this point I'm totally fine with both options, it's only for the sake of someone else doing the review and asking why I'm skipping ```ENAExposureManager``` and call ```ENManager``` (the Apple framework) directly.

steins-code: We should do it everywhere in the same way and not introduce new complexity ... In statehandler we also call ENManager

AdnanZiSAP: Ok, will change it then. I think we introduced complexity at the point we started calling ```ENManager``` directly, instead of channelling the call through ```ENAExposureManager```. But nevermind, I agree that my PR shouldn't introduced further complexity.

AdnanZiSAP: I'd changed this, but after that I've simplified the logic to adjust it to the current state of development, so this is not needed anymore. 

Please check [here](https://github.com/corona-warn-app/cwa-app-ios/pull/517/commits/e7ed188224d3aff072ed466b6c0b274e8290ee9c).

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
AdnanZiSAP: Based on my previous comment, this is outdated now. Please check [here](https://github.com/corona-warn-app/cwa-app-ios/pull/517/commits/e7ed188224d3aff072ed466b6c0b274e8290ee9c).

aleksandr-tikhonov (commented):
https://github.com/corona-warn-app/cwa-app-ios/blob/e7ed188224d3aff072ed466b6c0b274e8290ee9c/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift#L144

we don't need to pass `completion` here?

AdnanZiSAP:
> Please don't merge it until I say it's fine.

You can disregard this now.

AdnanZiSAP:
> https://github.com/corona-warn-app/cwa-app-ios/blob/e7ed188224d3aff072ed466b6c0b274e8290ee9c/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift#L144
> 
> we don't need to pass `completion` here?

Not sure that I understand the question. I didn't touch that part, I only added one additional button to the dialog.

aleksandr-tikhonov (approved):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -134,13 +144,20 @@ final class OnboardingInfoViewController: UIViewController {
...
 
-	private func updateUI() {
+	private func updateUI(exposureManagerState: ExposureManagerState) {
 		guard isViewLoaded else { return }
 		guard let onboardingInfo = onboardingInfo else { return }
 
 		titleLabel.text = onboardingInfo.title
 
-		imageView.image = UIImage(named: onboardingInfo.imageName)
+		let exposureNotificationsNotSet = exposureManagerState.status == .unknown
```
steins-code, line 153: exposureManagerState.status == .unknown is false doesnt implicate exposureNotificationsNotSet ... See more details here: https://developer.apple.com/documentation/exposurenotification/enstatus?language=objc

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -134,13 +144,20 @@ final class OnboardingInfoViewController: UIViewController {
...
+		let exposureNotificationsEnabled = exposureManagerState.enabled
+		let exposureNotificationsDisabled = !exposureNotificationsEnabled && !exposureNotificationsNotSet
```
steins-code, line 155: This logic is wrong... if you want to the state if risk identification is turned active or not looking only on enabled and ENStatus different from unknown is not enough.. Consider the ENStatus Bluetooth off ... then it still shows active tracing but tracing is clearly not active in this case

steins-code (changes requested):
please make use of authorizationStatus

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
AdnanZiSAP: This point is resolved in a call. Just to summarise, to determine should two buttons be shown in the 3rd onboarding step, with user getting the chance to decide whether to accept it or not, or one button should be shown, along with the state view, and taking into the consideration that after the reset Exposure Manager is not activated yet, the only way to determine and differentiate this is ```exposureManagerState.status == .unknown```.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
AdnanZiSAP: Based on the discussion we have on the phone, this is fine in most cases. There's one edge case that I found during testing while playing with bluetooth toggle is, that when user activates the tracing during onboarding, and then in the next step he goes to settings and turns bluetooth off, then goes back, he should be asked to enable bluetooth (or to decline), and he wasn't. 

Now it's fixed with [this commit](https://github.com/corona-warn-app/cwa-app-ios/pull/517/commits/73e8946b88b50ec123c265a121d90691ec23db21).

steins-code: Yes, for now its fine, still I find the variable naming not the best :D ... but I'll approve it now because the functionality is correct now and we need to support reset functionality in some way ... I guess it might need some refactoring in the future ;)

steins-code (approved):

steins-code:
https://jira.itc.sap.com/browse/EXPOSUREAPP-1470
https://jira.itc.sap.com/browse/EXPOSUREAPP-1475

AdnanZiSAP:
Updated the branch and tested to verify that it still works as expected. 

This one is ready to be merged.

AdnanZiSAP:
On hold until today's UX/UI meeting.

AdnanZiSAP:
Removed On hold, UX/UI confirmed the design.

PR #518, opened 2020-06-11
[Lint] Turn off unnecessary rules in test code

Jeehut:
## Description
This is the second follow-up smaller PR to https://github.com/corona-warn-app/cwa-app-ios/pull/104. Together with the other PRs I will be posting, all SwiftLint warnings should be fixed to allow us to enable the `--strict` mode of SwiftLint on the CI to prevent any new warnings getting merged in the future.

## Related PRs
https://github.com/corona-warn-app/cwa-app-ios/pull/516, https://github.com/corona-warn-app/cwa-app-ios/pull/522, https://github.com/corona-warn-app/cwa-app-ios/pull/527, https://github.com/corona-warn-app/cwa-app-ios/pull/531, https://github.com/corona-warn-app/cwa-app-ios/pull/534.

_Note that all these PRs are completely independent and **can be merged in any order**._

PR #516, opened 2020-06-11
[Lint] Fixes some trivial SwiftLint warnings

Jeehut:
## Description
This is the first follow-up smaller PR to https://github.com/corona-warn-app/cwa-app-ios/pull/104. Together with the other PRs I will be posting, all SwiftLint warnings should be fixed to allow us to enable the `--strict` mode of SwiftLint on the CI to prevent any new warnings getting merged in the future.

## Related PRs
https://github.com/corona-warn-app/cwa-app-ios/pull/518, https://github.com/corona-warn-app/cwa-app-ios/pull/522, https://github.com/corona-warn-app/cwa-app-ios/pull/527, https://github.com/corona-warn-app/cwa-app-ios/pull/531, https://github.com/corona-warn-app/cwa-app-ios/pull/534.

_Note that all these PRs are completely independent and **can be merged in any order**._

PR #104, opened 2020-06-01
[Technical] Fix SwiftLint warnings & configure AnyLint checks

Jeehut:
- [x] Fixes all SwiftLint warnings that were trivially fixable without bigger refactoring (to reduce merge conflicts).
- [x] Configured SwiftFormat as a run-script and ran it to fix many formatting inconsistencies.
- [x] Added some custom Regex based checks using [AnyLint](https://github.com/Flinesoft/AnyLint) (the better SwiftLint custom rules alternative). I've configured a few checks (most of them with autocorrection) and corrected any issues.

You might want to add the `anylint` command to your CI pipeline, but you can also just keep it in the repo to correct some issues manually from time to time if you don't trust the checks yet.

Note: I will not install proprietary software from Microsoft just to ping you about this PR. I think you should improve your workflow in that regard. Also I couldn't find a link to the mentioned Microsoft Teams channel in the README. But this PR is ready for review.

PR #522, opened 2020-06-11
[Lint] Reintroduce rule identifier_name & fix warnings

Jeehut:
## Description
This is the third follow-up smaller PR to https://github.com/corona-warn-app/cwa-app-ios/pull/104. Together with the other PRs I will be posting, all SwiftLint warnings should be fixed to allow us to enable the `--strict` mode of SwiftLint on the CI to prevent any new warnings getting merged in the future.

In this specific PR, I've turned on the `identifier_name` rule which is enabled by default in SwiftLint for a reason (and was turned off for some reason) and the many cases where it was violated show really nicely why the rule is a good rule that should be kept turned on. Many variables were not really speaking and too short, this rule will prevent that in the future.

I've added a few exceptions for the common 2-character English words in development within the configuration (like `id` and `db`) but the many appearances of `vc`, `nc`, `i` and other too short and not very concise names were fixed.

Note also that I fixed variable names where for example `keys` was used for an Optional and `_keys` for the strong typed variable. I used speaking names in such cases like `keysOptional` and `keys`, removing the need for the leading underscore, which is not very Swifty (more Obj-C like).

## Related PRs
https://github.com/corona-warn-app/cwa-app-ios/pull/516, https://github.com/corona-warn-app/cwa-app-ios/pull/518, https://github.com/corona-warn-app/cwa-app-ios/pull/527, https://github.com/corona-warn-app/cwa-app-ios/pull/531, https://github.com/corona-warn-app/cwa-app-ios/pull/534.

### DIFF
--- a/src/xcode/ENA/.swiftlint.yml
+++ b/src/xcode/ENA/.swiftlint.yml
@@ -196,7 +196,6 @@ opt_in_rules:
 # Disable rules which are in the default set
 disabled_rules:
   - type_name
-  - identifier_name
   - redundant_string_enum_value # disabled because it conflicts with explicit_enum_raw_value
 
 # Included paths (disables --path option)
@@ -214,6 +213,15 @@ reporter: "xcode"
 
 
 ### Rule specific configuration
+
+identifier_name:
+  max_length: 60
+  excluded:
+    - db
+    - id
+    - ok
+    - on
+    - to
 
 line_length:
   warning: 500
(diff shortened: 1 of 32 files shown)


ChristianKienle:
Thank you very much for this PR. I will put this on hold for the next couple of days because we are in a critical phase in the project at the moment. I hope you understand. We will come back to you once this phase is over.

Jeehut:

PR #527, opened 2020-06-11
[Lint] Reconfigure rule line_length & fix warnings

Jeehut:
## Description
This is the fourth follow-up smaller PR to https://github.com/corona-warn-app/cwa-app-ios/pull/104. Together with the other PRs I will be posting, all SwiftLint warnings should be fixed to allow us to enable the `--strict` mode of SwiftLint on the CI to prevent any new warnings getting merged in the future.

In this specific PR, I've turned on the due to it's loose configuration effectively turned off rule `line_length` where I really don't understand why it was effectively turned off. As you can see in the many fixe I needed to make, already many lines really got out of hand and were way too long. It is a very important rule for any programming language.

Note that I've chosen a `max_length` of 160 and not the default 120 because nowadays monitors are widescreen and have a high resolution, also in Swift development brevity isn't necessarily a goal in variable and class naming, conciseness is much more important which can sometimes lead to some longer lines. I think 160 is a good compromise. If you disagree, I think merging my PR still makes sense as it's a good step forward from 500 and 120 can be reached in a subsequent PR if needed.

## Related PRs
https://github.com/corona-warn-app/cwa-app-ios/pull/516, https://github.com/corona-warn-app/cwa-app-ios/pull/518, https://github.com/corona-warn-app/cwa-app-ios/pull/522, https://github.com/corona-warn-app/cwa-app-ios/pull/531, https://github.com/corona-warn-app/cwa-app-ios/pull/534.

### DIFF
--- a/src/xcode/ENA/.swiftlint.yml
+++ b/src/xcode/ENA/.swiftlint.yml
@@ -215,11 +215,7 @@ reporter: "xcode"
 
 ### Rule specific configuration
 
-line_length:
-  warning: 500
-  error: 1000
-  ignores_function_declarations: true
-  ignores_comments: true
+line_length: 160
 
 trailing_whitespace:
   ignores_empty_lines: true
--- a/src/xcode/ENA/.swiftlint.yml
+++ b/src/xcode/ENA/.swiftlint.yml
@@ -201,7 +201,9 @@ disabled_rules:
 
 # Included paths (disables --path option)
 included:
-  - ENA
+  - ENA/Source
+  - ENATests
+  - ENAUITests
 
 # Excluded paths (takes precedence over 'included')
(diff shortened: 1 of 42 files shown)


ChristianKienle:
Thank you very much @Jeehut. I am putting this on hold for now due to our bug fixing phase. We will remove the on hold label once this phase is over.

Jeehut:
@ChristianKienle Thanks for the info, do you have a time schedule for how long the bug fixing phase will take? I'm asking this because I think that fixed warnings may actually help a team be faster rather than consuming more time, especially on open source projects.

But of course it may make sense to put this on hold for a few dozen of hours, I just hope it's not too long as merging this fast will prevent unnecessary merge conflicts and help preventing new violations to be introduced.

Jeehut:

PR #536, opened 2020-06-11
On exposure detection error, clear package store

SebastianWild:
## Description

It could be that an error occurs during the exposure detection process (ex. internal signature validation fails). We should handle these errors when they are reported by the system and clear the `DownloadedPackagesStore`


ChristianKienle:
no longer needed. cc @mlenkeit <3

PR #538, opened 2020-06-11
[Bug] Remove invalid separator line in risk detail screen

eisvogel:
This PR removes a redundant line is the risk detail screen.

Take not on the first line before "Letzte Risiko-Ermittlung". This change applies to both disabled and outdated state.

|Before|After|
|:--|:--|
|image|image|


i336616 (approved):
👍

haosap (approved):

ChristianKienle (approved):

PR #544, opened 2020-06-11
[Feature] Add support for reduced transparency accessibility setting

eisvogel:
This PR adds support for the reduced transparency accessibility setting for footer views with buttons (used during onboarding, in the risk detail screen and the submission flow).

Basically the translucency of the footer view is replaced by the defined ENA background color.

|Normal Appearance|Reduced Transparency|
|:--|:--|
|image|image|


johannesrohwer (commented):
works as expected!

thomasaugsten (approved):

johannesrohwer (approved):

PR #552, opened 2020-06-12
Show Alert on exposure detection error

SebastianWild:
## Description

For debugging & troubleshooting purposes it might be useful to receive an alert with a message & trace when exposure detection fails.

Not sure if this PR will be used yet, this is to be discussed.


thomasaugsten:
I think it is a good idea

ChristianKienle (approved):

PR #586, opened 2020-06-12
removed unused code, added tests, slight api improvements

ChristianKienle:


ChristianKienle:
removed on hold because the target branch is **next**.

inf2381 (approved):

Levino:
How can you be sure that these changes to not break something? What is the test coverage on the files touched in this PR? Also I kind of think the changes solve multiple different issues but might even change behaviour and are not only refactoring. The PR title is as telling as "Doing some work". Has the reviewer really checked out the code run and tested the app thoroughly?

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
```diff
@@ -41,7 +41,7 @@ final class HomeLowRiskCellConfigurator: HomeRiskLevelCellConfigurator {
 		self.detectionInterval = detectionInterval
 		super.init(
 			isLoading: false,
-			isButtonEnabled: detectionMode == .manual && manualExposureDetectionState == .possible,
+			isButtonEnabled: manualExposureDetectionState == .possible,
```
Levino, line 44: For example this button. So it used to be enabled when `detectionMode` was the same as the value of `.manual` (which probably is a constant), so meaning "detection mode is manual" (instead of "automatic"). Where does the requirement come from that now this button should be disabled when "detection mode is manual or automatic"?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/__tests__/RiskProviderTests.swift
```diff
@@ -133,39 +99,42 @@ final class RiskProviderTests: XCTestCase {
 
 		let exposureSummaryProvider = ExposureSummaryProviderMock()
 
-		let expectThatNoSummaryIsRequested = expectation(description: "expectThatNoSummaryIsRequested")
-		expectThatNoSummaryIsRequested.isInverted = true
+		let detectionRequested = expectation(description: "expectThatNoSummaryIsRequested")
```
Levino, line 102: Please explain this. Why do you change the variable name but not the string value for the description? Is this the same? Why add this mental mapping?

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
ChristianKienle: @levino Thank you very much for this question. I think the change above has no effect on the actual UI. `manualExposureDetectionState` now returns `nil` in case `detectionMode` is `.automatic`. Thus `detectionMode == .manual` can be removed as a condition. This PR also comes with tests.

SebastianWild (approved):

PR #594, opened 2020-06-13
fix: fix tan validation on cancelled swipe-back (closes #591)

### LINKED ISSUE #591: [UI] TAN Entry view breaks when gestures are cancelled
## Describe the bug
Just a quick reminder that in the latest development build, the TAN entry view submit button still breaks when doing the back swipe gesture and cancelling it half way. The button then gets enabled and clickable by the user even without a valid TAN entered. Another bug regarding the TAN entering view is when you swipe down to close the pop up view and cancel it half way, the button snaps to the bottom and gets invisible because of the keyboard laying above it.

FCDC9162-CC99-416F-BE1A-3F1A6B6C4B01
DF24DB59-8BF9-4B13-BD11-E5E8E23B9A0B

hansemannn:
## Description

This change fixes an issue that occurs when the user attempts to close the TAN view by using the "swipe-left" gesture . The fix is to use the current TAN validation state instead of a hardcoded `true` boolean value.

Closes #591.


thomasaugsten:
This maybe also disable the button on the intro screen when you swipe back twice

tens0rfl0w:
Does this also fix the disappearing button when the swipe-down gesture is cancelled? If not, please don't close the mentioned issue after merging this PR. 

Thanks!

hansemannn:
@thomasaugsten This fix is only for the TAN view, but I can look into the intro bug as well. Can you reference the issue link? It should be handled in an own pull request though :)

@tens0rfl0w: Yep, I just added one more commit that fixes the disappearing as well. It was caused by a lifecycle-delegate not being update correctly. Works fine noow!

thomasaugsten:
Looks very good thanks a lot for your help here. But we will need time to test this intensively.

hansemannn:
Thanks for the update - happy to help!

thomasaugsten:
The only issue I found when swiping back you have still the animation of the activated button

hansemannn:
@thomasaugsten I tried to keep the fix as isolated as possible to not mix up bug fixes in a single branch. Should I add that fix here as well or via an own PR?

thomasaugsten:
please add it here. Because of circleCI we have to make a own PR with your changes for this issue.

tens0rfl0w:
There are still two bugs open:

- First, the text a the top says 'Selection' instead of 'TAN Entry' after cancelling the back swipe gesture.


- And if you go to the TAN entry view from the 'Request TAN' view and cancel the back swipe there, the button text says 'Call' instead of 'Next'. (Selection view->Hotline view->TAN entry view)

hansemannn:
@tens0rfl0w Thanks! I will add those as well (apprx. early tomorrow)

hansemannn:
@tens0rfl0w The "Call" button state in the TAN input is fixed. For the navigation-item title, it seems like the TAN input view controller (`ExposureSubmissionTanInputViewController`) is not pushed intro the view controller child hierarchy correctly (notable if you compare the view controller transitions from the "Select" VC -> "Call" VC and the "Select" VC -> "TAN" VC.

 Once that's fixed, we can likely remove some other hacks that are currently in the code base as well.

tens0rfl0w:
@hansemannn Great! Thanks for your work on this!🚀

EisvogelSAP:
Hey @tens0rfl0w, @hansemannn, thanks for you contribution. Unfortunately the whole navigation in the submission flow is a little flawed. Please have a look at this PR #568. Further fixes should be applied to the new code base, please.

hansemannn:
@EisvogelSAP Once that PR is merged, I can rebase here and we should be good to go.

tens0rfl0w:
@hansemannn Sadly I would say no after checking out the new PR.😐
Seems like it fixed all glitches mentioned here. But there is still some room for improvement in the flow and smoothness of the view, so maybe there you can bring in your knowledge if you want! 😊

hansemannn:
Sounds reasonable! Closing this one.

PR #601, opened 2020-06-14
[Bug] Fix home navigation bar preferred alpha when scrolling

eisvogel:
This PR fixes the invalid alpha channel of the navigation bar on the home screen when scrolling. The nav bar should not be visible when scrolled to the top. When scrolling down it should slowly appear and each full opacity before the first element reaches the bottom edge of the navigation bar.


inf2381 (approved):

i336616 (approved):
Tested on device, the nav bar appears/disappears as described now.

PR #605, opened 2020-06-14
More HTTPClient Tests + (Small) Signature Verification Refactor

SebastianWild:
## Description

This PR adds a good amount of unit tests to `HTTPClient`, adds some convenience features to help unit test network requests, and makes a small refactoring to the way `HTTPClient` performs signature verification. Signature verification changes were necessary in order to properly unit test that `HTTPClient` actually performs the verification.

We go from 49.5% unit test coverage in `HTTPClient` to 92.5% 👌

However the PR got a lot larger than I expected. 90% changes in unit tests but still 1000+ changes...

## Notable Changes

- We now test POST body content for outgoing requests
- Since there were now so many unit tests for `HTTPClient`, I split them up into multiple `XCTestCase` subclasses.
- Signature verification of `SAPDownloadedPackage` no longer is inside an extension, but broken out into it's own type - `SAPDownloadedPackage.Verifier`. This makes it easier to inject the verification process into `HTTPClient` for use in unit testing. By default, a proper verifier is injected.
- Introduced type in test target `MockNetworkStack` to reduce code repetition and make it easier to bootstrap a `MockUrlSession` and mock `Data` response payloads, status codes, etc.
- Adjusted `MockUrlSession` so that outgoing `URLRequests` can be observed - for example to test the content of POST requests.


src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -23,18 +23,23 @@ final class HTTPClient: Client {
 	// MARK: Creating
 	init(
 		configuration: Configuration,
+		packageVerifier: @escaping SAPDownloadedPackage.Verification = SAPDownloadedPackage.Verifier().verify,
```
SebastianWild, line 26: This is perhaps the most impactful change. 
In order to make unit testing possible for the signature verification, or to skip verification if a unit test wants to, we need a method to extract the verification process from the `HTTPClient`.

I thought passing in a closure for the verification process is a good way of doing this. It's defaulted for the production version but can be replaced as required for a unit test.

ChristianKienle: very nice.

ChristianKienle (approved):
perfect.

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+SubmitTests.swift
```diff
@@ -0,0 +1,212 @@
...
+	func testSubmit_Success() {
+		// Arrange
+		let mockResponse = HTTPURLResponse(url: mockUrl, statusCode: 200, httpVersion: nil, headerFields: nil)
+		let mockURLSession = MockUrlSession(
+			// cannot be nil since this is not a a completion handler can be in (response + nil body)
+			data: Data(),
+			nextResponse: mockResponse,
+			error: nil
+		)
+		let client = HTTPClient(configuration: .fake, session: mockURLSession)
```
haosap, line 50: why not use HTTPClient.makeWith?

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+SubmitTests.swift
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		waitForExpectations(timeout: expectationsTimeout)
+	}
+
+	func testSubmit_Error() {
+		// Arrange
+		let mockURLSession = MockUrlSession(data: nil, nextResponse: nil, error: TestError.error)
+
+		let client = HTTPClient(configuration: .fake, session: mockURLSession)
```
haosap, line 66: why not use HTTPClient.makeWith?

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+SubmitTests.swift
```diff
@@ -0,0 +1,212 @@
...
+
+		// Assert
+		XCTAssertNotNil(error)
+	}
+
+	func testSubmit_SpecificError() {
+		// Arrange
+		let mockURLSession = MockUrlSession(data: nil, nextResponse: nil, error: TestError.error)
+
+		let client = HTTPClient(configuration: .fake, session: mockURLSession)
```
haosap, line 87: same question as before.

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+SubmitTests.swift
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		waitForExpectations(timeout: expectationsTimeout)
+	}
+
+	func testSubmit_ResponseNil() {
+		// Arrange
+		let mockURLSession = MockUrlSession(data: nil, nextResponse: nil, error: nil)
+
+		let client = HTTPClient(configuration: .fake, session: mockURLSession)
```
haosap, line 114: same question as before.

haosap (changes requested):

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+SubmitTests.swift
SebastianWild: Yeah true, these unit tests were copy and pasted from the old unit test class and I did not modify them.

I will do that 👍

PR #607, opened 2020-06-15
[HOTFIX] Call requestRisk on sceneWillEnterForeground

i336616:
In sceneWillEnterForeground(_ scene: UIScene), call riskProvider.requestRisk(userInitiated: false)

This will ensure that every time the user goes back into the app, the app will attempt to request the latest risk.

Internally, the RiskProvider ensures that the logic behind throttling the actual API requests is applied.


ChristianKienle (approved):

haosap (approved):

ChristopherSchmitz:
tested and approved from my side

ChristopherSchmitz (approved):

PR #609, opened 2020-06-15
Fix/improve submission handling

thomasaugsten:
Reorder submission flow


src/xcode/ENA/ENA/Source/Services/ExposureSubmissionService.swift
```diff
@@ -219,17 +219,24 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 					keys[i].transmissionRiskLevel = UInt8(1)
 				}
 			}
+			self.getTANForExposureSubmit(hasConsent: true, completion: { result in
+				switch result {
+				case let .failure(error):
+					completionHandler(error)
+				case let .success(tan):
+					self.client.submit(keys: keys, tan: tan) { error in
```
johannesrohwer, line 227: I would suggest to move this to a separate (private) method.

johannesrohwer (commented):
Good refactoring that moves the exposure submission flow to the appropriate (service) layer which makes it less error prone (less mixing with view conrtoller code) and easier to test.
I would recommend to make getTANForExposureSubmit() private (if thats possible).

ChristianKienle (approved):

SebastianWild (approved):

PR #634, opened 2020-06-16
show indicator progress on the risk cell

aleksandr-tikhonov:
Hi,

This PR adds missed progress indicator to risk cell.

Screenshot 2020-06-16 at 10 35 08

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
@@ -70,7 +70,8 @@ final class HomeLowRiskCellConfigurator: HomeRiskLevelCellConfigurator {
 			let numberDaysString = String(numberDays)
 			let totalDaysString = String(totalDays)
 			let saveDays = String(format: AppStrings.Home.riskCardLowSaveDaysItemTitle, numberDaysString, totalDaysString)
-			let item2 = HomeRiskImageItemViewConfigurator(title: saveDays, titleColor: titleColor, iconImageName: "Icons_TracingCircleFull - Dark", iconTintColor: titleColor, color: color, separatorColor: separatorColor)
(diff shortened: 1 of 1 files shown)


inf2381 (approved):

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
```diff
@@ -70,7 +70,8 @@ final class HomeLowRiskCellConfigurator: HomeRiskLevelCellConfigurator {
 			let numberDaysString = String(numberDays)
 			let totalDaysString = String(totalDays)
 			let saveDays = String(format: AppStrings.Home.riskCardLowSaveDaysItemTitle, numberDaysString, totalDaysString)
-			let item2 = HomeRiskImageItemViewConfigurator(title: saveDays, titleColor: titleColor, iconImageName: "Icons_TracingCircleFull - Dark", iconTintColor: titleColor, color: color, separatorColor: separatorColor)
+			let progressImage: String = numberDays >= totalDays ? "Icons_TracingCircleFull - Dark" : "Icons_TracingCircle-Dark_Step \(numberDays)"
+			let item2 = HomeRiskImageItemViewConfigurator(title: saveDays, titleColor: titleColor, iconImageName: progressImage, iconTintColor: titleColor, color: color, separatorColor: separatorColor)
```
ChristianKienle, line 74: We could re-use the circular progress indicator @haosap wrote - right? :)

ChristianKienle (approved):

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
EisvogelSAP: I wouldn't suggest that, as it is not aligned with the iconography. Also it is more difficult to to a custom view at this point (same goes for risk detail) than sticking with icons.

PR #639, opened 2020-06-16
fix(localization): ExposureSubmissionError_NoRegistrationToken

svewag:
use "Registrierungs-Token" instead of "Registrations-Token"


ChristianKienle (approved):

ChristianKienle (changes requested):
Revoking my previous review... should this go into master?

haosap:
Hi @steins-code ,
Any text changes should be aligned with translator?

haosap:
> Revoking my previous review... should this go into master?

I think any text changes should be aligned with translator: https://github.com/corona-warn-app/cwa-app-ios/issues/440

inf2381:
As simple as this PR is, we have to close it unfortunately and mention it in #440

PR #646, opened 2020-06-16
Adds a mostly empty App Framework target in order to accommodate a smooth transition to a dedicated framework 

ChristianKienle:
New: **CWAppKit**

Screen Shot 2020-06-16 at 14 53 55

At this moment the framework only contains our `UIColor` extension (ENAColor) + the tests + the assets for the colors.

Rationale: Move most of the app into it's own framework target. The app-specific things remain in the app-target. By having at least one framework we can separate things a bit more clearly. `internal` is the default access level which means that something is available in the whole target. Now we can use `internal` to actually make something internal. The unit tests of `CWAppKit` are now not embedded in an app. This means:

- much faster
- runs only on Simulator

In case we (also) need embedded unit tests we could create another test target.

@inf2381 and myself will make sure this PR does not break our build pipeline before merging it to development.


tens0rfl0w:
Hey @ChristianKienle,

How is the status on that feature? :)
Just wanted to check back, because I currently work on an app extension for the newly introduced WidgetKit in iOS 14 (just for fun right now) and this would definitely need a framework to connect to the then shared database (in an app group) for retrieving the exposure status😊

Thanks! :)

PR #651, opened 2020-06-16
settings model tests

ChristianKienle:


src/xcode/ENA/ENA/Source/Models/Settings/SettingsViewModel.swift
```diff
@@ -17,48 +17,53 @@
...
+		tracing: Main = .tracing,
+		notifications: Main = .notifications,
+		reset: String = AppStrings.Settings.resetLabel
+	) {
 		self.tracing = tracing
 		self.notifications = notifications
 		self.reset = reset
 	}
-
-	static let model = SettingsViewModel(
```
ChristianKienle, line 31: This was a hidden singleton in our app. It was discovered by writing a couple of unit tests for this class. I was able to simply replace `SettingsViewModel.model` by `SettingsViewModel()` because `SettingsViewModel.model` was only used once in our app. So we did not really use it as a singleton.

SebastianWild (approved):

PR #654, opened 2020-06-16
minor improvements?

ChristianKienle:


src/xcode/ENA/ENA.xcodeproj/project.pbxproj
```diff
@@ -2947,9 +2948,9 @@
 					"-DSQLCIPHER_CRYPTO_CC",
 					"-DNDEBUG",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = de.rki.coronawarnapp;
+				PRODUCT_BUNDLE_IDENTIFIER = "de.rki.coronawarnapp-dev";
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = $IPHONE_APP_DIST_PROF_SPECIFIER;
+				PROVISIONING_PROFILE_SPECIFIER = "Corona-Warn-App-Dev";
```
ChristianKienle, line 2953: Had to do this - there is actually another PR open that fixed this already.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionIntroViewController.swift
```diff
@@ -98,10 +102,6 @@ private extension DynamicTableViewModel {
 }
 
 private extension ExposureSubmissionIntroViewController {
-	enum Segue: String, SegueIdentifiers {
```
ChristianKienle, line 101: No longer needed.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionNavigationController.swift
```diff
@@ -82,19 +84,20 @@ class ExposureSubmissionNavigationController: UINavigationController, UINavigati
...
+	private(set) var exposureSubmissionService: ExposureSubmissionService?
+	private weak var submissionDelegate: ExposureSubmissionNavigationControllerDelegate?
 
 	// MARK: - Initializers.
 
 	init?(
 		coder: NSCoder,
 		exposureSubmissionService: ExposureSubmissionService,
-		homeViewController: HomeViewController? = nil,
+		submissionDelegate: ExposureSubmissionNavigationControllerDelegate?,
```
ChristianKienle, line 95: Removed the hard dependency to `HomeViewController`. `submissionDelegate` could now even come non-optional in `init` but this would break the tests and I did not want to make this PR bigger than needed. But this enabled this controller to be tested without a `HomeViewController`. This was possible already (by passing nil) but made it impossible to test certain things.

thomasaugsten (approved):

PR #661, opened 2020-06-16
Chore/quick unit tests

SebastianWild:
## Description

Some super simple & quick unit tests for `Risk.swift` and `String+Today.swift`

### DIFF
--- a/src/xcode/ENA/ENA/Source/Extensions/String+Today.swift
+++ b/src/xcode/ENA/ENA/Source/Extensions/String+Today.swift
@@ -17,16 +17,14 @@
 
 import Foundation
 
-private var todayFormatter = DateFormatter.packagesDateFormatter
-
 extension String {
 	static func formattedToday() -> String {
-		todayFormatter.string(from: Date())
+		DateFormatter.packagesDateFormatter.string(from: Date())
 	}
 }
 
 extension DateFormatter {
-	static var packagesDateFormatter: DateFormatter {
+	static var packagesDateFormatter: DateFormatter = {
 		let formatter = DateFormatter()
 		formatter.dateFormat = "yyyy-MM-dd"
 		formatter.timeZone = TimeZone(abbreviation: "UTC")
@@ -34,5 +32,5 @@ extension DateFormatter {
 		formatter.calendar = Calendar(identifier: .gregorian)
 
 		return formatter
-	}
+	}()
 }
--- a/src/xcode/ENA/ENA/Source/Extensions/String+Today.swift
+++ b/src/xcode/ENA/ENA/Source/Extensions/String+Today.swift
@@ -19,10 +19,18 @@ import Foundation
 
 extension String {
 	static func formattedToday() -> String {
+		DateFormatter.packagesDateFormatter.string(from: Date())
+	}
+}
+
+extension DateFormatter {
+	static var packagesDateFormatter: DateFormatter {
 		let formatter = DateFormatter()
 		formatter.dateFormat = "yyyy-MM-dd"
 		formatter.timeZone = TimeZone(abbreviation: "UTC")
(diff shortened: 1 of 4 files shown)


src/xcode/ENA/ENA/Source/Extensions/__tests__/String+TodayTests.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+
+@testable import ENA
+import Foundation
+import XCTest
+
+final class StringTodayTests: XCTestCase {
+	func testTodayFormat() {
+		let formatted = String.formattedToday()
+		XCTAssertNotNil(formatted.range(of: #"\d{4}-\d{2}-\d{2}"#, options: .regularExpression), "Expected format to be in yyyy-MM-dd")
```
ChristianKienle, line 27: We had a terrible bug in the beginning. The format `formattedToday` used internally was:

```
"yyyy-mm-dd"
```

This gave is dates like:

```
2020-01-23
2020-02-23
2020-13-23
```

but also:

```
2020-56-23
// and
2020-59-23
```
Your regex would not catch that bug. Right? :) 

The correct format would have been:

```
"yyyy-MM-dd"
```

(uppercase m)

We even went so far at some point to write our custom date formatter... Sadly `formattedToday` is not really testable because it uses `Date()`... Not sure if it is worth it changing that now though. Approved.

ChristianKienle (approved):

src/xcode/ENA/ENA/Source/Extensions/__tests__/String+TodayTests.swift
SebastianWild: 🤔 It will not, no.

I considered injecting the `Date` into that function to make more sophisticated unit tests but was not sure if it is worth it. With your comment it may be more worth it than I originally thought.

src/xcode/ENA/ENA/Source/Extensions/String+Today.swift
```diff
@@ -17,12 +17,22 @@
 
 import Foundation
 
+private var todayFormatter = DateFormatter.packagesDateFormatter
```
lennartstolz, line 20: One minor brain fart regarding 6c87f8ea9dac79fbf337d1f8ffc83560f3ec8a57:

```swift
extension DateFormatter {
    static var packagesDateFormatter: DateFormatter = {
        print("Init formatter")
        let formatter = DateFormatter()
        formatter.dateFormat = "yyyy-MM-dd"
        formatter.timeZone = TimeZone(abbreviation: "UTC")
        formatter.locale = Locale(identifier: "en_US_POSIX")
        formatter.calendar = Calendar(identifier: .gregorian)

        return formatter
    }()
}
```

This also makes sure the `packagesDateFormatter` will only be instantiated once and you don't need to have it stored in this `private var`.

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/RiskTests.swift
```diff
@@ -0,0 +1,59 @@
...
+		XCTAssertEqual(details.numberOfDaysWithActiveTracing, 1)
+	}
+
+	func testGetNumberOfDaysActiveTracing_OneDayExact() {
+		let details = mockDetails(hours: 25)
+		XCTAssertEqual(details.numberOfDaysWithActiveTracing, 1)
+	}
+
+	func testGetNumberOfDaysActiveTracing_FourteenDaysExact() {
+		let details = mockDetails(hours: 336)
```
haosap, line 46: it might be better as 24 * 14, 336 is magic. :-)

haosap (changes requested):
Good job, again!

inf2381 (approved):

PR #669, opened 2020-06-16
Fix: Privacy protection window never shown

maybeniko:
### Description
Now when you open the app switcher from the app directly the protection view does not get triggered due to focus. It would be more nice if we call it when the scene will resign active instead of only when entering the background state. For this case we need to change the method to `sceneWillResignActive` and the privacy protection window gets called immediately when entering the app switcher mode.

### DIFF
--- a/src/xcode/ENA/ENA/Source/SceneDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/SceneDelegate.swift
@@ -89,7 +89,7 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 		appUpdateChecker.checkAppVersionDialog(for: window?.rootViewController)
 	}
 
-	func sceneDidEnterBackground(_ scene: UIScene) {
+	func sceneWillResignActive(_ scene: UIScene) {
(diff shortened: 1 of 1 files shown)


SebastianWild (changes requested):
Thank you for the PR!
We actually had it implemented this way, but removed that approach in #427 - the reasoning being that it did not play nice with action sheets presented by the app, and also appeared when ex. swiping down for the control center. See below:

IMG_5CF7262836ED-1

The issue can be seen in the info screen (`ExposureSubmissionHotlineViewController`), when tapping on the listed phone number or the button.

I'm actually not sure how to fix this nicely, so that we could show the privacy screen immediately on `sceneWillResignActive(_ :)` while still preserving the looks of the action sheets 🤔

I would love to know how to fix this for all cases 😀

EisvogelSAP:
This is by design. The privacy protection screen is only supposed to show up to protect your privacy, not to obscure the app content when opening system dialogs or going to the task switcher. Hence, it only shows up, when the app goes into background.

Please also refer to the official Apple documentation: https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background

PR #673, opened 2020-06-17
[Bug] Fix ActionTableViewCell accessibility issue (closes #670)

### LINKED ISSUE #670: [Bug] switches in Settings do not work with VoiceOver
When VoiceOver is activated it is not possible to change the settings for _exposure logging_ and _notifications_.

1.) Turn on VoiceOver
2.) Open the App
3.) Goto: Settings
4.) Click on: *Notifications* or *Exposure Logging*
5.) Click on the On/Off switch 
6.) See that the value does not change

Tested on: iPhone X, iOS 13.5.1
App Version: 1.0.2

SebastianWild:
## Description

#670 Refers to a VoiceOver accessibility issue, where the double tap action is not handled.
`ActionTableViewCell` does implement a custom action, however. This is accessible via a swipe up/down.

Does it make sense to have a custom action as well as to handle the double tap action? I've gone ahead and added this but would love feedback 😀


src/xcode/ENA/ENA/Source/Views/ENSetting/ActionTableViewCell.swift
```diff
@@ -117,4 +117,8 @@ class ActionTableViewCell: UITableViewCell, ActionCell {
 		}
 	}
 
+	override func accessibilityActivate() -> Bool {
+		toggle(self)
```
haosap, line 121: why do you call the toggle?

SebastianWild: I called that since that is what the accessibility custom action already calls.
But looking at it now I think we can refactor this a little bit. Let me see.

haosap (approved):

i336616 (approved):
👍

PR #674, opened 2020-06-17
[Bugfix] Only request user notifications where necessary

### LINKED ISSUE #242: [Bug] Notification settings broken if authorization skipped during onboarding
## Describe the bug

When skipping notification authorization during onboard, the notification settings are broken
* When accessing the notification settings, an authorization popup occurs unexpectedly
* If this popup occurence is fixed, then the settingsOff becomes pointless as a user won't be able to activate the notifications in the iOS settings app as they new will be registered

## Expected behaviour

* Skipping the notification authorization will let the user enable the notifications later through the settings menu
* No unexpected popups should occour

General:
1. Perform Onboarding, but skip notification authorization
2. Open settings 
3. Tap on notifications -> Popup occurs

iOS Settings issue:
Remove the authorization requests in the settings controllers
1. Perform Onboarding, but skip notification authorization
2. Open settings 
3. Tap on notifications
4. Tap to open iOS Settings -> No notifications settings availabile

## Technical details

- iPhone 8 Simulator w/ iOS 13.5

## Possible Fix

To my best knowledge, there is no way to register notifications in the iOS settings without an pop-up appearing.

I've already created a pull request #201 that would fix this issue by tracking if a user skipped notification authorization during onboard and provide a nicer way to enable them later.
#201 was intended to get rid of the unexpected popup, but I've just realized that the underlying issue is more severe than I've initially thought.

inf2381:
Fixes #242 

## Checklist

* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
@@ -61,6 +61,9 @@ class NotificationSettingsViewController: UIViewController {
 			object: UIApplication.shared
 		)
 
+		// Setup view to prevent unrendered content behind the UserNotification alert
+		setupView()
+
 		notificationSettings()
 	}
 
@@ -87,12 +90,10 @@ class NotificationSettingsViewController: UIViewController {
 	private func notificationSettings() {
 		let center = UNUserNotificationCenter.current()
 
-		center.getNotificationSettings { [weak self] settings in
+		center.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
 			guard let self = self else { return }
 
-			let authorized = (settings.authorizationStatus == .authorized) || (settings.authorizationStatus == .provisional)
-
-			self.viewModel = authorized ? NotificationSettingsViewModel.notificationsOn(self.store) : NotificationSettingsViewModel.notificationsOff()
+			self.viewModel = granted ? NotificationSettingsViewModel.notificationsOn(self.store) : NotificationSettingsViewModel.notificationsOff()
 
 			DispatchQueue.main.async {
 				self.setupView()
--- a/src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
@@ -93,6 +93,12 @@ class NotificationSettingsViewController: UIViewController {
 		center.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
 			guard let self = self else { return }
 
+			if let error = error {
+				log(message: "Error while requesting notifications permissions: \(error.localizedDescription)")
+				self.viewModel = NotificationSettingsViewModel.notificationsOff()
+				return
+			}
+
 			self.viewModel = granted ? NotificationSettingsViewModel.notificationsOn(self.store) : NotificationSettingsViewModel.notificationsOff()
 
 			DispatchQueue.main.async {
--- a/src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
@@ -134,21 +134,14 @@ final class SettingsViewController: UITableViewController {
 	private func notificationSettings() {
 		let currentCenter = UNUserNotificationCenter.current()
 
-		currentCenter.getNotificationSettings { [weak self] (settings) in
+		currentCenter.getNotificationSettings { [weak self] settings in
 			guard let self = self else { return }
 
-			switch settings.authorizationStatus {
-			case .notDetermined, .denied:
-				self.settingsViewModel.notifications.setState(state: false)
-			case .authorized, .provisional:
-				if self.store.allowRiskChangesNotification || self.store.allowTestsStatusNotification {
-					self.settingsViewModel.notifications.setState(state: true)
-				} else {
-					self.settingsViewModel.notifications.setState(state: false)
-				}
-			@unknown default:
-				logError(message: "Unknown state of UserNotificationSettings")
-				self.settingsViewModel.notifications.setState(state: false)
+			if (settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional)
+			&& (self.store.allowRiskChangesNotification || self.store.allowTestsStatusNotification) {
+				  self.settingsViewModel.notifications.setState(state: true)
+			} else {
+				  self.settingsViewModel.notifications.setState(state: false)
 			}
 
 			DispatchQueue.main.async {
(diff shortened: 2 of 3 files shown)


AdnanZiSAP:
I'm on it...

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
```diff
@@ -87,12 +90,10 @@ class NotificationSettingsViewController: UIViewController {
 	private func notificationSettings() {
 		let center = UNUserNotificationCenter.current()
 
-		center.getNotificationSettings { [weak self] settings in
+		center.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
```
AdnanZiSAP, line 93: If we don't want to handle the error here, please remove unused parameter (_). It produces a warning. But, I would rather use it and log the error here, the same way as it was logged before in ```SettingsViewController```.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
```diff
@@ -134,18 +134,20 @@ final class SettingsViewController: UITableViewController {
 	private func notificationSettings() {
 		let currentCenter = UNUserNotificationCenter.current()
 
-		currentCenter.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
+		currentCenter.getNotificationSettings { [weak self] (settings) in
```
AdnanZiSAP, line 137: No need for braces around the parameter.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
```diff
@@ -134,18 +134,20 @@ final class SettingsViewController: UITableViewController {
...
 			guard let self = self else { return }
 
-			if let error = error {
-				log(message: "Error while requesting notifications permissions: \(error.localizedDescription)")
+			switch settings.authorizationStatus {
```
AdnanZiSAP, line 140: Switch is not needed here IMO. There are only two states when notifications are enabled, ```.authorized``` and ```.provisional```.

```swift
if (settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional) && (self.store.allowRiskChangesNotification || self.store.allowTestsStatusNotification) {
      self.settingsViewModel.notifications.setState(state: true)
} else {
      self.settingsViewModel.notifications.setState(state: false)
}
```

Something like that. That's also encouraged by Apple in the [official doc](https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications).

AdnanZiSAP (changes requested):

AdnanZiSAP:
Works good. Only a few comments in a wake of us not being under the timeline anymore and having to take a bit more care about the code now. :)

sebastianpicklum:
Hi @inf2381 , thanks for the heads-up.

I've compile the branch and there is a popup still coming up when entering settings.
See https://youtu.be/Un2RYn9W1_E

inf2381:
> Hi @inf2381 , thanks for the heads-up.
> 
> 
> 
> I've compile the branch and there is a popup still coming up when entering settings.
> 
> See https://youtu.be/Un2RYn9W1_E

Yes. This is intended. Because we need to show the alert at some point. If we don’t show it and then forward the user to the system settings, there won’t be an entry for notifications.

sebastianpicklum (commented):
> > Hi @inf2381 , thanks for the heads-up.
> > I've compile the branch and there is a popup still coming up when entering settings.
> > See https://youtu.be/Un2RYn9W1_E
> 
> Yes. This is intended. Because we need to show the alert at some point. If we don’t show it and then forward the user to the system settings, there won’t be an entry for notifications.

Absolutly. But this might confuse some users since they are asked for authorization without any context. 
I would suggest having the pop-up appear when clicking on the the toggle switches in the notification settings. That wouldn't confuse the user as it should be clear that enabling notifications would result in an authorization pop-up appearing.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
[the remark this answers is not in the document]
inf2381: Our current design looks like this if the notifications are not enabled / authorized, yet.
IMG_7878

I totally agree with your opinion but since the RKI needs to sign-off changes on the UI, I would recommend closing the current issue (the alert now shows up on the correct screen) and directly create a feature request and discuss that with the designers and the RKI.

inf2381: Fixed

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettingsViewController.swift
inf2381: Fixed

inf2381:
> Absolutly. But this might confuse some users since they are asked for authorization without any context.
> I would suggest having the pop-up appear when clicking on the the toggle switches in the notification settings. That wouldn't confuse the user as it should be clear that enabling notifications would result in an authorization pop-up appearing.

Our current design looks like you can see in the subsequent screenshot if the notifications are not enabled / authorized, yet.
I totally agree with your opinion but since the RKI needs to sign-off changes on the UI, I would recommend closing the current issue (the alert now shows up on the correct screen) and directly create a feature request and discuss that with the designers and the RKI. 
IMG_7878

AdnanZiSAP (approved):

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsViewController.swift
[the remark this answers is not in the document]
AdnanZiSAP: The popup will not appear here. This is fine.

AdnanZiSAP:
> > Absolutly. But this might confuse some users since they are asked for authorization without any context.
> > I would suggest having the pop-up appear when clicking on the the toggle switches in the notification settings. That wouldn't confuse the user as it should be clear that enabling notifications would result in an authorization pop-up appearing.
> 
> I attached a screenshot of how the current screen looks like if the notifications are not enabled / authorized, yet.
> I totally agree with your opinion but since the RKI needs to sign-off changes on the UI, I would recommend closing the current issue (the alert now shows up on the correct screen) and directly create a feature request and discuss that with the designers and the RKI.
> IMG_7878

Toggle switches are not shown before notifications are enabled. And this suggestion is not good anyway, because if user declined the permission already (while onboarding for example) the alert will not be shown even if authorization is requested. User must be aware that he needs to go into the Settings and enable notifications. 

It's best to close this PR and the issue and check with the PO how and when the alert should be shown. In my opinion, the way it looks now, Settings screen was a better place for it. 😀 

We have a process in place, with stakeholders, PO, designers, etc., and we should not make decisions without using it.

haosap (approved):

PR #676, opened 2020-06-17
[fix] change timeout interval

steins-code:
increase timeout interval to avoid wrong internet off states

### DIFF
--- a/src/xcode/ENA/ENA/Source/SceneDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/SceneDelegate.swift
@@ -18,6 +18,7 @@
 import BackgroundTasks
 import ExposureNotification
 import UIKit
+import Connectivity
 
 final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDependencies {
 	// MARK: Properties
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -186,6 +187,8 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 	}
 
 	private func presentHomeVC() {
+		Connectivity.urlSessionConfiguration.timeoutIntervalForRequest = 15.0
+		Connectivity.urlSessionConfiguration.timeoutIntervalForResource = 15.0
```
haosap, line 191: what are the old values?

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -186,6 +187,8 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
...
 		enStateHandler = ENStateHandler(
 			initialExposureManagerState: exposureManager.preconditions(),
 			reachabilityService: ConnectivityReachabilityService(
```
haosap, line 194: In fact, the EnStateHandler has a strong reference to ConnectivityReachabilityService and meanwhile, ConnectivityReachabilityService also keeps the strong ref to EnStateHandler. please check it.

haosap (changes requested):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
ChristianKienle: 5s

haosap (approved):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
steins-code: 5s

ChristianKienle (approved):

PR #682, opened 2020-06-17
[Technical] Marked the "not implemented" initializers as "unavailable" (closes #681)

### LINKED ISSUE #681: [Technical] Mark "not implemented" initializers as "unavailable"
When building subclasses of `UIKit` classes conforming to `NSCoding` (e.g. `UIViewController`) 
and adding custom initializers for those classes, "forces" you to implement the required initializer: 

```swift
required init?(coder _: NSCoder) {.. }
```

In most scenarios, this is unwanted behavior and you stick with the Xcode autosuggestion which causes a `fatalError` at runtime if the initializer gets called:

```swift
required init?(coder: NSCoder) {
    fatalError("init(coder:) has not been implemented")
}
```

Marking them as "unavailable" has two main benefits:

1) It will cause an error at compile time instead of runtime 
(to catch the misuse of those initializers earlier).

2) It will stop Xcode code completion suggesting those initializers.

## Current Implementation

```swift
required init?(coder: NSCoder) {
    fatalError("init(coder:) has not been implemented")
}
```

This causes a runtime crash similar to the following when this initializer gets called:

runtime-error

## Suggested Enhancement

```swift
@available(*, unavailable)
required init?(coder: NSCoder) {
    fatalError("init(coder:) has not been implemented")
}
```

This will cause the following compile time error:

compile-time-error

## Expected Benefits

1) Avoid the risk of having runtime crashes when those initializers get called accidentally.

2) Stopping Xcode code completion suggesting those initializers, like this:

throwing-initializer-available

vs.

throwing-initializer-not-available

## Known Downsides

None. (last famous words 😉 )

lennartstolz:
This PR addresses the issue described in #681 .

## Description

When providing a custom initializer for subclasses of (mostly) UIKit
classes conforming to 'NSCoding' the required initializer must be
provided by those subclasses.

Most of them will never be used and stay with the auto suggestion
of Xcode causing a `fatalError`:

```swift
required init?(coder: NSCoder) {
    fatalError("init(coder:) has not been implemented")
}
```

As this should never be called, I marked those initializers
with `available(*, unavailable)`. Calling those initializers
will now fail at compile time instead of causing a
runtime crash.

Additionally, it will stop Xcode suggesting those initializers
as code completions.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Developer Menu/DMConfigurationViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/DMConfigurationViewController.swift
@@ -31,6 +31,7 @@ final class DMConfigurationViewController: UITableViewController, RequiresAppDep
 		title = "⚙️ Configuration"
 	}
 
+	@available(*, unavailable)
 	required init?(coder _: NSCoder) {
 		fatalError("init(coder:) has not been implemented")
 	}
@@ -131,6 +132,7 @@ private class DMConfigurationCell: UITableViewCell {
 		super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
 	}
 
+	@available(*, unavailable)
 	required init?(coder _: NSCoder) {
 		fatalError("init(coder:) has not been implemented")
 	}
--- a/src/xcode/ENA/ENA/Source/Developer Menu/DMQRCodeScanViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/DMQRCodeScanViewController.swift
@@ -36,6 +36,7 @@ final class DMQRCodeScanViewController: UIViewController {
 		super.init(nibName: nil, bundle: nil)
 	}
 
+	@available(*, unavailable)
 	required init?(coder _: NSCoder) {
 		fatalError("init(coder:) has not been implemented")
 	}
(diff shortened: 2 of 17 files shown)


lennartstolz:
@inf2381 I may need your help here (And I saw that it also is a "frequent" issue of some contributors). Do you have any idea how to unblock the issue with `circleci`?
It says the `test-an-scan` pipeline is unauthorized.

(Unfortunately, I didn't work with `circleci` for ages)

AFAIK it's happening in this PR as well: #639

inf2381:
image
Could you please login to CircleCI and grant access to your organization?

lennartstolz:
Thanks @inf2381 . 
Yeah I tried everything on my side, even removed all private projects and organizations. 
All of this didn't help. ;)

inf2381:
> Thanks @inf2381 .
> Yeah I tried everything on my side, even removed all private projects and organizations.
> All of this didn't help. ;)

The issue was that you didn't have access to the context on CircleCI where the access token for SonarCloud is stored. I deleted the context from (config.yaml)[https://github.com/corona-warn-app/cwa-app-ios/commit/a22bc9d33e257e8d8b153646d2b2149e169ca714] and stored an access key directly in the environment of CircleCI. It seems to be working now 🥳

inf2381 (approved):

johannesrohwer (approved):

PR #685, opened 2020-06-17
DetectionMode adjustments

tschulze:
## Checklist

* [X] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [X] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
* [X] Set a speaking title. Format: {task_name} (closes #{issue_number}). For example: Use logger (closes # 41)
* [X] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)
* [X] Create Work In Progress [WIP] pull requests only if you need clarification or an explicit review before you can continue your work item.
* [X] Make sure that your PR is not introducing _unnecessary_ reformatting (e.g., introduced by on-save hooks in your IDE)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/DetectionMode.swift
+++ b/src/xcode/ENA/ENA/Source/Models/DetectionMode.swift
@@ -28,7 +28,7 @@ enum DetectionMode {
 }
 
 extension DetectionMode {
-	static func from(backgroundStatus: UIBackgroundRefreshStatus) -> DetectionMode {
+	static func fromBackgroundStatus(_ backgroundStatus: UIBackgroundRefreshStatus = UIApplication.shared.backgroundRefreshStatus) -> DetectionMode {
 		switch backgroundStatus {
 		case .restricted, .denied:
 			return .manual
--- a/src/xcode/ENA/ENA/Source/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate.swift
@@ -193,8 +193,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 extension AppDelegate: ENATaskExecutionDelegate {
 	func executeExposureDetectionRequest(task: BGTask, completion: @escaping ((Bool) -> Void)) {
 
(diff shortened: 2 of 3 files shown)


src/xcode/ENA/ENA/Source/Models/DetectionMode.swift
```diff
@@ -28,8 +28,8 @@ enum DetectionMode {
 }
 
 extension DetectionMode {
-	static func from(backgroundStatus: UIBackgroundRefreshStatus) -> DetectionMode {
-		switch backgroundStatus {
+	static func fromBackgroundRefreshStatus() -> DetectionMode {
```
lennartstolz, line 31: IMHO this change would make the method harder to test, as you would need to swizzle/mock the `backgroundRefreshStatus` of `UIApplication.shared` when you'd like to test if it returns the expected values.

What do you think about moving it into the method signature as a default argument?

```swift
static func from(backgroundStatus: UIBackgroundRefreshStatus = UIApplication.shared.backgroundRefreshStatus) { ... }
```

This way you would get rid of the duplicated code you wanted to remove, but you could even add tests for this method.

tschulze: This was my first idea, but `DetectionMode.from()` looks a little bit strange.

lennartstolz: Fair point, didn't think about that. But what's about?

```swift
static func fromBackgroundStatus(_ backgroundStatus: UIBackgroundRefreshStatus = UIApplication.shared.backgroundRefreshStatus) { ... }
```

This should solve it, no? :)

tschulze: I kept the old method (for mocking) and the new one. Does this look better for you?

ChristianKienle (approved):

PR #687, opened 2020-06-17
[Bug] Submission Flow Navigation

eisvogel:
This PR re-implements the submission navigation flow, thereby fixing several issues, inconsistencies and bugs, including, but not limited to:
* inconsistent button handling
* navigation bar misalignments
* uncomfortable animations
* complex source code
* incompatibility with button accessibility

Following this PR we can start trying to integrate the button accessibility PR #439, which also includes an improved activity state for asynchronous task during test submission.

Please review this PR carefully as it may break the current submission flow, if something went haywire during this refactoring.

Issues mentioned in #591, #594, #622 should have been resolved.

PR #697, opened 2020-06-17
[Technical] Added '@discardableResult' attribute to 'ENATanInput' (closes #696)

### LINKED ISSUE #696: [Technical] Fix the Swift Compiler Warning of 'ExposureSubmissionTanInputViewController'
This awesome project currently has almost no Swift compiler warnings. 👏 
One of the remaining one is caused by overriding the `becomeFirstResponder` method of `ENATanInput`.

The underlying issue is also discussed in the Swift project itself: [SR-2198](https://bugs.swift.org/browse/SR-2198)

#### TL;DR

The `@discardableResult` attribute (or the Objective-C equivalent `warn_unused_result`) is not inherited if you override any method in a subclass. And this is on purpose.

## Current Implementation

When you currently call the `becomeFirstResponder`  and `resignFirstResponder` methods of `ENATanInput` the way you would call "direct implementations" of `UIResponder`:

```swift
foo.becomeFirstResponder() // In most scenarios you don't care about the result.
```

it causes the following Swift compiler warning:

> Result of call to 'becomeFirstResponder()' is unused

## Suggested Enhancement

By adding the `@discardableResult` attribute to the corresponding methods, the warning will be gone.

E.g.

```swift
@discardableResult
override func becomeFirstResponder() -> Bool {
	delegate?.enaTanInputDidBeginEditing?(self)
	return super.becomeFirstResponder()
}
```

## Expected Benefits

The Swift compiler warning is gone. 🎉 

## Discarded Alternatives

You could also get rid of this warning by ignoring the return value like this:

```swift
_ = foo.becomeFirstResponder()
```

I discarded this approach because it would mean that the API needs to be used in a different way than it was initially intended (and moves the "problem" to the point of use).

lennartstolz:
This PR addresses the issue described in #696  .

This removes the Swift compiler warning when calling the `becomeFirstResponder` and `resignFirstResponder` methods of the `ENATanInput` class.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Views/BaseElements/ENATanInput.swift
+++ b/src/xcode/ENA/ENA/Source/Views/BaseElements/ENATanInput.swift
@@ -117,11 +117,13 @@ extension ENATanInput {
 }
 
 extension ENATanInput {
+	@discardableResult
 	override func becomeFirstResponder() -> Bool {
 		delegate?.enaTanInputDidBeginEditing?(self)
 		return super.becomeFirstResponder()
 	}
 
+	@discardableResult
 	override func resignFirstResponder() -> Bool {
 		delegate?.enaTanInputDidEndEditing?(self)
 		return super.resignFirstResponder()


EisvogelSAP:
Hey @lennartstolz, thanks for your contribution. This is already part of #687.

PR #701, opened 2020-06-18
Change "Mundschutz" to "Mund-Nasen-Bedeckung"

CeleritasCognito:
This pull-request changes the translation for "Mundschutz" to "Mund-Nasen-Bedeckung" this is because "Mundschutz" is confusing many people and they wear it only on their mouth instead of the mount and nose. It is also the term used from RKI:
https://www.rki.de/SharedDocs/FAQ/NCOV2019/FAQ_Mund_Nasen_Schutz.html


inf2381:
Thanks for your suggestion. Please have a look at #440 as we cannot influence the texts within the app.

PR #728, opened 2020-06-19
[Chore] Split background tasks on develop branch

i336616:
This PR changes the operation of the background tasks.

Previously, the background task only triggered once every 4 hours minimum. On triggering the task, the fetch-test-results API call is executed, and on completion, the exposure-detection API call is executed, effectively executing the two API asynchronous calls in series.

This change splits the execution of the API calls into two separate tasks. This allows the exposure detection call to maintain the higher priority, and adhere to the minimum timing specified by the ExposureNotification framework.

No changes have been made to the logic to determine whether to bypass the API calls, which are still driven by the RiskProvider class (for the exposure-detection API call) and the ENAExposureSubmissionService class (for the fetch-test-results API call)


ChristianKienle (approved):

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -22,14 +22,17 @@ import UIKit
 enum ENATaskIdentifier: String, CaseIterable {
 	// only one task identifier is allowed have the .exposure-notification suffix
 	case primaryBackgroundTask = "exposure-notification"
+	case secondaryBackgroundTask = "fetch-test-results"
 
 	var backgroundTaskScheduleInterval: TimeInterval? {
 		switch self {
 		case .primaryBackgroundTask: return nil
+		case .secondaryBackgroundTask: return nil
```
johannesrohwer, line 30: would it be more clear to name the tasks `FetchTestResultBackgroundTask` or something like that instead of primary/secondaryBackgroundTask?

johannesrohwer (approved):
Splitting the background task into two separate ones makes the code a lot easier to understand. Thanks!

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -22,14 +22,17 @@ import UIKit
 enum ENATaskIdentifier: String, CaseIterable {
 	// only one task identifier is allowed have the .exposure-notification suffix
 	case primaryBackgroundTask = "exposure-notification"
```
melloskitten, line 24: why the odd naming of primary and secondary? Why not give them an actual descriptive name? like the `fetchTestResultsTask`.

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -22,14 +22,17 @@ import UIKit
...
+	case secondaryBackgroundTask = "fetch-test-results"
 
 	var backgroundTaskScheduleInterval: TimeInterval? {
 		switch self {
 		case .primaryBackgroundTask: return nil
```
melloskitten, line 29: what is the reason for having a switch statement here, if the return value is `nil` regardless?

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -50,7 +53,8 @@ final class ENATaskScheduler {
 	typealias CompletionHandler = (() -> Void)
 
 	private func registerTasks() {
-		registerTask(with: .primaryBackgroundTask, taskHander: executeBackgroundTask(_:))
+		registerTask(with: .primaryBackgroundTask, taskHander: executeExposureDetectionRequest(_:))
```
melloskitten, line 56: `taskHander` ? shouldn't it be `taskHandler`?

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -62,12 +66,18 @@ final class ENATaskScheduler {
 
 	func scheduleTasks() {
 		scheduleTask(for: .primaryBackgroundTask, cancelExisting: true)
+		scheduleTask(for: .secondaryBackgroundTask, cancelExisting: true)
 	}
 
 	func cancelTasks() {
 		BGTaskScheduler.shared.cancelAllTaskRequests()
```
melloskitten, line 73: If you cancel all task requests, doesn't it mean that you would be canceling the `secondaryBackgroundTask` from the `primaryBackgroundTask` and vice versa? Do you really want this?

melloskitten (changes requested):
Just a few minor questions, the rest looks fine to me :)

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
i336616: 👍

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
i336616: Agreed, have updated accordingly!

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
i336616: It's using the standard enum syntax.

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
i336616: Yep, have fixed that now!

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
melloskitten: Why is it not possible to change
```swift

var backgroundTaskScheduleInterval: TimeInterval? {
		switch self {
		case .exposureNotification: return nil
		case .fetchTestResults: return nil
		}
	}
```

to 

```swift

var backgroundTaskScheduleInterval: TimeInterval?
```

if it is nil anyways, regardless of the case.

i336616: Apologies, I thought you were referring to the case lines above. I've updated the fetch-test-results to a 2hr schedule

melloskitten: No problem. Thanks for fixing it! 😄

melloskitten:
Looks awesome! Thanks a lot! 🎈 🥳

melloskitten (approved):

src/xcode/ENA/ENA/Resources/Info_Debug.plist
```diff
@@ -4,8 +4,8 @@
 <dict>
 	<key>BGTaskSchedulerPermittedIdentifiers</key>
 	<array>
-		<string>de.rki.coronawarnapp.exposure-notification</string>
-		<string>de.rki.coronawarnapp.fetch-test-results</string>
+		<string>$(PRODUCT_BUNDLE_IDENTIFIER).exposure-notification</string>
+		<string>$(PRODUCT_BUNDLE_IDENTIFIER).fetch-test-results</string>
```
haosap, line 8: why do you replace it with the environment var?

johannesrohwer: i guess its done to make sure there are no collisions with different app versions, e.g. one that was installed via test flight vs. a debug build? (e.g. the debug target uses de.rki.coronawarnapp-dev, production targets use de.rki.coronawarnapp)

i336616: It allows us to use the app bundle ID as the prefix for the identifier. In the code, we use Bundle.main.bundleIdentifier as the prefix. This means that we can deploy either app targets during dev (eg from 'de.rki.coronawarnapp' to 'de.rki.coronawarnapp-dev'), and ensure that background tasks from one instance will not collide with the other app's background task identifiers. It's also the method used by Apple's sample code for exposure notifications.

PR #745, opened 2020-06-22
EXPOSUREAPP-1531 Sharing function is missing explanation text

AdnanZiSAP:
# What this PR does 

Fixes a bug reported in [#1531](https://jira.itc.sap.com/browse/EXPOSUREAPP-1531), and closes [subtask #1579](https://jira.itc.sap.com/browse/EXPOSUREAPP-1579). 

The issue was that the share text was missing from share messages, specifically reported for Whatsapp. That is now fixed. 

### iMessage

IMG_0F68D46DBFE5-1

### Apple Mail

IMG_990857B8325F-1

### Gmail

IMG_450488A44E1E-1

### Whatsapp

IMG_BD9ADC78AC79-1


src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
```diff
@@ -79,11 +79,15 @@ final class FriendsInviteController: UIViewController {
 
 extension FriendsInviteController: UIActivityItemSource {
 	func activityViewControllerPlaceholderItem(_: UIActivityViewController) -> Any {
-		shareTitle
+		return shareTitle
```
haosap, line 82: Do you prefer the `return`?

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
```diff
@@ -79,11 +79,15 @@ final class FriendsInviteController: UIViewController {
...
 	}
 
-	func activityViewController(_: UIActivityViewController, itemForActivityType _: UIActivity.ActivityType?) -> Any? {
-		shareUrl
+	func activityViewController(_: UIActivityViewController, itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
+		if activityType == .mail {
+			return ""
```
haosap, line 87: if you return a nil, what will happen?

haosap (commented):
Hi @AdnanZiSAP ,
it looks good.
Thanks.

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
```diff
@@ -79,11 +79,15 @@ final class FriendsInviteController: UIViewController {
...
 extension FriendsInviteController: UIActivityItemSource {
 	func activityViewControllerPlaceholderItem(_: UIActivityViewController) -> Any {
-		shareTitle
+		return shareTitle
 	}
 
-	func activityViewController(_: UIActivityViewController, itemForActivityType _: UIActivity.ActivityType?) -> Any? {
-		shareUrl
+	func activityViewController(_: UIActivityViewController, itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
+		if activityType == .mail {
```
johannesrohwer, line 86: what is different for mail here?

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
AdnanZiSAP: I don't, but the whole project has it, so I change these too, to be consistent.

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
AdnanZiSAP: The different is that we have the subject for mail (```activityViewController(_: UIActivityViewController, subjectForActivityType _: UIActivity.ActivityType?)```), and I didn't want to repeat here, since the same text is used in the subject.

Unfortunately some other mail clients, like Gmail, are not treated as ```mail```, so for them this doesn't apply, but it's good since Gmail takes this body text as the subject text too.

Please have a look at screenshots above.

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
AdnanZiSAP: The subject is gone from Apple Mail app. :)

haosap (approved):

johannesrohwer (approved):

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/FriendsInviteController.swift
johannesrohwer: thanks for the explanation! :)

Issue #767, opened 2020-06-24
[UI] Nav bar transparency


tens0rfl0w:
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is not already reported in another issue

## Describe the bug
Just a short one: If you open the FAQ from the home view and close it, the nav bar stays solid and does not go transparent when scrolling back to the top. Sometimes the opposite happens and the navbar stays transparent even when scrolling.

![image](https://user-images.githubusercontent.com/57523343/85552239-17586100-b623-11ea-9dc9-ed7a7ed3feb5.jpeg)
![image](https://user-images.githubusercontent.com/57523343/85552344-33f49900-b623-11ea-8265-ff467bf5dede.png)

---
Internal Tracking ID: EXPOSUREAPP-1944

EisvogelSAP:
Hey @tens0rfl0w, thanks for pointing this out. Will have fix incoming for this tomorrow morning. :-)

EisvogelSAP:
Hey @tens0rfl0w, sorry for the delay here. Unfortunately I was overruled. This needs to be aligned internally before we can continue here.

Again, thanks for raising this issue to us and sorry for the delay.

tens0rfl0w:
Hey @EisvogelSAP, 

I have seen the discussion in the PR. No worries, luckily it breaks no functionality!🙂

aleksandr-tikhonov:
I'd like to provide a comparison of custom and default nav bar implementations

### Custom
on the top
![custom 1](https://user-images.githubusercontent.com/61236601/86588928-f0881c00-bf8c-11ea-9ace-bd41d948ec9e.PNG)
on the middle/bottom
![custom 2](https://user-images.githubusercontent.com/61236601/86588934-f2ea7600-bf8c-11ea-9555-de5f3bbd6cf9.PNG)


### Default
on the top
![default 1](https://user-images.githubusercontent.com/61236601/86588940-f847c080-bf8c-11ea-871f-c56a33886f6d.PNG)
on the middle/bottom
![default 2](https://user-images.githubusercontent.com/61236601/86588947-fbdb4780-bf8c-11ea-97af-df20b9cad4a9.PNG)

I also attached the video with demonstration how it works.
[demo videos.zip](https://github.com/corona-warn-app/cwa-app-ios/files/4878556/demo.videos.zip)

tens0rfl0w:
[...]

PR #776, opened 2020-06-25
[Bug] Navigation bar opacity

eisvogel:
This PR addresses #767. Thanks to @tens0rfl0w. However, also included is some refactoring for the actual `AppNavigationController`.

The changes affect the navigation bar on the home screen as well as during onboarding. Please test both flows while reviewing. The navigation bar should not be visible when scrolled to the top. When scrolling down it should slowly appear and reach full opacity before the first element reaches the bottom edge of the navigation bar when on the home screen. During onboarding images may touch the navigation bar before it has full opacity. The navigation bar for app information and settings should appear solid white.

While testing, please make sure to try and 'break' things by using swipe gestures as well as the back button. Please also test correct behavior when canceling a transition during a swipe gesture. When opening the FAQ the navigation controller is being hidden. This is were #767 comes into play. Please test correct behavior after coming back from a modally presented view controller.


EisvogelSAP:
Closes @767.

inf2381 (approved):

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
```diff
@@ -24,17 +24,19 @@ class AppNavigationController: UINavigationController {
 	private var scrollViewObserver: NSKeyValueObservation?
 	private var defaultScrollEdgeAppearance: UINavigationBarAppearance?
 
-	override func viewDidLoad() {
-		super.viewDidLoad()
+	override func loadView() {
+		super.loadView()
```
haosap, line 28: Why do you do the view setup in the `loadView()` method?
I think Apple suggests doing the view config in the `viewDidLoad()` method.

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
```diff
@@ -24,17 +24,19 @@ class AppNavigationController: UINavigationController {
 	private var scrollViewObserver: NSKeyValueObservation?
 	private var defaultScrollEdgeAppearance: UINavigationBarAppearance?
 
-	override func viewDidLoad() {
-		super.viewDidLoad()
+	override func loadView() {
```
AdnanZiSAP, line 27: Is this the right method to do this kind of stuff in? What difference does it make in the behaviour in practice comparing to ```viewDidLoad```?

haosap (changes requested):

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
EisvogelSAP: @haosap will resolve this on as duplicate of the above ...

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
EisvogelSAP: `loadView` vs. `viewDidLoad` is mostly a matter of personal preference. But a general rule of thumb is to use `loadView` when creating view controllers programmatically where as `viewDidLoad` can be used to customize a view loaded from a storyboard/nib/xib. As this controller does not have an underlying storyboard I switched to `loadView`.

https://stackoverflow.com/questions/3423785/what-is-the-difference-between-loadview-and-viewdidload

AdnanZiSAP: But you're not loading any view, right? :) You just setup an already loaded view.

The [official doc](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621454-loadview) states that: 
"_You can override this method in order to create your views manually. If you choose to do so, assign the root view of your view hierarchy to the view property. ... Your custom implementation of this method should not call super.
... If you want to perform any additional initialization of your views, do so in the viewDidLoad() method._"

So from this it looks like ```ViewDidLoad``` is appropriate here. And furthermore, it says that ```super``` should not be called if this is used.

I'm asking because I'm trying to figure out what is the fix here, how this switch from ```viewDidLoad``` to ```loadView``` helps us fixing the bug. If it helps, this is my assumption, that it's part of the fix.

AdnanZiSAP:
Just finished getting into this code, understanding what was the old code in ```AppNavigationController ``` was doing and then afterwards understanding what the new code is doing. 

I couldn't figure out what's old and what's new, and why some things are changed, probably also because there was some refactoring applied, so instead I tried to figure them out one by one, first the old state and then the new state. And I couldn't figure out either of them. :) When I look what we want to do at the end - we want to customize the navigation bar, and all this complexity, before and afterwards equally, I'm wondering does it have to be this complex.

This bug that was reported, it's pretty hard to reproduce, during my testing it was 1/10. Like I have to put the app in the background for 5 times or so, before it appears once with broken navigation bar, and it's all so random. There might be some other bugs that we don't know about too. And there might be that we create new bugs with this PR too, because it's so complex. And if more bugs appears, fixing them will be very hard.

Perhaps the best idea would be to scrap everything and start as new, first with what we want to achieve and then what's the best and/or simplest way to do it. Or even change requirements if we have to go into all this complexity to make it happen. 

I wouldn't go this way if for nothing else simply because it's hard to predict that by fixing this we won't introduce more hard to trace bugs. 

I would, therefore, like more colleagues to take part in this and give their feedback about this PR, and of course about what I wrote above. :)

EisvogelSAP:
Hey @AdnanZiSAP, thanks for you feedback.

The issue raised by @tens0rfl0w is actually quite easy to reproduce. Every time you open up the FAQ page (which is a modal presentation rather than a navigation push transition) the scroll observer was invalidated in `viewWillDisappear`. When navigating back this observer was not reestablished, resulting in the error.

What I actually did in this PR was to "start as new", trying to reevaluate how to best tackle what is required by design. This PR is the result. I have tried to reengineer the `AppNavigationController` so that it is easier to read and more stable and protected against accidental "fixes" which could break functionality.

Feel free to suggest alternative ways of producing the required result. In my opinion this PR greatly reduces the complexity of the said controller. But it seems I am the only one. :-D

@corona-warn-app/cwa-members As per @AdnanZiSAP request please review this PR.

aleksandr-tikhonov:
I agree with @AdnanZiSAP, this implementation of navigation controller looks quite complex and maybe it's overkill here. It could fix current bug, but it could bring more bugs in the future in specific cases.

EisvogelSAP:
Hey @aleksandr-tikhonov, thanks for you feedback.

I now have two statements about the complexity of the `AppNavigationController` with no clear guidance on where to go from here. If you have any ideas on how to improve this, please come forward. Otherwise I feel we are stalling bug fixes and progress for the sake of trying to re-evaluate previously accepted concepts and changes.

As I stated before: This PR is trying to reduce complexity as far as possible making it easier to understand. Please feel free to suggest alternative approaches. For me this is the best approach of tackling the requirements of the design.

I am also open to discard this PR keeping the old implementation and just adding a minor fix for the bug raised by @tens0rfl0w.

aleksandr-tikhonov:
I'd to consider 2 approaches:

1. very easy, just to use standard `UINavigationController`, it just looks a little bit different from the design, but still nice and bug free. 
2. keep your changes, it's 100% align with design, but we should be ready to support it in the future.

I'd like to continue with **1.** approach.

What other devs think?

EisvogelSAP:
Hey @aleksandr-tikhonov,

OK, then let's go with the second approach, as the design specifications are very clear. Changing design specifications in not in our development scope.

Removing the `AppNavigationController` breaks a very important design specification on the home screen. Furthermore, the onboarding flow will lose functionality as well as some sub screens including settings and app information.

@aleksandr-tikhonov @AdnanZiSAP If you disagree, please align with the design team on how to resolve this.

PR #789, opened 2020-06-26
[Bug] Fix onboarding check for UITests

i336616:
Moved test for isOnboarded to func scene(_ scene: UIScene, willConnectTo _: UISceneSession, options _: UIScene.ConnectionOptions), setting the store.isOnboarded value to the UserDefaults value passed through from the UITest


inf2381 (approved):

inf2381:
Changes already in develop. GitHub seems to have internal issues at the moment

PR #800, opened 2020-06-29
Remove error when submitting 0 keys

melloskitten:
### Description

This PR implements the following functionality: [JIRA Issue #1534](https://jira.itc.sap.com/browse/EXPOSUREAPP-1534)
When the user has no TEK, we silently ignore this and move to the Warn Others screen.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionWarnOthersViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionWarnOthersViewController.swift
@@ -59,17 +59,19 @@ class ExposureSubmissionWarnOthersViewController: DynamicTableViewController, EN
 		navigationFooterItem?.isPrimaryButtonLoading = true
 		navigationFooterItem?.isPrimaryButtonEnabled = false
 		exposureSubmissionService?.submitExposure { error in
-			if let error = error {
+			switch error {
+			case .noKeys:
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionWarnOthersViewController.swift
```diff
@@ -59,17 +59,19 @@ class ExposureSubmissionWarnOthersViewController: DynamicTableViewController, EN
 		navigationFooterItem?.isPrimaryButtonLoading = true
 		navigationFooterItem?.isPrimaryButtonEnabled = false
 		exposureSubmissionService?.submitExposure { error in
-			if let error = error {
+			switch error {
+			case .noKeys:
+				self.performSegue(withIdentifier: Segue.sent, sender: self)
+			case .none:
+				self.performSegue(withIdentifier: Segue.sent, sender: self)
```
EisvogelSAP, line 66: Looks like code duplication. Could also be written in single case statement. Nothing major ... just an idea.

eisvogel (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionWarnOthersViewController.swift
melloskitten: Thank you, I overlooked! Will change 👍

steins-code (approved):

PR #821, opened 2020-07-01
Tracing Status Animation

eisvogel:
This PR adds an animation to the tracing status cell on the home screen when tracing is enabled.

|Light|Dark|
|:--:|:--:|
|ezgif-4-6a69335ee7af|ezgif-4-865f5f0b9d22|


src/xcode/ENA/ENA/Source/Models/Home/HomeActivateCellConfigurator.swift
```diff
@@ -52,10 +51,24 @@ final class HomeActivateCellConfigurator: CollectionViewCellConfigurator {
...
 		cell.accessibilityLabel = cell.titleLabel.text ?? ""
 	}
+
+	private func animateEnabledTracing(cell: ActivateCollectionViewCell) {
+		guard !cell.iconImageView.isAnimating else { return }
+
+		cell.iconImageView.animationDuration = 61.0 / 30
+		cell.iconImageView.animationImages = (0...60).compactMap({ UIImage(named: String(format: "Icons_Risikoermittlung_%02d", $0)) })
+		cell.iconImageView.startAnimating()
+	}
```
haosap, line 71: My suggestion is put this view logic to the `ActivateCollectionViewCell`, the `ActivateCollectionViewCell` exposes the properties to the `Configurator`, for example:
```swift
  class ActivateCollectionViewCell {
       func setTracingEnable(_ isEnable:Bool) {
           // bla bla ...
       }
  }
```
The Configurator only need to call enable/disable.

haosap (changes requested):

src/xcode/ENA/ENA/Source/Models/Home/HomeActivateCellConfigurator.swift
EisvogelSAP: Hey @haosap I am not sure if that would be the right approach here. Unfortunately the current architecture of the home screen cells is a little confusing all together, where logic is either part of the configurator or the view implementation. I don't really see the point of this separation especially in this case, where there is almost no logic and the view implementation currently does not hold any logic at all.

@aleksandr-tikhonov Since you are the architect of the home screen ... what is your take on this?

haosap: Hi @aleksandr-tikhonov ,
What's your idea here? I think the animation is the property/attribute of the view itself. It's a black box of the callee.

aleksandr-tikhonov: Hi, I agree with @haosap that it's better to have view related logic in view. 

I'll try to describe this pattern with cells and configurators shortly.

Configurators just configure the cell, they provide data for the cells and cell shouldn't know about configurators or other models, it just receives view-related data (strings, ints) from outside.

And also configurator shouldn't know about implementation of the cell, so it's better to have 
```
class Cell: UICollectionViewCell {
	@IBOutlet private var titleLabel: UILabel!

	func setupTitle(text: String) {
		titleLabel.text = text
	}
}

class Configurator {
	func config(cell: Cell) {
		cell.setupTitle(text: "Title for the cell")
	}
}
```
than
```
class Configurator {
	func config(cell: Cell) {
		cell.titleLabel.text = "Title for the cell"
	}
}

```
Our current implementation uses second approach and it's not good, it breaks open-closed solid principle, it's place for refactoring. 

About animation, it pure logic for the cell, not for configurator. And I believe, it should be on `Cell` and implementation of animation should be hidden in the cell.

😊

EisvogelSAP: Hey @hao, @aleksandr-tikhonov, I have refactored the configurator according to your proposal.

haosap (approved):

haosap:
Hi @EisvogelSAP ,
Could you please check the SonarCloud? After that, we can merge it.
Thanks.

carsten801:
Hi @haosap, 
sounds great that this is coming in a future release!
Maybe this PR should be inked with the issue https://github.com/corona-warn-app/cwa-wishlist/issues/76.
Have a great day!
Carsten

aleksandr-tikhonov (approved):

PR #825, opened 2020-07-02
Fix blank Data Privacy page after 5 minutes without user interaction

eisvogel:
This PR removes the HTML reload mechanism as it is not required anymore with the current implementation thereby. Closes EXPOSUREAPP-1500


inf2381 (approved):

haosap (commented):
Hi @EisvogelSAP ,
If we remove the code, it won't affect the dark mode?

EisvogelSAP:
Hey @haosap, thanks for pointing this out. Actually this breaks switching from dark to light mode when on the page. I didn't take that into account. I had only tested the view when switch light/dark mode before navigating to that screen. Originally the trait collection callback was required for dynamic type only. I'll try to figure out another approach on how to tackle the issue at hand.

haosap:
Hi @EisvogelSAP ,
Is there any update? It's the part of next release. 
Thanks.

i336616 (approved):
Tested this on device, background the app and switching between dark/light/dark, it works as expected.

PR #834, opened 2020-07-03
Introduce Coordinator

AdnanZiSAP:
# What this PR does

This PR introduces Coordinator pattern into the app to tackle one source of the issue of Home VC having too much responsibilities. 

## Problem

image

## Motivation

image

## Solution

The coordinator object is created to handle main navigation in home screen.

- ```SceneDelegate``` now communicates with ```Coordinator``` instead of ```HomeViewController```.
- ```HomeViewController``` doesn't know about any other view controller anymore.
- ```SceneDelegate``` knows only about root view controller.

This PR doesn't introduce any functional changes, it's basically only restructuring.  When reviewing please make sure that there is no functional impact that this PR introduces. Also, please test that all paths from home screen are still accessible.


aleksandr-tikhonov (approved):

aleksandr-tikhonov:
Added show tests for Coordinator

src/xcode/ENA/ENA/Source/Coordinator.swift
```diff
@@ -0,0 +1,282 @@
...
+				delegate: self,
+				detectionMode: state.detectionMode,
+				exposureManagerState: state.exposureManager,
+				initialEnState: enStateHandler.state,
+				risk: state.risk,
+				exposureSubmissionService: self.exposureSubmissionService
+			)
+		}
+
+		homeController = vc // strong ref needed
```
haosap, line 68: is there reason to do it like this,
```swift
 homeController = vc 
```

haosap (changes requested):
Nice work, I didn't see any big issue from my side. Only some question.

src/xcode/ENA/ENA/Source/Coordinator.swift
aleksandr-tikhonov: I think, it was just copied from app/scene delegate.

aleksandr-tikhonov: so we need to keep reference to `homeViewController` in order to update it later

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -170,88 +160,15 @@ final class HomeViewController: UIViewController {
...
-			completion: nil
-		)
-	}
-
-	func showInviteFriends() {
-		navigationController?.pushViewController(
-			FriendsInviteController.initiate(for: .inviteFriends),
-			animated: true
-		)
+		delegate?.showExposureSubmission(with: result)
```
melloskitten, line 163: What is the reason behind having methods like this? I am not a fan of wrapping one method in another method and introducing 3 lines of code whether you could have just had one. 
I would just use `delegate?.showExposureSubmission(with: result)` where the previous method was called.
Additionally, this way of wrapping methods in another one makes it harder for people to understand the indirections. Again, if it were a method that really capsulates complex logic, having an extra method is perfectly acceptable. But this is a no-brainer line of code.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -170,88 +160,15 @@ final class HomeViewController: UIViewController {
...
 	}
 
 	func showExposureNotificationSetting() {
```
melloskitten, line 166: same as above

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -170,88 +160,15 @@ final class HomeViewController: UIViewController {
...
-		exposureDetectionController = vc as? ExposureDetectionViewController
-		present(vc, animated: true)
-	}
-
-	func showAppInformation() {
-		navigationController?.pushViewController(
-			AppInformationViewController(),
-			animated: true
-		)
+		delegate?.showExposureDetection(state: homeInteractor.state, isRequestRiskRunning: homeInteractor.riskProvider.isLoading)
```
melloskitten, line 171: same as above

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -170,88 +160,15 @@ final class HomeViewController: UIViewController {
...
 	}
 
 	private func showScreenForActionSectionForCell(at indexPath: IndexPath) {
```
melloskitten, line 174: this method for example makes of course sense to refactor it in the way that it has been done here.

src/xcode/ENA/ENA/Source/Coordinator.swift
```diff
@@ -0,0 +1,282 @@
...
+			risk: state.risk
+		)
+		let vc = AppStoryboard.exposureDetection.initiateInitial { coder in
+			ExposureDetectionViewController(
+				coder: coder,
+				state: state,
+				delegate: self
+			)
+		}
+//		delegate?.addToUpdatingSetIfNeeded(vc)
```
melloskitten, line 159: dead code

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -19,10 +19,18 @@ import ExposureNotification
 import UIKit
 
 protocol HomeViewControllerDelegate: AnyObject {
-	func homeViewControllerUserDidRequestReset(_ controller: HomeViewController)
+	func showRiskLegend()
+	func showExposureNotificationSetting(enState: ENStateHandler.State)
+	func showExposureDetection(state: HomeInteractor.State, isRequestRiskRunning: Bool)
+	func setExposureDetectionState(state: HomeInteractor.State, isRequestRiskRunning: Bool)
```
melloskitten, line 25: why does the ExposureDetectionState has to be set in this particular way? I thought that the coordinator would only do the navigation logic. But now it is also doing side-effects by setting certain states.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -19,10 +19,18 @@ import ExposureNotification
...
+	func showExposureSubmission(with result: TestResult?)
+	func showInviteFriends()
+	func showWebPage(from viewController: UIViewController, urlString: String)
+	func showAppInformation()
+	func showSettings(enState: ENStateHandler.State)
+	func addToUpdatingSetIfNeeded(_ anyObject: AnyObject?)
```
melloskitten, line 31: Please add a comment what this method is doing, it's very generic and not understandable.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -43,34 +52,25 @@ final class HomeViewController: UIViewController {
 				exposureManagerState: exposureManagerState,
 				enState: initialEnState,
 				risk: risk
-			))
+			), exposureSubmissionService: exposureSubmissionService)
 		navigationItem.largeTitleDisplayMode = .never
-		addToUpdatingSetIfNeeded(homeInteractor)
+		delegate.addToUpdatingSetIfNeeded(homeInteractor)
```
melloskitten, line 57: Why is this `addToUpdatingSetIfNeeded` needed?

melloskitten (dismissed):
Hi, thanks for your contribution to improve the code quality of the Home View Controller. I have left some comments.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
aleksandr-tikhonov: The idea comes from theory of coordinator pattern. View Controllers shouldn't know about another View Controllers or coordinators, the only thing they just do is say that it completed the flow or done some action. So View Controller doesn't manage the flow itself, for View Controller doesn't matter who handle this, it just delegate it. 

And it's good thing, home vc doesn't know about exposure detection vc or settings vc, home vc just should say `detailButtonTapped`, `settingsActionRequested` or `finishSubmitFlow` by delegates or closures and another object (coordinator) will do logic of creating new vc/scene/module and routing it.

src/xcode/ENA/ENA/Source/Coordinator.swift
AdnanZiSAP: Aleksandr is right. It was just copied from ```SceneDelegate```.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: This was wrapped because the method is called from 4 places in this view controller, and once from ```HomeInteractor```.  There are two reasons: First, so it could be a single point of call to the delegate method, and second, without this ```HomeInterator``` would need to know about the ```Coordinator``` and it shouldn't.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: The same answer as above.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: The same answer as above, minus the ```HomeInteractor``` part.

src/xcode/ENA/ENA/Source/Coordinator.swift
AdnanZiSAP: This was copied over. I left it as it was because there might be a reason someone left that commented-out. It's not a dead code, it's commented-out code. Dead code needs to be executable to be dead code and commented-out code isn't executable. :)

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: Coordinator is not only navigating, but _coordinating_ communication between different view controllers.
Coordinator can set state by design. It provides dependencies to view controllers and those dependencies usually control state and behaviour of view controller. For example, in model driven design, model contains state, and it's provided to the view controller by coordinator. 

In this case we have a scenario of two view controllers that need to share the same state. Coordinator is the right object to communicate that state between them IMO.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: This is used to add certain view controllers to list of objects that need an update of exposure notification setting when it's been updated. Those controllers are ```ExposureNotificationSettingViewController``` and ```SettingsViewController```. The update is triggered from ```ENStateHandler```.

I could argue this design (and I did actually, but arguments were on the other side :)), but this is again an old design that I just placed a bit differently here.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: Agree on this. I'm not a big fan of comments, as IMHO if something needs commenting then it's either too complex or too badly designed. I will come up with a better name for this instead.

src/xcode/ENA/ENA/Source/Coordinator.swift
AdnanZiSAP: This line should be removed now as it doesn't exist in develop anymore.

src/xcode/ENA/ENA/Source/Coordinator.swift
```diff
@@ -0,0 +1,280 @@
...
+//
+
+import UIKit
+import Connectivity
+
+protocol CoordinatorDelegate: AnyObject {
+	func coordinatorUserDidRequestReset()
+}
+
+class Coordinator: RequiresAppDependencies {
```
inf2381, line 27: It would be great to have some SwiftDoc comments here and also for the delegate protocol.

src/xcode/ENA/ENA/Source/Coordinator.swift
```diff
@@ -0,0 +1,280 @@
...
+			exposureManagerState: state.exposureManagerState,
+			detectionMode: state.detectionMode,
+			isLoading: isRequestRiskRunning,
+			risk: state.risk
+		)
+		exposureDetectionController?.state = state
+	}
+
+	func showExposureSubmission(with result: TestResult? = nil) {
+		// A strong reference to the coordinator is passed to the exposre submission navigation controller
```
inf2381, line 174: typo :)

AdnanZiSAP: haha, copied over with a typo. :) Will fix.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
AdnanZiSAP: Renamed [here](https://github.com/corona-warn-app/cwa-app-ios/pull/834/commits/5dcc8f8b5dd25b92193c97f15ed0bdb8ee3de90c).

src/xcode/ENA/ENA/Source/Coordinator.swift
AdnanZiSAP: Fixed [here](https://github.com/corona-warn-app/cwa-app-ios/pull/834/commits/8dd0bfadd2f934b7dd2e3307a125a2e331920873).

src/xcode/ENA/ENA/Source/Coordinator.swift
AdnanZiSAP: Added documentation comments [here](https://github.com/corona-warn-app/cwa-app-ios/pull/834/commits/dbb1c8dd3d4d072e93a46c35edb0c25220356199).

AdnanZiSAP:
@haosap @melloskitten Since you will not follow up on this PR, I was free to resolve your conversations.

inf2381 (approved):

haosap (approved):

haosap:
Let's merge it after 1.2 release.

ChristianKienle (approved):

PR #838, opened 2020-07-03
ExposureDetectionViewController refresh countdown

johannesrohwer:
## Description

This PR adds a countdown to the refresh/update button in the ExposureDetectionViewController (see JIRA 1627).

## Checklist


src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -20,7 +20,11 @@ import Foundation
 import UIKit
 
 final class ExposureDetectionViewController: DynamicTableViewController, RequiresAppDependencies {
-	// MARK: Properties
+	// MARK: Properties.
```
melloskitten, line 23: Any reason for not doing // MARK: _-_ foo bar baz ? (Note the dash -)

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -186,7 +190,7 @@ extension ExposureDetectionViewController {
 		tableView.reloadData()
 	}
 
-	private func updateCheckButton() {
+	private func updateCheckButton(_ time: String? = nil) {
```
melloskitten, line 193: A comment here, just shortly explaining what this method does (in case of side effects) might help

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
 			checkButton.isEnabled = true
 		case .manual:
 			footerView.isHidden = false
-			checkButton.setTitle(AppStrings.ExposureDetection.buttonRefresh, for: .normal)
-			checkButton.isEnabled = riskProvider.manualExposureDetectionState == .possible
+
+			let nextRefresh = riskProvider.nextExposureDetectionDate()
+			let now = Date()
+
+			if nextRefresh > now, countdown == nil {
```
melloskitten, line 211: Why must it explicitly be checked that the countdown is nil here?

melloskitten: Would prefer having `countdown == nil` linebreak

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+				scheduleCountdownTimer(to: nextRefresh)
+			} else if let countdown = countdown, countdown.end != nextRefresh {
```
melloskitten, line 213: same goes here with the linebreak

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+				// Make sure to schedule new countdown if the next refresh time has changed.
+				scheduleCountdownTimer(to: nextRefresh)
+			} else if nextRefresh - 1 > now {
```
melloskitten, line 216: why this arbitrary -1? whats the reason?

melloskitten: What was the reason?

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+				showCountdownButton(with: time)
+			} else {
+				showActiveRefreshButton()
+			}
```
melloskitten, line 220: tbh this entire method could use some comments.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+
+	private func scheduleCountdownTimer(to end: Date) {
+		countdown = CountdownTimer(countdownTo: end)
+		countdown?.delegate = self
+		countdown?.start()
+	}
+
+	private func showActiveRefreshButton() {
+		self.checkButton.setTitle(AppStrings.ExposureDetection.buttonRefresh, for: .normal)
+		self.checkButton.isEnabled = riskProvider.manualExposureDetectionState == .possible
```
melloskitten, line 232: It seems odd to me that a method that is called `showActiveRefreshButton()` uses a boolean check to set a button to an enabled or disabled state. Does that make sense to you either?

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+	}
+
+	private func showCountdownButton(with time: String? = nil) {
+		guard let time = time else { return }
+		UIView.performWithoutAnimation {
+			self.checkButton.setTitle(String(format: AppStrings.ExposureDetection.refreshIn, time), for: .normal)
+			self.checkButton.isEnabled = riskProvider.manualExposureDetectionState == .possible
```
melloskitten, line 239: Copy-pasta'ing from above, because this method has the same 'problem'
> It seems odd to me that a method that is called `showCountdownButton()` uses a boolean check to set a button to an enabled or disabled state. Does that make sense to you either?

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+		}
+	}
+}
+
+// MARK: CountdownTimerDelegate methods.
```
melloskitten, line 244: Please check whether you want to omit the dash from all MARK comments 👍

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -200,8 +204,51 @@ extension ExposureDetectionViewController {
...
+
+extension ExposureDetectionViewController: CountdownTimerDelegate {
+	func update(time: String) {
+		self.updateCheckButton(time)
+	}
+
+	func done() {
+		self.updateCheckButton()
```
melloskitten, line 252: Perhaps just making an `invalidateCheckButton()` instead of using the `updateCheckButton` would be more clean here. Then you also don't need to check in `updateCheckButton` whether time is nil or not. Also would be more self-explanatory imho.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskProvidingConfiguration.swift
```diff
@@ -41,12 +41,25 @@ extension RiskProvidingConfiguration {
...
+		guard let lastExposureDetectionDate = lastExposureDetectionDate else {
+			return Calendar.current.date(
+				byAdding: exposureDetectionInterval,
+				to: .distantPast,
+				wrappingComponents: false
+			) ?? .distantPast
+		}
+
+		// The lastExposureDetection should not be in the future.
+		guard lastExposureDetectionDate < currentDate else {
```
melloskitten, line 54: In which case can this happen?

src/xcode/ENA/ENA/Source/View Helpers/AppStrings.swift
```diff
@@ -207,6 +207,7 @@ enum AppStrings {
 		static let refreshedFormat = NSLocalizedString("ExposureDetection_Refreshed_Format", comment: "")
 		static let refreshedNever = NSLocalizedString("ExposureDetection_Refreshed_Never", comment: "")
 		static let refreshingIn = NSLocalizedString("ExposureDetection_RefreshingIn", comment: "")
```
melloskitten, line 209: What's the difference between "RefreshingIn" and "RefreshIn" ? seems a little bit confusing to me there.

src/xcode/ENA/ENA/Source/View Helpers/CountdownTimer.swift
```diff
@@ -0,0 +1,83 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+import Foundation
+
+// MARK: - CountdownTimer
```
melloskitten, line 22: any particular reason why you're putting a MARK in a class/file that is named like that anyways?

src/xcode/ENA/ENA/Source/View Helpers/CountdownTimer.swift
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	deinit {
+		timer?.invalidate()
+		timer = nil
+	}
+
+	func start() {
+		timer?.invalidate()
+		timer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true, block: action)
```
melloskitten, line 46: break would be nice here

src/xcode/ENA/ENA/Source/View Helpers/CountdownTimer.swift
```diff
@@ -0,0 +1,83 @@
...
+			timer?.invalidate()
+			self.delegate?.done()
+			return
+		}
+
+		self.update()
+	}
+
+	private func update() {
+		let components = Calendar.current.dateComponents([.hour, .minute, .second], from: Date(), to: end)
```
melloskitten, line 65: same goes with this linebreak here tbh but I leave it up to you

src/xcode/ENA/ENA/Source/View Helpers/CountdownTimer.swift
```diff
@@ -0,0 +1,83 @@
...
+
+	static func format(_ components: DateComponents) -> String {
+		let hours = String(format: "%02d", components.hour ?? 0)
+		let minutes = String(format: "%02d", components.minute ?? 0)
+		let seconds = String(format: "%02d", components.second ?? 0)
+		return "\(hours):\(minutes):\(seconds)"
+	}
+}
+
+// MARK: CountdownTimerDelegate
```
melloskitten, line 77: dash dash dash dash dash dash dash 😂

src/xcode/ENA/ENA/Source/View Helpers/__tests__/CountdownTimerTests.swift
```diff
@@ -0,0 +1,94 @@
...
+	var doneCallback: (() -> Void)?
+
+	func update(time: String) {
+		self.updateCallback?(time)
+	}
+
+	func done() {
+		self.doneCallback?()
+	}
+}
```
melloskitten, line 94: Thanks for writing tests! yay!

melloskitten (changes requested):
Thanks for your work! 🥳  Some minor issues here and there, would appreciate if you took a look 👍

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: i want to avoid scheduling multiple countdown timers

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: here i wanted to make sure that the button does not become clickable when the riskProvider does not explicitly allow updating. I add a comment to explain this defensive implementation.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: here we can safely disable the button. I adjusted my code accordingly

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskProvidingConfiguration.swift
johannesrohwer: i rewrote this code and one of the test made sure this is the behavior. I guess some other code calls this method with Date.distantPast when no other date is available.

src/xcode/ENA/ENA/Source/View Helpers/CountdownTimer.swift
johannesrohwer: looks nicer in the minimap

src/xcode/ENA/ENA/Source/View Helpers/AppStrings.swift
johannesrohwer: slightly different text strings. agree, its not too clear tho

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: i agree with your concerns, however the current way is safer when it comes to changing parameters, e.g. a user switching on automatic mode. I added comments to the updateCheckButton method to make the flow a bit more clear

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
melloskitten: great! thanks for adding the comment.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
melloskitten: Considering the state the  HomeVC + all of its related classes are, this defensive approach is alright. Thanks for clarifying.

melloskitten (approved):
Thanks for integrating the changes. Happy merging!

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: this prevents the countdown to show 00:00:00 as an explicit value and directly enables the button. looks nicer imo

steins-code (approved):

PR #864, opened 2020-07-07
RiskTests test errors

aleksandr-tikhonov:


src/xcode/ENA/ENA/Source/Services/Risk/__tests__/RiskTests.swift
```diff
@@ -24,7 +24,7 @@ import XCTest
 final class RiskTests: XCTestCase {
 	func testGetNumberOfDaysActiveTracing_LessThanOneDay() {
 		let details = mockDetails(hours: 23)
-		XCTAssertEqual(details.numberOfDaysWithActiveTracing, 0)
+		XCTAssertEqual(details.numberOfDaysWithActiveTracing, 1)
```
aleksandr-tikhonov, line 27: If you have a look at `ActiveTracing`, you will see 

```
	var inHours: Int {
		// Hours are intentionally rounded down.
		// We could also simply cast this to `Int` (what we actually do here as well)
		// but we still call rounded(…) to make it more explicit.
		Int((interval / TimeInterval.SEC_PER_HOUR).rounded(.down))
	}
	
	var inDays: Int {
		Int((interval / TimeInterval.SEC_PER_DAY).rounded(.toNearestOrAwayFromZero))
	}
```

I think, @ChristianKienle 's idas was round value to nearest number, so 23 hours should rounds to 1 day.

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/RiskTests.swift
```diff
@@ -38,7 +38,7 @@ final class RiskTests: XCTestCase {
 	}
 
 	func testGetNumberOfDaysActiveTracing_OneDayExact() {
-		let details = mockDetails(hours: 25)
```
aleksandr-tikhonov, line 41: I think, here's typo

steins-code (approved):

PR #866, opened 2020-07-07
Add home screen risk cell update button countdown

johannesrohwer:
## Description

This PR implements a countdown for the update button that is shown for the risk cell on the home screen. (see JIRA 1626)

## Checklist


src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -427,3 +431,30 @@ extension HomeInteractor {
...
+	func update(time: String) {
+		guard let indexPath = self.indexPathForRiskCell() else { return }
+		guard let cell = homeViewController.cellForItem(at: indexPath) as? RiskLevelCollectionViewCell else { return }
+
+		// We pass the time and let the configurator decide whether the button can be activated or not.
+		riskLevelConfigurator?.timeUntilUpdate = time
+		riskLevelConfigurator?.configureButton(for: cell)
+	}
+
+	func done() {
```
haosap, line 455: IMO, it's better to use the Apple style to name the delegate method, for example,
```swift
 func  countdownTimer(_ timer: CountdownTimer, didUpdate time: String)
```

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -427,3 +431,30 @@ extension HomeInteractor {
...
+
+// MARK: - CountdownTimerDelegate methods.
+
+/// The `CountdownTimerDelegate` is used to update the remaining time that is shown on the risk cell button until a manual refresh is allowed.
+extension HomeInteractor: CountdownTimerDelegate {
+	private func scheduleCountdownTimer() {
+		let nextUpdate = riskProvider.nextExposureDetectionDate()
+		countdownTimer = CountdownTimer(countdownTo: nextUpdate)
+		countdownTimer?.delegate = self
+		countdownTimer?.start()
```
haosap, line 443: I see, the timer `invalidate`  is only called in the CoutdownTimer class:
```swift
 deinit {
		timer?.invalidate()
		timer = nil
	} 
```
My question is, if the app goes background, the Timer still runs? What will happened if the user opens the App after 2 days?

haosap (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
johannesrohwer: I agree, I will adjust that!

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
johannesrohwer: thanks for bringing this up, I discovered a bug in my implementation where the timer was not stopped as I expected due to a reference cycle. I added a manual invalidate() method to make sure the timer is handled correctly.
Regarding the opening of the app after some time has passed: The timer is guaranteed to be up-to-date since the scene delegate will trigger the updateExposureState() method of the home VC when the app comes into foreground.
To make sure the timer does not drain the users battery when the app goes into background, I added a few lines that invalidate the handler when the app goes into background.

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeHighRiskCellConfigurator.swift
```diff
@@ -73,18 +73,30 @@ final class HomeHighRiskCellConfigurator: HomeRiskLevelCellConfigurator {
...
+
+	override func configureButton(for cell: RiskLevelCollectionViewCell) {
+		super.configureButton(for: cell)
+
 		let buttonTitle: String
 		if isLoading {
 			buttonTitle = AppStrings.Home.riskCardStatusCheckButton
 		} else {
-			let intervalDisabledButtonTitle = String(format: AppStrings.Home.riskCardIntervalDisabledButtonTitle, intervalString)
+			let intervalDisabledButtonTitle: String
```
melloskitten: I have added a commit that cleans up this part of the code. Please check if you're fine with it.

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
```diff
@@ -108,18 +108,29 @@ final class HomeLowRiskCellConfigurator: HomeRiskLevelCellConfigurator {
...
+
+	override func configureButton(for cell: RiskLevelCollectionViewCell) {
+		super.configureButton(for: cell)
+		
 		let buttonTitle: String
 		if isLoading {
 			buttonTitle = AppStrings.Home.riskCardStatusCheckButton
 		} else {
-			let intervalDisabledButtonTitle = String(format: AppStrings.Home.riskCardIntervalDisabledButtonTitle, intervalString)
+			let intervalDisabledButtonTitle: String
```
melloskitten: same as above, I have added a commit that cleans up this code. If you'd be so kind, please check it.

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeRiskLevelCellConfigurator.swift
```diff
@@ -109,8 +110,11 @@ class HomeRiskLevelCellConfigurator: HomeRiskCellConfigurator {
 		cell.bodyLabel.accessibilityIdentifier = AccessibilityIdentifiers.RiskCollectionViewCell.bodyLabel
 		cell.detectionIntervalLabel.accessibilityIdentifier = AccessibilityIdentifiers.RiskCollectionViewCell.detectionIntervalLabel
 		cell.updateButton.accessibilityIdentifier = AccessibilityIdentifiers.RiskCollectionViewCell.updateButton
-
 	}
+
+	/// Convenience method that can be overwritten to configure the button without running the full configure(_:) method.
+	/// This is handy when very frequent updates such as the update countdown are applied to the button.
+	func configureButton(for cell: RiskLevelCollectionViewCell) { }
```
melloskitten, line 117: I'm currently thinking whether this method could get a better name. Is this almost like a reload view method?

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeUnknownRiskCellConfigurator.swift
```diff
@@ -64,19 +64,29 @@ final class HomeUnknownRiskCellConfigurator: HomeRiskLevelCellConfigurator {
...
+
+	override func configureButton(for cell: RiskLevelCollectionViewCell) {
+		super.configureButton(for: cell)
+
 		let buttonTitle: String
 		if isLoading {
 			buttonTitle = AppStrings.Home.riskCardStatusCheckButton
 		} else {
-			let intervalDisabledButtonTitle = String(format: AppStrings.Home.riskCardIntervalDisabledButtonTitle, intervalString)
+			let intervalDisabledButtonTitle: String
```
melloskitten: same as above.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -262,11 +263,12 @@ extension ExposureDetectionViewController {
 // MARK: - CountdownTimerDelegate methods.
 
 extension ExposureDetectionViewController: CountdownTimerDelegate {
-	func update(time: String) {
+
+	func countdownTimer(_ timer: CountdownTimer, didUpdate time: String) {
```
melloskitten, line 267: why are you passing the timer, if you are not using it? is it necessary?

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -262,11 +263,12 @@ extension ExposureDetectionViewController {
...
 		self.updateCheckButton(time)
 	}
 
-	func done() {
+	func countdownTimer(_ timer: CountdownTimer, didEnd done: Bool) {
```
melloskitten, line 271: same goes here, is it necessary to pass the timer object?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -427,3 +431,43 @@ extension HomeInteractor {
 		homeViewController.showExposureNotificationSetting()
 	}
 }
+
+// MARK: - CountdownTimerDelegate methods.
```
melloskitten, line 435: good job with the comments. makes me happy to see. 💯

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -427,3 +431,43 @@ extension HomeInteractor {
...
+		countdownTimer?.delegate = self
+		countdownTimer?.start()
+	}
+
+	@objc
+	private func invalidateCountdownTimer() {
+		countdownTimer?.invalidate()
+	}
+
+	func countdownTimer(_ timer: CountdownTimer, didEnd done: Bool) {
```
melloskitten, line 459: it seems to me as if the timer is not needed. please check.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -139,15 +139,18 @@ extension RiskProvider: RiskProviding {
 
 	/// Returns the next possible date of a exposureDetection
 	/// Case1: Date is a valid date in the future
-	/// Case2: Date is in the past (could be .distantPast) to indicate that a detection may be performed
-	/// Case2 can only happen when shouldPerformExposure returns true
+	/// Case2: Date is in the past (could be .distantPast) (usually happens when no detection has been run before (e.g. fresh install).
+	/// For Case2, we need to calculate the remaining time until we reach a full 24h of tracing.
```
melloskitten, line 143: nice. comments rule. yay! 👍  🥇

melloskitten (changes requested):
I have added some comments and cleaned up some of the code. Would be cool if you looked through these commits, and checked whether they make sense to you. Thanks! 👍 

In case there aren't enough tests: Please make sure to add tests, cheers! 😄

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
haosap: This is a kinda pattern of delegate method. It's always good to pass the "event source" or "sender".

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeLowRiskCellConfigurator.swift
johannesrohwer: thanks for improving this :)

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: the timer is not used here, however I am trying to stick to the apple style delegates as Hao proposed where you get passed a reference to the caller as first parameter.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
melloskitten: Hm. But isn't this against YAGNI principle?

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
melloskitten: Same comment I gave before to Hao, what about the YAGNI principle?

src/xcode/ENA/ENA/Source/Models/Home/HomeRiskCellConfigurator/HomeRiskCellConfigurators/HomeRiskLevelCellConfigurator.swift
melloskitten: Checked with @johannesrohwer, this seems to be intrinsinc to how the rest of the home vc works. Every change has to configure the cell anew.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
johannesrohwer: It is roughly following the guidelines by apple:
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html

I agree it is not necessarily required, however I would favour to follow the convention and pass the reference as first parameter.

melloskitten: Thank you for clarifying. I agree with you.

melloskitten (approved):

haosap (approved):
Well done!💯

PR #810, opened 2020-06-30
Update home screen on the app start

aleksandr-tikhonov:
This PR closes #770.

Previously `SceneDelegate` provided Risk equals `nil` to `HomeIntercator` when starting the application, `HomeIntercator` calculated `RiskLevel` as `.unknownInitial` if `Risk` is `nil` and we see "grey" screen. Then we're getting updated Risk from `riskProvider` and it triggers update ui of home screen with new `Risk`. That's why we see this quick ui update from "grey" to another risk.

This PR removed default `Risk` from `HomeInteractor`, so only `riskProvider` is responsible for providing Risk for the app. When app starts, `Risk` is `nil` and we don't show any risk cell, when we're getting updated Risk from `riskProvider` we update home screen.

Please review this PR carefully, thanks.
 
***Updated***

Also this PR fixes issue when risk detail screen shows _inactive_ state, but home screen doesn't

PR #874, opened 2020-07-08
Home Screen Risk Cell Update on App start + Risk Calc Refinements

### LINKED ISSUE #770: [UI] Load risk view before rendering

SebastianWild:
## Description

This PR is very closely related to #810 - this PR just adds one more commit and changes the base from `develop` to `staging` as there is an important fix in there. After speaking with @inf2381 we determined to not rebase #810 but to cherry-pick the changes. 

*Leaving as WIP/Draft pending team feedback on approach*

I'll copy the description of that PR here, then add the changes I made on top:

>This PR closes #770.
>
>Previously `SceneDelegate` provided Risk equals `nil` to `HomeIntercator` when starting the application, `HomeIntercator` calculated `RiskLevel` as `.unknownInitial` if `Risk` is `nil` and we see "grey" screen. Then we're getting updated Risk from `riskProvider` and it triggers update ui of home screen with new `Risk`. That's why we see this quick ui update from "grey" to another risk.
>
>This PR removed default `Risk` from `HomeInteractor`, so only `riskProvider` is responsible for providing Risk for the app. When app starts, `Risk` is `nil` and we don't show any risk cell, when we're getting updated Risk from `riskProvider` we update home screen.
>
>Please review this PR carefully, thanks.
> 
>***Updated***
>
>Also this PR fixes issue when risk detail screen shows _inactive_ state, but home screen doesn't

In addition to this, some adjustments were made to the `RiskProvider`.
In the current code, the precondition check is done as follows:
```
guard numberOfEnabledHours >= TracingStatusHistory.minimumActiveHours else {
    completeOnTargetQueue(risk: nil)
    return
}
```

This was fine before, but with the other changes as part of this PR, a `nil` Risk returned will cause the Home Screen to not display a risk cell at all. However some preconditions (like the `TracingStatusHistory.minimumActiveHours` not being met above), actually return valid risk levels. Instead of simply returning nil, https://github.com/corona-warn-app/cwa-app-ios/pull/874/commits/782b79dfabba26ecea213483e199ad64a32f7e27 returns these as actual risks should their conditions apply.

## Known Issues

This PR changes the UX of the home screen in some significant ways, which we should consider more carefully before merging. Previously, the home screen pretty much always shows a risk cell, but now if there is a `nil` risk, nothing is shown.


ChristianKienle (approved):
LGTM 😘

inf2381 (approved):

PR #884, opened 2020-07-10
Display underline in ENATanInput when character is invalid

eisvogel:
This PR addresses [JIRA-1602](https://jira.itc.sap.com/browse/EXPOSUREAPP-1602), thereby displaying the underline in the `ENATanInput` when the entered character is invalid.


src/xcode/ENA/ENA/Source/Views/BaseElements/ENATanInput.swift
```diff
@@ -346,7 +346,7 @@ private class ENATanInputLabel: UILabel {
 
 		self.textColor = textColor
 
-		if false != text?.isEmpty {
+		if false != text?.isEmpty || !isValid {
```
johannesrohwer, line 349: this does the job but is hard to read. is there a way to make this nicer? if there is no way, maybe add a comment?

johannesrohwer (changes requested):
Your PR seems to resolve the bug that was described in 1602. 🥳  Just a minor comment from my side

src/xcode/ENA/ENA/Source/Views/BaseElements/ENATanInput.swift
MichaelPFrey:
```suggestion
		if text?.isEmpty ?? true || !isValid {
```
this assumes that the reader understands `Optionals` in general and `??` specifically.
```suggestion
		if nil == text || text!.isEmpty || !isValid {
```
this assumes that the reader (including code quality QA tools) is aware of _short circuit evaluation_, as without `nil == text ||` the `text!.isEmpty ` would potentially result in an runtime error. `!.isEmpty` could be read as "not empty" by a non-swift programmer. To prevent that, brackets could be used:
```suggestion
		if nil == text || (text!).isEmpty || !isValid {
```
I am not a swift programmer. For me, my third suggestion is easiest to read, as I only would have to look up what the `!` in `(text!)` means.

haosap: Hi @EisvogelSAP ,
Would you mind adding some comment in your code which helps the upcoming developers to maintain it.
Thanks.

haosap (approved):

haosap:
Hi @johannesrohwer ,
Please recheck the changes. Thanks.

johannesrohwer (approved):
Awesome, thank you for rewriting the expression! 🎉

PR #945, opened 2020-07-24
Unknown Outdated Risk Detail Screen

eisvogel:
This PR addresses [JIRA-1555](https://jira.itc.sap.com/browse/EXPOSUREAPP-1555):
- updated risk unknown outdated after 48 hours screen
- enabled manual refresh button for unknown outdated state
- minor refactoring for colors, reducing complexity in `DynamicTableViewModel`

In this PR only UI new UI requirements are implemented. While reviewing please ensure to check if the underlying logic supports the new user story as well.


AdnanZiSAP:
Since I'm doing the logic, I will also check this one.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -201,8 +201,11 @@ extension ExposureDetectionViewController {
 			checkButton.setTitle(AppStrings.ExposureDetection.buttonEnable, for: .normal)
 			return
 		}
-		
-		switch state.detectionMode {
+
+		var mode = state.detectionMode
+		if .unknownOutdated == state.risk?.level { mode = .manual }
```
AdnanZiSAP, line 206: This looks like to be unnecessary. This state should never happen bar bugs in background task execution.

EisvogelSAP: Technically I agree, but this was the requirement I got. If for whatever reason the automatic update does not work the use needs a way of manually triggering the update.

AdnanZiSAP (approved):

AdnanZiSAP:
Approved. Some slight changes in logic are needed for this to work as expected (previous risk state is wrong), but UI-wise looks good. This needs to be merged so that the upcoming PR for logic could be tested.

EisvogelSAP:
@AdnanZiSAP thanks for the review. Maybe it would be wiser to merge this PR into the upcoming PR to avoid a "polluted" dev branch. We don't want to break functionality, do we?

AdnanZiSAP:
> @AdnanZiSAP thanks for the review. Maybe it would be wiser to merge this PR into the upcoming PR to avoid a "polluted" dev branch. We don't want to break functionality, do we?

@EisvogelSAP You're welcome. I'm also fine with that, but nothing will be broken with this. The correct card will be shown, only some info inside it will be wrong. And currently a totally wrong card is shown. So we might go ahead and merge this first as well. 

We first need second approval anyway. @SebastianWild could you please have a look at this.

SebastianWild (commented):
UI + refactoring looks good to me.

I'm guessing the translated strings are not in scope for this PR?
I ask because ex. a device in English language still presents a German string in the unknown risk card. Also tapping into the unknown risk card detail still mentions "Your exposure logging could not be updated for more than 48 hours", not sure if that's the right message there either.

If these changes are out of scope, I think we're good to go!

AdnanZiSAP:
German string is a bug that's been handled elsewhere. If it's on Home screen.

SebastianWild (approved):
Looks good, just need a few translations (ex. unknown outdated risk detail page).

PR #966, opened 2020-07-28
Risk calculation manually - after 2 days inactivity update not easy to find - the state switching logic

AdnanZiSAP:
# What this PR does

Implements missing logic for the story [#1555](https://jira.itc.sap.com/browse/EXPOSUREAPP-1555). On home screen, only the new introduced configurator is used for ```unknownOutdated``` state. On Risk detail screen, everything was already in place, only the previous risk label showed the wrong state, which is now fixed.

**For this logic to be tested, the #945 PR needs to be merged first (or merged here), so that Risk Detail screen UI changes are in place. The PR for Home screen UI changes is already merged.
You can test this by merge branch for #945 into this branch locally, like Sebastian already did.**

## How to test this PR

1. Install the app.
2. Disable background refresh for the app.
3. Make sure the risk update has been performed at least once. If not hit the Update button. If the update button is not enabled (fresh install), you can hack it by changing the date on the device to more than 24 h in future. 
4. To make sure the new card is *not* shown before 48 h from the previous update passes, change the date on the device to say 25 h in the future. The Risk card shouldn't change. 
5. Again change the date on the device, this time to 49 h in the future (or more) from the original state (at least one update already performed- step 2), and verify that on both Home screen and Risk detail screen, the new card from the [#1555](https://jira.itc.sap.com/browse/EXPOSUREAPP-1555) is shown.


SebastianWild:
I merged #945 into this branch and did a test, looks good to me.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -212,11 +212,13 @@ extension HomeInteractor {
...
-				previousRiskLevel: store.previousRiskLevel,
-				lastUpdateDate: dateLastExposureDetection
-			)
+			riskLevelConfigurator = HomeUnknown48hRiskCellConfigurator(
+				isLoading: isRequestRiskRunning,
+				lastUpdateDate: dateLastExposureDetection,
+				detectionInterval: detectionInterval,
+				detectionMode: detectionMode,
+				manualExposureDetectionState: riskProvider.manualExposureDetectionState,
+				previousRiskLevel: store.previousRiskLevel)
```
aleksandr-tikhonov, line 221: `HomeUnknown48hRiskCellConfigurator ` completely replaced `HomeInactiveRiskCellConfigurator`?

if `HomeInactiveRiskCellConfigurator` isn't used anymore, you can delete it from the project

aleksandr-tikhonov (approved):

SebastianWild (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
AdnanZiSAP: I will check it. Thanks.

AdnanZiSAP: ```HomeInactiveRiskCellConfigurator``` is still used for ```inactive``` state.  I knew I saw it somewhere else. :) Which I good because I have a feeling that the story with risk calculation states will not end with this. :)

aleksandr-tikhonov (approved):

PR #1010, opened 2020-08-05
Add Swift Doc and Comments to Risk Calculation

haosap:
The following actions will be done in the PR:
1. Add more comments and API Doc to all risk calculation related classes.
2. Code refactoring to risk calculation

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
@@ -21,7 +21,7 @@ import Foundation
 import UIKit
 
 /**
-An instance of RiskConsumer is an object which is interesting in the result of risk calculation.
+An instance of RiskConsumer is an object which is interested in the result of risk calculation.
 One **RiskProvider** can have more **RiskConsumer** which means one RiskProvider can offer more calculation results for more RiskConsumers
 */
 
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
@@ -18,6 +18,12 @@
 //
 
 import Foundation
+import UIKit
+
+/**
+An instance of RiskConsumer is an object which is interesting in the result of risk calculation.
+One **RiskProvider** can have more **RiskConsumer** which means one RiskProvider can offer more calculation results for more RiskConsumers
+*/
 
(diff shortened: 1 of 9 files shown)


src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
```diff
@@ -18,6 +18,12 @@
 //
 
 import Foundation
+import UIKit
+
+/**
+An instance of RiskConsumer is an object which is interesting in the result of risk calculation.
```
MichaelPFrey, line 24: typo?
https://dictionary.cambridge.org/de/worterbuch/englisch/interesting
https://dictionary.cambridge.org/de/worterbuch/englisch/interested
```suggestion
An instance of RiskConsumer is an object which is interested in the result of risk calculation.
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -87,7 +117,11 @@ extension RiskProvider: RiskProviding {
 			lastExposureDetectionDate: store.summary?.date)
 	}
 
+
+
 	/// Called by consumers to request the risk level. This method triggers the risk level process.
+	/// - Parameter userInitiated: If true, it's called by user, if false, called by task scheduler
+	/// - Parameter completion: It is an optional value, if it is set to nil, the attachted RiskConsumers will be notified once the risk calculation is done.
```
MichaelPFrey, line 124: typo?
```suggestion
	/// - Parameter completion: It is an optional value, if it is set to nil, the attached RiskConsumers will be notified once the risk calculation is done.
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -21,16 +21,46 @@ import Foundation
...
+
+// 2. Attach the callback block to handle the result
+riskConsumer.didCalculateRisk = {risk in
+    // bla bla bla
+}
+
+// 3. Obtain the instance of RiskProvider and attach the RiskConsumer to the RiskProvider
+riskProvider.observeRisk(riskConsumer)
 
+// 4. Ask for the result
```
melloskitten, line 51: General question: Why are some comment blocks done with `/*    */' and some with `///` ?

haosap: In general, they're equivalent. For multilines, /** */ is better. /// is suitable for single line doc.

PR #1024, opened 2020-08-07
Support for redeemed test result

melloskitten:
## Description

This PR adds compatibility for the 'redeemed' test result. You can find information about this [here](https://jira.itc.sap.com/browse/EXPOSUREAPP-1743). Thanks for reviewing.


thomasaugsten (approved):

aleksandr-tikhonov (approved):

src/xcode/ENA/ENA/Resources/Localization/ar.lproj/Localizable.strings
```diff
@@ -461,6 +461,10 @@
 
 "ExposureSubmissionResult_testInvalidDesc" = "هناك مشكلة في تقييم اختبارك. الرجاء الاتصال بدائرة الصحة العامة لمعرفة كيفية المتابعة.";
 
+"ExposureSubmissionResult_testRedeemed" = "Test has errors";
+
+"ExposureSubmissionResult_testRedeemedDesc" = "There was a problem evaluating your test. Your QR Code has already expired.";
```
haosap, line 466: Hi @steins-code 
Is it against the translation process?

thomasaugsten: What is the exact issue because we have to create placeholder in all language files?

haosap: We have some internal process/agreement with translation team. @steins-code will inform them and I'll inform QA lead to avoid some unnecessary bug tickets.

PeterBuchholz: Talked to translation colleagues: Sabine Loss and Gudrun Wolber.
Outcome was that we only need the new text in german.

PR #1031, opened 2020-08-10
Introduce separate property to enable background fake requests

johannesrohwer:
## Description

This PR adds a dedicated property to enable or disable background fake requests


melloskitten (approved):
looks good to me! 👍

aleksandr-tikhonov (approved):

PR #1036, opened 2020-08-11
Dead man notifications

johannesrohwer:
## Description

This PR adds dead man notifications as described in JIRA 2044 and decouples the different background sub tasks.

# Notes
The PR is on hold until the following items have been checked:
* [x] Final texts
* [ ] Dev testing

## Checklist


src/xcode/ENA/ENA/Source/AppDelegate+ENATaskExecutionDelegate.swift
```diff
@@ -24,70 +24,79 @@ import UIKit
...
-
-			// NOTE: We are currently fetching the test result first, and then execute
-			// the exposure detection check. Instead of implementing this behaviour in the completion handler,
-			// queues could be used as well. Due to time/resource constraints, we settled for this option.
-			self.executeExposureDetectionRequest(task: task) { exposureDetectionSuccess in
-				self.executeFakeRequests() {
-					completion(fetchTestResultSuccess && exposureDetectionSuccess)
-				}
-			}
+	func executeENABackgroundTask(completion: @escaping ((Bool) -> Void)) {
```
haosap, line 27: what is the purpose of this method? From the comment, I cannot figure it out. All the results are just discarded.

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -59,20 +59,26 @@ final class ENATaskScheduler {
 		let identifierString = taskIdentifier.backgroundTaskSchedulerIdentifier
 		BGTaskScheduler.shared.register(forTaskWithIdentifier: identifierString, using: .main) { task in
 			self.scheduleTask()
+			let backgroundTask = DispatchWorkItem {
+				execute(task)
+			}
```
haosap, line 64: The reason you import DispatchWorkItem is because you want to cancel the task later or get the completion signal?

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -59,20 +59,26 @@ final class ENATaskScheduler {
...
+
 			task.expirationHandler = {
-				logError(message: "ERROR: Task has expired.")
 				self.scheduleTask()
+				backgroundTask.cancel()
```
haosap, line 68: One reminder:
Cancellation does not affect the execution of a work item that has already begun.

haosap (changes requested):
Please check some comments. thanks. 💯

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
johannesrohwer: I introduced the DispatchWorkItem here in order to allow future cancellation of the BG task and its subtasks. This is currently not really functional since the underlying tasks do not allow preemptive cancellation. However, for the exposure detection task interruption without cancellation may be a source or undefined behavior or bugs as described by chris k. I therefore wanted to prepare the background tasks to be able to forward the cancel signal that is provided by the expirationHandler.

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
johannesrohwer: absolutely right! see my other comment about the DispatchWorkItem for more clarifications

src/xcode/ENA/ENA/Source/AppDelegate+ENATaskExecutionDelegate.swift
johannesrohwer: Agree, the method is a bit confusing- I added a comment that explicitly states that we always call completion(true) when all subtasks finished, regardless of their individual success. In this context here, completion(true) indicates nothing but that all subtasks were able to finish.

haosap (approved):

melloskitten (approved):

PR #1038, opened 2020-08-11
Link contact form on imprint page

steins-code:
https://jira.itc.sap.com/browse/EXPOSUREAPP-1654

In this PR the phone numbers shall be removed from the imprint page in app information.
Instead, links should add the possibility to get linked to a contact form.
IMG_38339AD37A31-1


johannesrohwer (approved):
> I just tested your changes, seems good to me! 💯  Just a small question: is the link supposed to be opened in directly in the browser instead of an in-app SFSafariViewController?

My understanding of the user story was it should really be only a link without using SFSafariViewController

tcfos (approved):

PR #1042, opened 2020-08-12
[OUTDATED] Test result header wraps around (closes 2065 and 2125 )

melloskitten:
This PR handles the following issues [here](https://jira.itc.sap.com/browse/EXPOSUREAPP-2125) and [here](https://jira.itc.sap.com/browse/EXPOSUREAPP-2065). Please check out the screenshots linked in these issues. 
The problem before was that the test result headers would truncate instead of wrap around.

Here is how it looks like now, on a iPhone 8 plus. 
If you can, please test on the small phone sizes. Preferably SE.
Thanks so much.


johannesrohwer (approved):
I tested this on an iPhone SE with maximum font size. Thanks for your changes 💯

haosap (approved):
Thanks for quick fix!

PR #1048, opened 2020-08-13
added missed "inactive" risk card

aleksandr-tikhonov:
Hi, this PR adds missed "Risiko-Ermittlung nicht möglich" risk card. According to below specification we should show this risk card for automatic mode and `.unknownOutdated` risk level.

###  Automatic mode 

Screenshot 2020-08-13 at 16 23 48

###  Manual mode 

Screenshot 2020-08-13 at 16 24 13

### Screenshot of this risk card

Screenshot 2020-08-13 at 16 17 45


src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -218,14 +218,23 @@ extension HomeInteractor {
...
+				inactiveConfigurator?.activeAction = inActiveCellActionHandler
+			} else {
+				riskLevelConfigurator = HomeUnknown48hRiskCellConfigurator(
+					isLoading: isRequestRiskRunning,
+					lastUpdateDate: dateLastExposureDetection,
+					detectionInterval: detectionInterval,
+					detectionMode: detectionMode,
+					manualExposureDetectionState: riskProvider.manualExposureDetectionState,
+					previousRiskLevel: store.previousRiskLevel)
+				inactiveConfigurator?.activeAction = inActiveCellActionHandler
```
ArturFriesen, line 236: Is this line needed? It looks like, in this case riskLevelConfigurator is returned and changing inactiveConfigurator has no effect.

marcussc:
Hmm I don't seem to understand how this can happen and if it does happen if the call to action is the right one. 



On the top users sees that Risiko-Ermittlung is a active, but the call for action is "Risiko-Ermittlung einschalten" 🤔

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
aleksandr-tikhonov: You're right, I remove this line of code

aleksandr-tikhonov:
> Hmm I don't seem to understand how this can happen and if it does happen if the call to action is the right one.
> 
> On the top users sees that Risiko-Ermittlung is a active, but the call for action is "Risiko-Ermittlung einschalten" 🤔

Oh, I think, I'd taken this screenshot with some modifications in the code like this

`let riskLevel: RiskLevel? = .unknownOutdated // state.exposureManagerState.enabled ? state.riskLevel : .inactive`
in `setupRiskConfigurator()`

it leads to misleading, I need to delete the screenshot or update it.

marcussc (approved):

ArturFriesen (approved):

PR #1057, opened 2020-08-17
[2214] Adopted current plausible deniability configuration (disabling PD)

whiskey:
## Description
Disabling all fake requests as per [EXPOSUREAPP-2214](https://jira.itc.sap.com/browse/EXPOSUREAPP-2214)

## Checklist

* [x] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)


haosap (approved):

src/xcode/ENA/ENA/Source/AppDelegate+PlausibleDeniability.swift
```diff
@@ -65,17 +67,19 @@ extension AppDelegate {
...
-			UIApplication.shared.applicationState == .active,
-			Int.random(in: 1...100) == magicNumber
-		else { return }
 
-		sendFakeRequest()
+	/// Randomly execute a fake request
+	/// - Parameter propability: the probability p to execute a fake request. Accepting values between 0 and 1.
+	/// - Returns: Bool to indicate wether the fake request has been sent
+	@discardableResult
+	func executeFakeRequestOnAppLaunch(propability p: Double) -> Bool {
```
nickguendling, line 75:
```suggestion
	func executeFakeRequestOnAppLaunch(probability p: Double) -> Bool {
```
```suggestion
	func executeFakeRequestOnAppLaunch(propability p: Double) -> Bool {
```

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -89,7 +89,8 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 	func sceneDidBecomeActive(_: UIScene) {
 		hidePrivacyProtectionWindow()
 		UIApplication.shared.applicationIconBadgeNumber = 0
-		(UIApplication.shared.delegate as? AppDelegate)?.executeFakeRequestOnAppLaunch()
+		// explicitely disabled as per #EXPOSUREAPP-2214
+		(UIApplication.shared.delegate as? AppDelegate)?.executeFakeRequestOnAppLaunch(propability: 0.0)
```
nickguendling, line 93:
```suggestion
		(UIApplication.shared.delegate as? AppDelegate)?.executeFakeRequestOnAppLaunch(probability: 0.0)
```

src/xcode/ENA/ENA/Source/AppDelegate+PlausibleDeniability.swift
```diff
@@ -65,17 +67,19 @@ extension AppDelegate {
...
+		assert(p <= 1, "p should be lighter than or equal 1.0")
```
nickguendling, line 76:
```suggestion
		assert(p <= 1, "p should be lower than or equal 1.0")
```

src/xcode/ENA/ENA/Source/AppDelegate+PlausibleDeniability.swift
```diff
@@ -65,17 +67,19 @@ extension AppDelegate {
...
+		assert(p >= 0, "p should be greater than or equal 0.0")
+		if Double.random(in: 0...1) <= p {
```
nickguendling, line 78: If I'm not mistaken, this can still be executed even if p is set to 0.0, because Double.random(in: 0...1) can return 0

src/xcode/ENA/ENA/Source/AppDelegate+PlausibleDeniability.swift
whiskey: good catch, thanks!

src/xcode/ENA/ENA/Source/AppDelegate+PlausibleDeniability.swift
whiskey: I was wondering the same but it seems the bounds seem not to be reached for `p = 0.0`. Due to time and other constraints, I haven't added a unit test (yet) but here is my local test for this implementation: 

```swift
// Will never trigger, even for 100M iteration. 
// But will work on very small values like `let p = 0.0.nextUp` (5e-324)
let p = 0.0

let total: Int = 1_000_000
var hits: Int = 0

for _ in 0...total {
    if Double.random(in: 0...1) <= p {
        hits += 1
    }
}

print("\(hits)/\(total) => \(Double(hits)/Double(total))")
```

whiskey: I would consider a compiler 'optimization' here and would expect hits with `NSDecimalNumbers` instead of `Double`.

To short cut the discussion, we could go with:
```suggestion
		if Double.random(in: 0.0.nextUp...1) <= p {
```

nickguendling: The nextUp solution looks good to me, just to be safe 👍

ArturFriesen (approved):

PR #1060, opened 2020-08-17
Revision of risk calculation method

aleksandr-tikhonov:
Hi,

This PR fill in the gap between Android and iOS implementations of _RiskCalculation_. Previously `.unknownOutdated` always overrules `.increased`, but it should do it only in one case when we don't have `.unknownInitial` conditions  and in other cases `.increased` overrules `.unknownOutdated`.

I added corresponding test case in _RiskCalculationTests_.

Here is description of risk levels, rules and the hierarchy of it from Android team.
[RiskLevelCalculation Handover.pdf](https://github.com/corona-warn-app/cwa-app-ios/files/5089758/RiskLevelCalculation.Handover.pdf)

BR


src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
```diff
@@ -75,33 +74,34 @@ enum RiskCalculation {
...
-			// This overrides all other levels
-			return .success(.inactive)
-		}
+		let isInactive = !preconditions.isGood
 
 		// Precondition 2 - If tracing is active less than 1 day, risk is .unknownInitial
-		if activeTracing.inHours < minTracingActiveHours, riskLevel < .unknownInitial {
-			riskLevel = .unknownInitial
-		}
+		let isTracingActiveLess1day = activeTracing.inHours < minTracingActiveHours
```
marcussc, line 82:
```suggestion
		let isTracingActiveLess1Day = activeTracing.inHours < minTracingActiveHours
```

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
```diff
@@ -75,33 +74,34 @@ enum RiskCalculation {
...
-		}
+		let isUnknownInitial = isTracingActiveLess1day || isNoSummary
 
-		guard let summary = summary else {
-			return .success(riskLevel)
+		let isUnknownOutdated = !providerConfiguration.exposureDetectionIsValid(lastExposureDetectionDate: dateLastExposureDetection ?? .distantPast)
+
+		var riskLevels: [RiskLevel] = [.low]
+
+		// returns RiskLevel with higher priprity
```
marcussc, line 93:
```suggestion
		// returns RiskLevel with higher priority
```

aleksandr-tikhonov: thanks

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
aleksandr-tikhonov: fixed

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/__tests__/RiskCalculationTests.swift
```diff
@@ -332,6 +332,29 @@ final class RiskCalculationTests: XCTestCase {
 		XCTAssertTrue(risk?.riskLevelHasChanged == false)
 	}
 
+	func testCalculateRisk_IncreasedOverridesUnknownOutdated() {
```
SebastianWild, line 335: If I understand correctly, for `.increased` to override `.unknownOutdated`, we need all three things to be `true`:

- Date of last exposure detection is older than 48 hours.
- Exposure detection returned a summary which will lead to `.increased` risk.
- `.unknownInitial` state is present, with tracing being active less than 24 hours.

According to these rules, I think we need to adjust this unit test in the following way:
- Change `dateLastExposureDetection` parameter to something like `Date().addingTimeInterval(.init(hours: -25))`, to nudge it over the limit defined in the `RiskProvidingConfiguration` on line 337
- Change `activeTracing` parameter to something like `.init(interval: 2 * 3600)` so that `.unknownInitial` condition triggers.

I think then this unit test covers the three conditions listed above (it still passes with these changes though 👍) We're getting super deep into edge cases here but that's good 😀

Is my thinking correct?

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
```diff
@@ -75,33 +74,34 @@ enum RiskCalculation {
...
-		guard let summary = summary else {
-			return .success(riskLevel)
+		let isUnknownOutdated = !providerConfiguration.exposureDetectionIsValid(lastExposureDetectionDate: dateLastExposureDetection ?? .distantPast)
+
+		var riskLevels: [RiskLevel] = [.low]
+
+		// returns RiskLevel with higher priority
+		var riskLevel: RiskLevel {
+			riskLevels.max() ?? .inactive
 		}
```
SebastianWild, line 96: This approach is cool, I've never actually used such a computed variable before - never thought of it 👍

SebastianWild (changes requested):
I like the refactoring approach!
If my thinking is correct, we might have to slightly tweak that one added unit tests - please take a look at my comment. 

We should also adjust some of the comments - I will push a commit to do that.

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculation.swift
aleksandr-tikhonov: I also have never used it before, just tried to write it and surprisingly it works 😅

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/__tests__/RiskCalculationTests.swift
aleksandr-tikhonov: so I understand it this way - `.increased` overrides `.unknownOutdated` in all cases, except one case.

Here is the case:

- `.unknownOutdated` state is present
- `.increased` is present
- `.unknownInitial` is **NOT** present

It's Scenario 3 from the document. So the result will be `.unknownOutdated` state.

In all other cases the result will be `.increased` state, for example:

1) 
- `.unknownOutdated` state is present
- `.increased` is present
- `.unknownInitial` is present
2)
- `.unknownOutdated` state is **NOT** present
- `.increased` is present
- `.unknownInitial` is **NOT** present
3)
- `.unknownOutdated` state is **NOT** present
- `.increased` is present
- `.unknownInitial` is present

In all these 3 cases the result will be `.increased` state.

So this unit test covers 2) case and you described 1) case.

:)

aleksandr-tikhonov: I added additional 2 unit tests for it

SebastianWild (approved):

PR #1073, opened 2020-08-21
[Proposal] handling for local settings and parameters

whiskey:
## Description

We previously used static strings to define host URLs within the `HTTPClient`. I would like to propose an alternative way similar to `.env` files in backend development. The goal is to prevent cluttered code, 'commit wars' or even a deployment to the wrong environment.

This approach uses the `LocalSettings.swift` file as reference for all kind of parameters such as URLs. The benefit is that this bears no danger of accidentally commit custom settings, e.g. `http://localhost`, to the main repository. 

### Alternatives

The alternative approach to use environment variables and address these in code is also viable but also has the potential to 'leak' unwanted Xcode schemes if the developer activates the 'shared' option on that scheme.

### Prerequisites

There is no further action to 'enable' these local settings. If no `LocalSettings.swift` file is present a default version will be created.

### Caveats

In case this system is adopted, all additional settings have to be added manually as there is no mechanism in place to 'fill up' the user's local settings with missing keys.

## Checklist

* [x] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)


ChristianKienle (commented):
I think this will break our CI. Double checking with Marc.

whiskey:
This would break the CI, thanks @ChristianKienle! 
I made a small change to the CI script. Without your note this would have been missed. Should we also tweak the script or tests to handle this case?

ChristianKienle (approved):

inf2381 (approved):

PR #1105, opened 2020-09-01
removed property from AppDelegate that never gets written to

marcussc:
## Description
Removed `exposureSubmissionService` from AppDelegate, since it never gets written.


ArturFriesen (approved):

whiskey (approved):
AppCode user? 😜

PR #1140, opened 2020-09-07
Feature/2525 calender component

nickguendling:
## Description
This is the calendar component for the onset of symptoms feature (https://jira.itc.sap.com/browse/EXPOSUREAPP-2525)


src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
```diff
@@ -0,0 +1,99 @@
...
+//
+
+import UIKit
+import Combine
+
+class DatePickerDayView: UIView {
+
+	// MARK: - Init
+
+	@available(*, unavailable)
```
ArturFriesen, line 27: Nice  👍

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/__tests__/DatePickerOptionViewModelTests.swift
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	func test() {
+		let viewModel = DatePickerOptionViewModel(
+			today: Date(timeIntervalSinceReferenceDate: 28 * 24 * 60 * 60),
+			calendar: .gregorian(with: .init(identifier: "en-US"))
+		)
+
+		XCTAssertEqual(viewModel.weekdays.count, 7)
+		XCTAssertEqual(viewModel.weekdays, ["S", "M", "D", "M", "D", "F", "S"])
```
ArturFriesen, line 101: 🤔 Shouldn't this return ["S", "M", "T", "W", "T", "F", "S"] with "en-US" ?

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
```diff
@@ -0,0 +1,99 @@
...
+
+	private var subscriptions = [AnyCancellable]()
+
+	private let titleLabel = DynamicTypeLabel()
+
+	private func setUp() {
+		titleLabel.translatesAutoresizingMaskIntoConstraints = false
+		addSubview(titleLabel)
+
+		NSLayoutConstraint.activate([
```
melloskitten, line 65: god bless,  programmatic constraints!!! 💯

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
```diff
@@ -0,0 +1,99 @@
...
+		layer.cornerRadius = bounds.width / 2
+	}
+
+	// MARK: - Private
+
+	private let viewModel: DatePickerDayViewModel
+
+	private var subscriptions = [AnyCancellable]()
+
+	private let titleLabel = DynamicTypeLabel()
```
melloskitten, line 59: Could you please check whether using ENALabel would make your life easier here? That class has a lot of pre-defined UI looks that were checked with the designers - perhaps they used one of the already existing design elements, then you could just reuse it.

melloskitten: In general, double-checking whether there are labels that could be ENALabels would be grand. 👍

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayViewModel.swift
```diff
@@ -0,0 +1,131 @@
...
+
+	// MARK: - Internal
+
+	var isSelected: Bool {
+		didSet {
+			update()
+		}
+	}
+
+	let fontSize: CGFloat = 16
```
melloskitten, line 54: I'm a little confused, wouldn't something like the font-size be rather part of the view than the view model?

melloskitten (changes requested):
Adding some minor comments. Great contribution!

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/__tests__/DatePickerOptionViewModelTests.swift
nickguendling: We only change the region, not the language, that's why it's still the German localization with the American order.

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayViewModel.swift
marcussc: Yeah I can see that, but where do you draw the line? The color, text and basically everything that styles the label comes from the ViewModel why not the fontSize as well? :)

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
nickguendling: I checked the ENALabel styles before, but the date picker uses different font size and weight combinations than we usually do so I couldn't use any of the existing styles. That's why I directly use the DynamicTypeLabel here. But I'm aware that we should use them wherever possible. :)

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayViewModel.swift
melloskitten: If I understand correctly the rest of the attributes are dynamic, while this one is static - I agree that drawing a line there is difficult, but I wanted to double check whether you have had thoughts about doing it like this :-) Which you did, as I see.

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
melloskitten: Is there the option to include this style into the ENALabel configuration, as an additional case? Or is this particular size / weight / color combination never going to be used again? Just asking, if it is a small change, including it into ENALabel seems to be more consistent :-) But I appreciate that you are aware of the original class, I just wanted to double check 💯

nickguendling: I guess that they probably won't be used again. It would be three new styles that are currently only used in this component. And as I see it, we currently partially mirror UIFont.TextStyle in our ENALabel, so these new styles would fall out of that concept. I would leave it with the DynamicTypeLabel for this special case if you're fine with it 😁

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/OptionGroup/DatePickerOption/DatePickerDayView.swift
melloskitten: That's alright then 💯

melloskitten (approved):

PR #1150, opened 2020-09-09
Extend the Client to support download keys from multi-countries.

ArturFriesen:
## Description
This PR extends the http client to support download of country packages.

Story: [2343](https://jira.itc.sap.com/browse/EXPOSUREAPP-2343)

## Checklist

* [ ] Make sure that your PR is not introducing _unnecessary_ reformatting (e.g., introduced by on-save hooks in your IDE)


src/xcode/ENA/ENA/Source/Client/Client.swift
```diff
@@ -38,15 +38,69 @@ protocol Client {
 	/// Gets the app configuration
 	func appConfiguration(completion: @escaping AppConfigurationCompletion)
 
+	#if INTEROP
+	/// Determines days that can be downloaded.
+	///
+	/// - Parameters:
+	///   - country: Country code
+	///   - completion: completion callback which includes the list of available days
+	func availableDays(
```
johannesrohwer, line 47: country code in which format/standard? i have seen that we use this also in other PRs, it could make sense to introduce an enum here? Just a thought, maybe for the future.

src/xcode/ENA/ENA/Source/Client/Client.swift
```diff
@@ -187,9 +230,37 @@ extension Client {
...
+		}
+
+		group.enter()
+		fetchHours(hours, day: day, country: country) { result in
+			hoursResult = result
+			group.leave()
+		}
+		group.notify(queue: .main) {
+			completeWith(FetchedDaysAndHours(hours: hoursResult, days: daysResult))
+		}
```
johannesrohwer, line 257: will fetchDays/fetchHours time out at some point or can we get stuck here?

src/xcode/ENA/ENA/Source/Client/Client.swift
```diff
@@ -243,4 +346,37 @@ extension Client {
...
+		let group = DispatchGroup()
+
+		hours.forEach { hour in
+			group.enter()
+			self.fetchHour(hour, day: day) { result in
+				switch result {
+				case let .success(hourBucket):
+					buckets[hour] = hourBucket
+				case let .failure(error):
+					errors.append(error)
```
johannesrohwer, line 366: the errors.append(...) is not thread safe, right?

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+DaysAndHoursTests.swift
```diff
@@ -37,6 +37,211 @@ final class HTTPClientDaysAndHoursTests: XCTestCase {
...
+		HTTPClient.makeWith(mock: stack).availableHours(day: "2020-05-12", country: "IT") { result in
+			switch result {
+			case let .success(hours):
+				XCTAssertEqual(
+					hours,
+					[1, 2, 3, 4, 5]
+				)
+				expectation.fulfill()
+			case let .failure(error):
+				XCTFail("a valid response should never yiled an error like \(error)")
```
johannesrohwer, line 143:
```suggestion
				XCTFail("a valid response should never yield an error like \(error)")
```

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMSubmissionStateViewController.swift
```diff
@@ -207,21 +207,42 @@ private extension Client {
...
+		}
+		
 		group.enter()
 		availableHours(day: .formattedToday()) { result in
 			if case let .success(hours) = result {
 				daysAndHours.hours = hours
 			}
 			group.leave()
 		}
+		#endif
```
johannesrohwer, line 245: too much low-level logic for a VC (but okay since this is just the developer menu)

johannesrohwer (changes requested):
Thank you for your contribution, I am very happy about the comments and all unit tests!! 💯  I just have a few minor comments.

src/xcode/ENA/ENA/Source/Client/Client.swift
whiskey: We could also create a `typealias` to the `id` field in `Country` and use this. Then it's clear and one knows where to get this identifier from.

ChristianKienle: Isn't the list of countries "dynamic"?

src/xcode/ENA/ENA/Source/Client/Client.swift
ChristianKienle: It will never get stuck. The dispatch group has no timeout and could get stuck but the URL session has one and the requests will time out and then the group will also "finish".

src/xcode/ENA/ENA/Source/Client/Client.swift
ChristianKienle: It is not but the completion handlers are all called on the Main Queue... so this should be fine.

ChristianKienle (commented):
From my point of view the PR is okay. Maybe you could address comments made by @johannesrohwer + the whitespace issues.

src/xcode/ENA/ENA/Source/Client/Client.swift
johannesrohwer: good point, so enums are probably not the way to go! How is this handled throughout the other parts of the app? Do we fetch flags and all other country information from the server or do we have a mix between dynamic and static content here?

ArturFriesen: We have a mix. We have static content with all countries. But the backend decides which of these countries are "unlocked".

ChristianKienle (approved):

johannesrohwer (approved):
Thank you for clearing up my questions everyone! Seems good to me 👍

PR #1156, opened 2020-09-11
Integrate onset of symptoms screens into the exposure submission flow

nickguendling:
## Description
This pull request integrates the onset of symptoms screens into the exposure submission flow. 
JIRA: https://jira.itc.sap.com/browse/EXPOSUREAPP-2481

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
@@ -220,7 +220,7 @@ extension ExposureSubmissionCoordinator {
 				switch selectedSymptomsOnsetOption {
 				case .exactDate(let date):
 					guard let daysSinceOnset = Calendar.gregorian().dateComponents([.day], from: date, to: Date()).day else {
-						fatalError()
+						fatalError("Getting days since onset from date failed")
 					}
 					self?.symptomsOnset = .daysSinceOnset(daysSinceOnset)
 				case .lastSevenDays:
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
@@ -18,7 +18,7 @@
 import UIKit
 import Combine
 
-class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController, ENANavigationControllerWithFooterChild {
+class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController, ENANavigationControllerWithFooterChild, RequiresDismissConfirmation {
 
 	typealias PrimaryButtonHandler = (SymptomsOnsetOption) -> Void
 
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
@@ -20,7 +20,9 @@ import Combine
 
 class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController, ENANavigationControllerWithFooterChild {
 
-	enum SymptomsOnsetSelectionOption {
+	typealias PrimaryButtonHandler = (SymptomsOnsetOption) -> Void
+
+	enum SymptomsOnsetOption {
 		case exactDate(Date)
 		case lastSevenDays
 		case oneToTwoWeeksAgo
@@ -32,7 +34,7 @@ class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController,
 
 	init?(
 		coder: NSCoder,
-		onPrimaryButtonTap: @escaping (SymptomsOnsetSelectionOption) -> Void
+		onPrimaryButtonTap: @escaping PrimaryButtonHandler
 	) {
 		self.onPrimaryButtonTap = onPrimaryButtonTap
 
@@ -55,7 +57,7 @@ class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController,
 	// MARK: - Protocol ENANavigationControllerWithFooterChild
 
 	func navigationController(_ navigationController: ENANavigationControllerWithFooter, didTapPrimaryButton button: UIButton) {
-		guard let selectedSymptomsOnsetSelectionOption = selectedSymptomsOnsetSelectionOption else {
+		guard let selectedSymptomsOnsetSelectionOption = selectedSymptomsOnsetOption else {
 			fatalError("Primary button must not be enabled before the user has selected an option")
 		}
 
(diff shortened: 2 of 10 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
```diff
@@ -18,9 +18,11 @@
 import UIKit
 import Combine
 
-class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController, ENANavigationControllerWithFooterChild {
+class ExposureSubmissionSymptomsOnsetViewController: DynamicTableViewController, ENANavigationControllerWithFooterChild, RequiresDismissConfirmation {
 
-	enum SymptomsOnsetSelectionOption {
+	typealias PrimaryButtonHandler = (SymptomsOnsetOption) -> Void
+
+	enum SymptomsOnsetOption {
```
melloskitten, line 25: What is the reason for you adding this information into the coordinator? Can you store this information somewhere else, preferably not the coordinator component?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
```diff
@@ -112,55 +112,178 @@ class ENAUITests_04_ExposureSubmissionUITests: XCTestCase {
...
-
 	}
 
-	func test_SubmitTAN() {
-		// Setup service mocks.
-		app.launchArguments += [UITestingParameters.ExposureSubmission.useMock.rawValue]
-		app.launchArguments += [UITestingParameters.ExposureSubmission.getRegistrationTokenSuccess.rawValue]
-		app.launchArguments += [UITestingParameters.ExposureSubmission.submitExposureSuccess.rawValue]
-		launch()
+	func test_SubmitTAN_SymptomsOptionNo() {
```
melloskitten, line 117: Thank you for adding UI tests that cover your new screens! 👍

melloskitten (changes requested):
Hi! Thanks a lot for your contribution. I have one minor comment in regards to adding additional attributes into the coordinator.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
nickguendling: This is in the view controller, I'm not sure what you mean here 😅

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
nickguendling: ☺️

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionSymptomsOnsetViewController.swift
melloskitten: That's weird. I could have bet I put this comment someplace else. Whatever, it is fine :-)

melloskitten (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -194,6 +195,51 @@ extension ExposureSubmissionCoordinator {
...
+				switch selectedSymptomsOption {
+				case .yes:
+					self?.showSymptomsOnsetScreen()
+				case .no:
+					self?.symptomsOnset = .nonSymptomatic
+					self?.showWarnOthersScreen()
+				case .preferNotToSay:
+					self?.symptomsOnset = .noInformation
+					self?.showWarnOthersScreen()
+				}
```
johannesrohwer, line 210: I guess @melloskitten's comment was aimed at these changes here in the coordinator (same applies for the showSymptomsOnsetScreen()). I would propose to handle this logic in the respective view controllers (if possible? Is there a specific reason why it is done here?). I would try to keep the coordinator free of any logic besides routing and dependency management.

johannesrohwer (changes requested):
Thank you for your contribution! 🚀  I really liked that you added UI tests and have just a small question, everything else looks good to me!

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -194,6 +195,51 @@ extension ExposureSubmissionCoordinator {
...
+
+		push(vc)
+	}
+
+	func showSymptomsOnsetScreen() {
+		let vc = createSymptomsOnsetViewController(
+			onPrimaryButtonTap: { [weak self] selectedSymptomsOnsetOption in
+				switch selectedSymptomsOnsetOption {
+				case .exactDate(let date):
+					guard let daysSinceOnset = Calendar.gregorian().dateComponents([.day], from: date, to: Date()).day else {
```
whiskey, line 222: Please use `Calendar.autoupdatingCurrent`. I know, we aren't covering others in the UI but for the delta calculation of days, `Calendar.autoupdatingCurrent` is valid. No need to specify a certain calendar here.

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: I am currently working on a refactoring of this part outside of this story. I think the switch is actually mainly routing and belongs here. I will move the symptomsOnset property as well as the exposure submission service to a separate coordinator model to make it more testable. The previous implementation seemed more like a factory to me than a coordinator. I don't think the view controller itself should have any clue which view controller precedes it or comes after it, that's why in this first step I pulled that information out of there. Would be great if we can leave it here for this story and discuss it in my pull request that will follow in the next days :)

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: Is the date difference in days the same in every calendar that is supported on iOS? I actually don't know this 🤷‍♂️ Since I didn't want to check and update that information once additional calendars are supported in iOS and we only support the Gregorian calendar here, I wanted to make sure the calculation is correct based on what the user selected in the interface. I would prefer to leave it as is and not take any risks here.

nickguendling: I just thought of an example where it could go wrong: If the `Calendar.autoupdatingCurrent` has a leap second in the range that is calculated, while the Gregorian calendar has not, the calculation would be one day off.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -194,6 +195,51 @@ extension ExposureSubmissionCoordinator {
...
+						fatalError()
```
ArturFriesen, line 223: An error message would be nice ☀️

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
johannesrohwer: thank you for clarifying! 👍  I understand your reasoning and I am looking forward to see your future PR that addresses your points.

johannesrohwer (approved):

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: Added one 😊

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: I'm still uncertain that we have to use gregorian here, but because I don't know for sure, I'm not insisting on this atm :)

PR #1158, opened 2020-09-11
Handle 'corrupted' database key

whiskey:
## Description

Under certain circumstances a database key stored in the keychain might mismatch with the expected one for the database. This might happen during backup restoration and other cases. The result is a loop on app launch that never leaves the onboarding, because the 'onboarding seen' flag is set to a non-existing database.

The current approach is twofold:

* for developers, the code accessing the secure storage is refactored to `throw` proper errors instead of just logging and omitting them
* for the app flow, if the initial opening/creation of the database fails, the whole store is reset (database & key) and retired. If that also fails, the applications throws a fatal error. 

This PR tackles JIRA issue [EXPOSUTEAPP-2323](https://jira.itc.sap.com/browse/EXPOSUREAPP-2323).

## Checklist


marcussc:
Please add Label and Milestone :)

src/xcode/ENA/ENA/Source/Workers/Keychain/KeychainHelper.swift
```diff
@@ -19,51 +19,70 @@
...
+				throw KeychainError.save(reason: message)
 			} else {
-				logError(message: "Failed to add keychain item '\(key)' due to unknown error")
+				let message = "Failed to add keychain item '\(key)' due to unknown error"
+				throw KeychainError.save(reason: message)
 			}
 		}
 
-		return addResult == errSecSuccess
+		return true
```
marcussc, line 79: If the functions `throws` and always returns true, why not remove the return type?

whiskey: I had the same thoughts but just wanted to finish and don't start a refactoring spree removing all these oldschool C-style patterns. There are tons of such issues, especially with the SQL handling. This is definitely something for my own refactoring list.

marcussc: So you didn't refactor it because it would be to big of an undertaking? Would be huge if you only refactored this occurrence?

whiskey: It seems I removed all reference to the old return value. This is refactored out and should be reflected here soon.

marcussc (approved):

src/xcode/ENA/ENA/Source/Workers/Keychain/__tests__/KeychainHelperTests.swift
```diff
@@ -0,0 +1,75 @@
...
+			XCTFail("Expected data for the given key")
+			return
+		}
+		XCTAssertEqual(String(data: retrieved, encoding: .utf8), "foo")
+	}
+
+	func testFailingKeyRetrieval() throws {
+		let keychain = try KeychainHelper(service: serviceID)
+		let retrieved = keychain.loadFromKeychain(key: testKey)
+		XCTAssertNil(retrieved)
```
ArturFriesen, line 66: 👍 for testing unhappy paths!

ArturFriesen (approved):

PR #1169, opened 2020-09-14
Bugfix: smallest text change ever

steins-code:
https://jira.itc.sap.com/browse/EXPOSUREAPP-2028

What has changed?
One bracket's position :)


janetback (approved):
UA approved

ArturFriesen (approved):

PR #1171, opened 2020-09-15
Remove outdated RKI number from Terms of Use

melloskitten:
## Description
This PR removes the outdated RKI number from the Terms of Use, as described in issue [2633](https://jira.itc.sap.com/browse/EXPOSUREAPP-2633). Note that it was only removed from the german localization, since the translators will take care of the other versions.


janetback (changes requested):
Hi Sandra, I don't think we should mention 'telefonisch' any more, if they have removed the hotline number.
I think you should change this sentence to '...können Sie das RKI per E-Mail unter....' .
Let me know if you need any more info. Thanks, Janet

janetback (approved):
UA approved

ArturFriesen (approved):

PR #1184, opened 2020-09-16
Fix: Make symptom options selectable on iOS 14

nickguendling:
## Description
Makes the symptom options selectable by adding the option group view to the content view. Otherwise the content view is put on top of them on iOS 14 and the tap gesture is not recognized by the underlying views.

Bug: https://jira.itc.sap.com/browse/EXPOSUREAPP-2698


ArturFriesen (approved):

marcussc:
Added on hold until we have a jira issue

marcussc (approved):

thomasaugsten:
https://jira.itc.sap.com/browse/EXPOSUREAPP-2698

PR #1194, opened 2020-09-18
EXPOSUREAPP-2547: Countdown stopped working when resuming app from ba…

SwissMountainFrog:
## Description
This fixes an issue that the countdown for manual risk update determination (automatic updates in the background are disabled) got stuck after the iOS app resumes (becomes the active app).

The issue was reported in https://github.com/corona-warn-app/cwa-app-ios/issues/1124 and https://jira.itc.sap.com/browse/EXPOSUREAPP-2547.


thomasaugsten:
fix for #1124

src/xcode/ENA/ENA/Source/Scenes/Home/HomeViewController.swift
```diff
@@ -100,6 +103,11 @@ final class HomeViewController: UIViewController, RequiresAppDependencies {
 		homeInteractor.requestRisk(userInitiated: false)
 		updateBackgroundColor()
 	}
+	
+	override func viewWillDisappear(_ animated: Bool) {
+		let notificationCenter = NotificationCenter.default
+        notificationCenter.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
```
ArturFriesen, line 109: Maybe this is intended, but I think adding the observer in viewDidLoad and removing it in viewWillDisappear could lead to a ViewController which does not react to the notification.

If you remove it in viewWillDisappear you should add it in viewWillAppear.
Or keep it in viewDidLoad and remove it in deinit.

Calls can be:
viewDidLoad -> viewWillAppear -> viewWillDisappear -> viewWillAppear and so on.

SwissMountainFrog: Hi Artur, thanks for that note. To be consistent, I will move the code from `viewWillDisappera` to `deinit`.

haosap: some remark:
If you're using `selector` to do the callback, you don't need to do the unregister. From Apple:

**If your app targets iOS 9.0 and later or macOS 10.11 and later, you do not need to unregister an observer that you created with this function**

SwissMountainFrog: Thanks @haosap for that. I removed `deinit` now completely

melloskitten (approved):

johannesrohwer (approved):

ArturFriesen (approved):

PR #1205, opened 2020-09-21
Country key download and exposure detection  (Interop – 2667, 2764)

ArturFriesen:
## Description

This PR introduces the download of key packages based on the country approach. All country key packages are combined for exposure detection.

Story: [2764](https://jira.itc.sap.com/browse/EXPOSUREAPP-2764)

## Checklist


src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -213,6 +196,23 @@ final class HTTPClient: Client {
...
+		appConfiguration { config in
+			guard let config = config else {
+				completion(nil)
+				return
+			}
+			guard config.hasExposureConfig else {
+				completion(nil)
+				return
+			}
+			completion(try? ENExposureConfiguration(from: config.exposureConfig, minRiskScore: config.minRiskScore))
```
johannesrohwer, line 212: the docs in the protocol say the completion handler will always be called on the main thread. is this guaranteed here?

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AppConfigTests.swift
```diff
@@ -97,6 +97,9 @@ final class HTTPClientAppConfigTests: XCTestCase {
 			case .failure(let error):
 				XCTFail("Country fetch error: \(error as NSError)")
 			case .success(let list):
+				
+				// As of 2020-09-16 the latest backend deployment returns zero (0) countries.
+				// Due to the current changes in INTEROP we should check this by end of Sept 2020!
```
johannesrohwer, line 102: should we add a ticket so we do not forget about this?

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V1/DownloadedPackagesStoreV1.swift
```diff
@@ -21,12 +21,12 @@ import FMDB
...
-	func package(for day: String, country: String) -> SAPDownloadedPackage?
-	func hourlyPackages(for day: String, country: String) -> [SAPDownloadedPackage]
-	func allDays(country: String) -> [String] // 2020-05-30
-	func hours(for day: String, country: String) -> [Int]
+	func set(country: Country.ID, day: String, package: SAPDownloadedPackage)
+	func set(country: Country.ID, hour: Int, day: String, package: SAPDownloadedPackage)
+	func package(for day: String, country: Country.ID) -> SAPDownloadedPackage?
+	func hourlyPackages(for day: String, country: Country.ID) -> [SAPDownloadedPackage]
+	func allDays(country: Country.ID) -> [String] // 2020-05-30
+	func hours(for day: String, country: Country.ID) -> [Int]
```
johannesrohwer, line 29: Country.ID >>> String 🎉

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/CountryKeypackageDownloader.swift
```diff
@@ -0,0 +1,65 @@
...
+			return
+		}
+
+		guard let deltaDaysAndHours = delegate?.exposureDetection(country: country, downloadDeltaFor: daysAndHours) else {
+			completion(.failure(.noDaysAndHours))
+			return
+		}
+
+		delegate?.exposureDetection(country: country, downloadAndStore: deltaDaysAndHours) { error in
+			if error != nil {
```
johannesrohwer, line 57: nitpick: guard

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
```diff
@@ -25,53 +25,104 @@ final class ExposureDetection {
...
-		guard let writtenPackages = delegate?.exposureDetectionWriteDownloadedPackages(self) else {
-			endPrematurely(reason: .unableToWriteDiagnosisKeys)
-			return
+
+		dispatchGroup.notify(queue: .main) {
+			if let error = errors.first {
+				self.endPrematurely(reason: error)
+			} else {
+				completion()
+			}
```
johannesrohwer, line 91: why do we collect all errors and only return the first one?

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/__tests__/ExposureDetectionTests.swift
```diff
@@ -75,18 +89,67 @@ final class ExposureDetectionTransactionTests: XCTestCase {
...
+		let expectationDetectionCompletion = expectation(description: "Detection completion was called.")
+		detection.start { _ in
+			expectationDetectionCompletion.fulfill()
+		}
+
+		waitForExpectations(timeout: 1.0)
+	}
+
+	func makeCountries() -> [Country] {
+		guard let enCountry = Country(countryCode: "FR"),
```
johannesrohwer, line 145: nitpick: frCountry? 
could also be shortened to something like:
`[Country(countryCode: "FR"), Country(countryCode: "IT")].compactMap { $0 }`
(which does not XCFail() when the country init fails but this should be covered by other tests..)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -254,7 +254,7 @@ extension RiskProvider: RiskProviding {
 			group.leave()
 		}
 
-		guard group.wait(timeout: .now() + .seconds(60)) == .success else {
+		guard group.wait(timeout: .now() + .seconds(60 * 8)) == .success else {
```
johannesrohwer, line 257: is any reasoning why 8 min is the perfect amount of time to wait?

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
```diff
@@ -25,53 +25,104 @@ final class ExposureDetection {
...
 			}
-			self.delegate?.exposureDetection(self, downloadConfiguration: self.useConfiguration)
-		}
+		})
+	}
+
+	private func getCountriesToDetect(supportedCountries: [Country]) -> [Country.ID] {
+		var countryIDs = supportedCountries.map { $0.id }
+		countryIDs.append(Country.defaultCountry().id)
+		return countryIDs
```
johannesrohwer, line 64: would it make sense to eliminate possible duplicates here? its unlikely but if we get a dup here (e.g. the default country) we might run into unexpected trouble somewhere else.

johannesrohwer (changes requested):
awesome PR 🥇 just a few minor points from my side!

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AppConfigTests.swift
ArturFriesen: Added subtask.https://jira.itc.sap.com/browse/EXPOSUREAPP-2783

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V1/DownloadedPackagesStoreV1.swift
ArturFriesen: 🥳

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/CountryKeypackageDownloader.swift
ArturFriesen: added guard

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
ArturFriesen: Good idea. Used a Set to do this.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
ArturFriesen: Its a dispatch group and the UI can only handle one error. If one error occurs, risk calculation will not be triggered.

ArturFriesen: I have added 2 tests for this unhappy path.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/__tests__/ExposureDetectionTests.swift
ArturFriesen: changed

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: Its a requirement from the story.

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
ArturFriesen: This comment is wrong. I removed it.
This function is only used by the ExposureDetection.
The ExposureDetection "forwards" this call to the UI on the main thread.

melloskitten (approved):
Thanks for your contribution. Looks good to me :-)

johannesrohwer (approved):
Thank you for implementing the feedback! Awesome contribution 🎉

PR #1211, opened 2020-09-22
Fix behaviour on key upload

marcussc:
## Description
Fixes: https://jira.itc.sap.com/browse/EXPOSUREAPP-2293

If the User declines sharing his TEKs with the App during Key Upload we want to stay on the same screen instead of showing an error message.

DontShareTEKs


nickguendling (approved):

whiskey (approved):
Ein Test wäre schön gewesen ;)

ChristianKienle (approved):

PR #1224, opened 2020-09-23
Removed RKI Phone Number

tcfos:

## Description
terms of use: removed phone number of RKI
text formatting of data privacy notes
Jira #2633


janetback (approved):
UA reviewed

marcussc (approved):

PR #1248, opened 2020-09-27
Caching App-Config (2790 - 2858)

whiskey:
## Description
As requested ([private link](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/e12a4a9c09ae0dcb28f4b45a4c669d3eb587b8b4/docs/spec/handling-of-app-config.md)), the app should cache the app configuration according to the given cache control headers. To allow caching, a new new `URLSessionConfiguration`
(**cachingSessionConfiguration**) was created and is used for the fetch.

The configuration itself is stored in the app's **SecureStore** like all other data. Accessing the configuration is specified in `AppConfigurationProviding` and realized via `CachedAppConfiguration`. Direct calls to the HTTPClient are marked as deprecated and will soon be removed/refactored.

The caching mechanism is centered around three protocols:

1. `AppConfigurationFetching` – "API requests / Client"
1. `AppConfigCaching` – "Local Cache / SecureStore"
1. `AppConfigurationProviding` – "Model / AppConfig provider"

## Fixed Tickets
* [EXPOSUREAPP-2790](https://jira.itc.sap.com/browse/EXPOSUREAPP-2790)
* [EXPOSUREAPP-2858](https://jira.itc.sap.com/browse/EXPOSUREAPP-2858)

## Checklist


src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -0,0 +1,91 @@
...
+					let error = CachedAppConfiguration.CacheError.dataVerificationError(message: "Failed to verify app config signature")
+					completion(.failure(error))
+					return
+				}
+
+				// serialize config
+				do {
+					let config = try SAP_ApplicationConfiguration(serializedData: package.bin)
+					let eTag = response.httpResponse.value(forHTTPHeaderField: "ETag")
+					let c = AppConfigurationFetchingResponse(config, eTag)
```
marcussc, line 81: `c` seems a little short :D

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinatorModel.swift
```diff
@@ -110,21 +110,25 @@ class ExposureSubmissionCoordinatorModel {
 		onError: @escaping (ExposureSubmissionError) -> Void
 	) {
 		isLoading(true)
-		appConfigurationProvider.appConfiguration { applicationConfiguration in
+		appConfigurationProvider.appConfiguration { result in
```
marcussc, line 113: <3

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
```diff
@@ -178,49 +202,48 @@ class ExposureSubmissionCoordinatorModelTests: XCTestCase {
...
-
-			waitForExpectations(timeout: .short)
-			XCTAssertEqual(isLoadingValues, expectedIsLoadingValues)
-
-			XCTAssertFalse(model.shouldShowSymptomsOnsetScreen)
-			XCTAssertEqual(model.supportedCountries, [])
-		}
-	}
+//	func testSymptomsOptionNoOrPreferNotToSaySelectedSupportedCountriesLoadFails() {
+//		let symptomOptions: [ExposureSubmissionSymptomsViewController.SymptomsOption] = [.no, .preferNotToSay]
```
marcussc, line 206: Hmm is this still needed or can it be removed?

whiskey:
## Caveat

Currently the app has some old, 'direct' calls to the HTTP client which are marked as deprecated and will be removed soon – as mentioned above. If you are using proxies like `mitmproxy` you'll see multiple calls to the configuration endpoint as if no caching is in place or it's broken. This will be fixed with the follow-up PR that removes the old calls.

I hope I covered most common cases in my unit tests to simulate the expected/intended behaviour.

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -52,4 +52,9 @@ final class MockTestStore: Store {
 	var allowTestsStatusNotification: Bool = true
 	var hourlyFetchingEnabled: Bool = true
 	var userNeedsToBeInformedAboutHowRiskDetectionWorks = false
+
+	// MARK: - AppConfigCaching
+	
+	var lastETag: String? = nil
```
marcussc, line 58:
```suggestion
	var lastAppConfigETag: String? = nil
```

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
whiskey: but it's scope is two adjacent lines!? 🤷‍♂️

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -0,0 +1,91 @@
...
+					completion(.success(c))
```
whiskey, line 82:
```suggestion
					let configurationResponse = AppConfigurationFetchingResponse(config, eTag)
					completion(.success(configurationResponse))
```

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
whiskey: I forgot to check with @nickguendling on that! I don't understand the logic that denying this should affect the local cache.

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
marcussc: Thanks for changing 👍

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
whiskey: fixed & reenabled

src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -93,7 +93,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 	let taskScheduler: ENATaskScheduler = ENATaskScheduler.shared
 
 	lazy var appConfigurationProvider: AppConfigurationProviding = {
-		CachedAppConfiguration(client: self.client)
+		// use a custom http client that uses/recognized caching mechanisms
+		let appFetchingClient = CachingHTTPClient(basedOn: client)
```
ArturFriesen, line 97: To minimize the dependency would it be possible to inject the Configuration?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -20,48 +20,87 @@
...
+	private let store: AppConfigCaching
+
+	init(client: AppConfigurationFetching, store: AppConfigCaching) {
 		self.client = client
+		self.store = store
+
+		guard shouldFetch() else { return }
+
+		// edge case: if no app config is cached, omit a potentially existing ETag to force fetch a new configuration
+		let etag = store.appConfig == nil ? nil : store.lastAppConfigETag
```
ArturFriesen, line 44: Can lastAppConfigETag exist without  a config?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
```diff
@@ -17,20 +17,35 @@
...
+@testable import ENA
 
-// swiftlint:disable:next type_body_length
 class ExposureSubmissionCoordinatorModelTests: XCTestCase {
 
+	// default provider for a static app configuration
+	let configProvider = CachedAppConfiguration(client: CachingHTTPClientMock(), store: MockTestStore())
+
+	override func setUpWithError() throws {
+		// uses a common database!
```
ArturFriesen, line 29: That does this comment mean? Is it a todo?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
```diff
@@ -17,20 +17,35 @@
...
 import XCTest
+@testable import ENA
 
-// swiftlint:disable:next type_body_length
 class ExposureSubmissionCoordinatorModelTests: XCTestCase {
 
+	// default provider for a static app configuration
+	let configProvider = CachedAppConfiguration(client: CachingHTTPClientMock(), store: MockTestStore())
+
+	override func setUpWithError() throws {
```
ArturFriesen, line 28: Nothing throws here. You can just use setUp()

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
```diff
@@ -17,20 +17,35 @@
...
+		// uses a common database!
+		let store = MockTestStore()
+		store.appConfig = nil
+		store.lastAppConfigETag = nil
+	}
+
+	override func tearDownWithError() throws {
```
ArturFriesen, line 35: Nothing throws here. You can just use setUp()

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
```diff
@@ -28,4 +31,15 @@ extension URLSessionConfiguration {
 		config.requestCachePolicy = .reloadIgnoringLocalAndRemoteCacheData // avoid stale data
 		return config
 	}
+
+
+	/// Session config with caching policy `.useProtocolCachePolicy`
+	///
+	/// - Note: Caching is not realized via `NSURLCache` but via the internal SecureStore
```
ArturFriesen, line 38: I'm neugierig why couldn't we use the default caching? Could you extend the comment here?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -20,48 +20,87 @@
...
+		self.appConfiguration(forceFetch: false, completion: completion)
 	}
-}
 
-private extension CachedAppConfiguration {
-	struct Cache {
-		let date: Date
-		let value: SAP_ApplicationConfiguration
+	/// Simple helper to simulate Cache-Control
+	/// - Note: This 300s value is because of current handicaps with the HTTPClient
```
ArturFriesen, line 98: Could you please explain, if possible, which handicaps?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: We currently have a very naïve store which does not provide any validation. We can in fact set/store ETags without app configuration. Future refactorings of the Store could clear the ETag if a config is deleted but that should be refactored later and not as part of this PR.

ArturFriesen: Maybe a solution is lastAppConfigETag and appConfig should be combined in a struct?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
whiskey: Just a very simple hint to the common database file. There is no need to have a property or even singleton for `MockTestStore`. Normally you would assume that changes to that store instance would only affect it and not also the stores in the tests. 

This could be elaborated in more detail but I just didn't do it.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: Bad architecture. I would have to refactor even more within the HTTPClient to pass HTTP responses including header fields.

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
whiskey: Everybody was scared of potential data privacy issues so I decided to move it to our home grown store 🤷‍♂️

src/xcode/ENA/ENA/Source/AppDelegate.swift
whiskey: It's not a hard dependency (on `client`). The caching client could also be initialized like `CachingHTTPClient()` and works the same. I used the existing 'scheme' to keep the changes simple. If you have something specific in mind, please suggest :)

(The AppDelegate definitely needs some love but that should be a dedicated refactoring/cleanup)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: In the database etag, config and further cache control is stored in separate fields. We need a setter for the app config that checks and verifies the pre- and postconditions on changes. I think, a simple struct doesn't help.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
whiskey: I changed the comment. It's updated here soon.

src/xcode/ENA/ENA/Source/AppDelegate.swift
ArturFriesen: The CachingHTTPClient only uses the HTTPClient.Configuration.
So I would suggest to only inject HTTPClient.Configuration.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
ArturFriesen: Ok, good point.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
ArturFriesen: Could you just mention the problem in the comment? And then its fine for me.

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
ArturFriesen: Ok, I understand. Suggestion for the comment: "Because of privacy reasons, caching is not realized via `NSURLCache` but via the internal encrypted SecureStore"

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
whiskey: Ok, done.

src/xcode/ENA/ENA/Source/AppDelegate.swift
whiskey: Ah, thanks for clarification. 

My initial idea was to use the original session as template as well (plus caching) but this changed. Will change this in a moment.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: Ok?

> /// Simple helper to simulate Cache-Control
	/// - Note: This 300 second value is because of current handicaps with the HTTPClient architecture
	///   which does not easily return response headers. This requires further refactoring of `URLSession+Convenience.swift`.
	/// - Returns: `true` is a network call should be done; `false` if cache should be used

ArturFriesen: Thanks!

marcussc (approved):

PR #1252, opened 2020-09-28
Submit days since onset of symptoms (2432)

nickguendling:
Submit the days since onset of symptoms in addition to the transmission risk level.
JIRA: https://jira.itc.sap.com/browse/EXPOSUREAPP-2432
Tech Spec: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/master/docs/spec/days-since-onset-of-symptoms.md


src/xcode/ENA/ENA/Source/Extensions/ENTemporaryExposureKey+Processing.swift
```diff
@@ -41,21 +41,24 @@ extension Array where Element: ENTemporaryExposureKey {
...
-				for exposureKey in exposureKeys {
-					/// Assign corresponding transmission risk level
-					exposureKey.transmissionRiskLevel = symptomsOnset.transmissionRiskVector[daysUntilToday]
-				}
-			} else {
-				/// Remove keys that have no corresponding transmission risk level
-				groupedExposureKeys[daysUntilToday] = nil
+		/// 2. Assign the corresponding transmission risk levels and days since onset of symptoms. Keys that have no corresponding transmission risk level are filtered out.
+		var processedExposureKeys = [SAP_TemporaryExposureKey]()
+		for (daysUntilToday, exposureKeys) in groupedExposureKeys where daysUntilToday >= 0 && daysUntilToday <= 14 {
```
mlenkeit, line 46: It might be worth considering renaming `daysUntilToday` to `ageInDays`. For me, it makes it a bit clearer that I cannot really be negative.

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
```diff
@@ -23,317 +23,422 @@ import ExposureNotification
...
+		
 		let days = 0...14
-		var keys = days.map { makeMockKey(daysUntilToday: $0) }
-		keys = keys.processedForSubmission(with: symptomsOnset, today: Date(timeIntervalSinceReferenceDate: 0))
-
-		keys.sort { $0.rollingStartNumber > $1.rollingStartNumber }
-
-		for (key, day) in zip(keys, days) {
+		let keys = days.map { makeMockKey(daysUntilToday: $0) }
+		var processedKeys = keys.processedForSubmission(with: symptomsOnset, today: Date(timeIntervalSinceReferenceDate: 0))
```
mlenkeit, line 34: Might be worth considering moving this section (until end of the for-loop) into a local helper-assertion to avoid code duplication in the code.

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
```diff
@@ -23,317 +23,422 @@ import ExposureNotification
...
+		
+		processedKeys.sort { $0.rollingStartIntervalNumber > $1.rollingStartIntervalNumber }
+		
+		for (key, day) in zip(processedKeys, days) {
```
mlenkeit, line 38: The risk I see with this test approach is that when you change something in SymptomsOnset, the tests will pass. This defies the purpose of having to change both prod and test code to make a new feature pass. 

While checking each value explicitly for all scenario might be too much, I would suggest to add spot checks on TRL and DSOS for maybe the first, last, and some element in the array. The expected value should be hard-coded in the test code.
I would prefer if we had specific spot checks in addition to just looping over this.

mlenkeit (changes requested):
2 proposals and one specific ask wrt the tests.

src/xcode/ENA/ENA/Source/Extensions/ENTemporaryExposureKey+Processing.swift
nickguendling: I like that name suggestion, changed it 👍

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
nickguendling: I moved all the asserts to a helper function, makes the tests much more compact and readable, thanks for the suggestion!

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
nickguendling: I agree that a change of values in SymptomsOnset should trigger a test to fail. We have separate tests that cover every single value in SymptomsOnsetTests.swift, that's why I decided not to check them explicitly again in the tests where the values are just applied. Do you agree with me, that the combination of tests are enough to cover double-entry bookkeeping or do you think those tests are not sufficient?

mlenkeit: Good counter-argument 👍
That makes me think whether it makes sense that all test cases of SymptomsOnsetTests should be moved instead and treat the fact that SymptomsOnset exists as a "design decision" that in itself does not have to be tested.
I'm a big supported of automated unit tests but I don't see it so religiously that every class needs to have its own test class. If a new class exists only because of a certain design decision and it's basically only used in one context (i.e. `processedForSubmission`), I think that's fine.
Anyway, massive duplication of the test logic here and in SymptomsOnsetTests should be avoided for sure. Benefit of having everything here would be that you could see at a glance the values for TRL, for DSOS, etc.
What do you think?

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
```diff
@@ -23,317 +23,202 @@ import ExposureNotification
...
-				 key.transmissionRiskLevel,
-				 transmissionRiskVector[day],
-				 "Transmission Risk Level vector not applied correctly! Expected \(transmissionRiskVector[day]), got \(key.transmissionRiskLevel)"
-			 )
-		}
+		assertCorrectProcessing(on: processedKeys, for: days, with: symptomsOnset)
 	}
-
+	
 	func testSubmissionPreprocess_KeysOlderThan14DaysAreeIgnored() {
```
mlenkeit, line 166: Typo: `Aree` -> `Are`

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
nickguendling: I would like to keep them independently. I don't see a huge benefit to bring these two sets of tests together if we can cleanly separate them and already have 😅

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
nickguendling: Fixed 🤓👌

src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey+ProcessingTests.swift
mlenkeit: Fair enough 👍

mlenkeit (approved):

SwissMountainFrog (approved):

marcussc (approved):

ChristianKienle (approved):

PR #1253, opened 2020-09-28
Fix repeated onboarding elements (2862)

steins-code:
Resolve duplicated elements during onboarding as described here: https://jira.itc.sap.com/browse/EXPOSUREAPP-2862

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
@@ -84,7 +84,7 @@ final class OnboardingInfoViewController: UIViewController {
 
 	private var onboardingInfos = OnboardingInfo.testData()
 	private var exposureManagerActivated = false
-	private var pageSetupDone = [OnboardingPageType: Bool]()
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -84,13 +84,15 @@ final class OnboardingInfoViewController: UIViewController {
 
 	private var onboardingInfos = OnboardingInfo.testData()
 	private var exposureManagerActivated = false
+	private var pageSetupDone = [OnboardingPageType: Bool]()
```
ArturFriesen, line 87: Wouldn't it be enough to have just a bool? The OnboardingInfoViewController is initialized only with one pageType. And the status "done" is independent of the type.

steins-code: done

marcussc (approved):

marcussc:
Please don't forget to set the fix version in Jira, thanks 👍

melloskitten (approved):
LGTM 💯

PR #1261, opened 2020-09-29
Exclude files w/ legal texts

tcfos:
exclude files matching the pattern \*.legal.\* from the translation process

Legal texts are provided for each language (EN/DE/TR) and must not be part of the translation process.


tcfos:
duplicate. Closed.

PR #1285, opened 2020-10-02
Feature/3017 improve dead man notification

### LINKED ISSUE #1144: Dead man notification does not work
## Avoid duplicates

* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is not already reported in another issue

## Describe the bug

As [reported previously](https://github.com/corona-warn-app/cwa-app-ios/issues/916) background checks are ~not working on my iPhone~ only work when I connect my phone to a charger for unknown reasons.
So I was patently waiting for the new [dead man notification](https://github.com/corona-warn-app/cwa-app-ios/blob/v1.3.0/src/xcode/ENA/ENA/Source/Models/Task%20Scheduling/ENATaskScheduler.swift#L115) that was added with 1.3.0. Unfortunately I received no notification within 43 hours (at which point I opened the app to force a check).

image

## Expected behaviour

After 36 hours of no check a local notification should prompt me to open the app.

Wait for 36 hours after the last check without opening the app.

## Technical details

- iOS Version: 13.7
- Device: iPhone 8 Plus

## Possible Fix
I have no idea really but I'm wondering if the local notification is only scheduled in an actual background task, not when the check is executed while the app is in the foreground. Because in my case the checks are only happening when I'm opening the app.

---
Internal Tracking ID: EXPOSUREAPP-2994

marcussc:
## Description
Before this PR the Deadman Notification gets pushed 36 hours into the future once the Background-Task was successfully started. But the outcome of the Background-Task was not taken into account.

This can be improved, the Deadman Notification will now only be pushed 36 hours into the future if the Riskcalculation was successfull, otherwise the 36 hours will still keep ticking down. 

Solves: https://jira.itc.sap.com/browse/EXPOSUREAPP-3017


src/xcode/ENA/ENA/Source/AppDelegate+ENATaskExecutionDelegate.swift
```diff
@@ -101,6 +101,9 @@ extension AppDelegate: ENATaskExecutionDelegate {
 				body: AppStrings.LocalNotifications.detectExposureBody,
 				identifier: ENATaskIdentifier.exposureNotification.backgroundTaskSchedulerIdentifier + ".risk-detection"
 			)
+			
+			// We were successful to calculate a risk in the Background, time to reset Deadman Notification
```
SwissMountainFrog, line 105: Code comment 🚀

ArturFriesen (approved):
Nice! 🌟🌟🌟

SwissMountainFrog (approved):

PR #1304, opened 2020-10-06
Allow submission without symptoms #3086

nickguendling:
## Description
Adds the option to skip the symptoms screens in the submission flow.
JIRA: https://jira.itc.sap.com/browse/EXPOSUREAPP-3086

Simulator Screen Shot - iPhone 11 Pro - 2020-10-06 at 14 08 09


marcussc (approved):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -50,9 +50,17 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 		self.window = window
 
 		#if UITESTING
-		if let isOnboarded = UserDefaults.standard.value(forKey: "isOnboarded") as? String {
+		if let isOnboarded = UserDefaults.standard.string(forKey: "isOnboarded") {
 			store.isOnboarded = (isOnboarded != "NO")
 		}
+
+		if let onboardingVersion = UserDefaults.standard.string(forKey: "onboardingVersion") {
```
ArturFriesen, line 57: Is this ever true?

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
nickguendling: Not yet. This was our first approach and we think it will be useful once we write a UI test for the delta onboarding screen, so we left it in 🤓

PR #1315, opened 2020-10-07
Feature/2900 improved tests

whiskey:
## Description

This PR combines Unit- & UI-Tests by using [test plans](https://useyourloaf.com/blog/xcode-test-plans/).

Also the `UITESTING` precompiler flag has been removed. For more information see [COMPILATION_CONDITIONS.md](https://github.com/corona-warn-app/cwa-app-ios/compare/feature/2991_merge-test-schemas?expand=1#diff-5892de266681065cfa630e404d94c6ca).

Fixed JIRA issues: 
* [EXPOSUREAPP-2900](https://jira.itc.sap.com/browse/EXPOSUREAPP-2900) (+ sub tasks)

## Checklist


src/xcode/ENA/ENA.xcodeproj/xcshareddata/xcschemes/ENAScreenshots.xcscheme
```diff
@@ -32,23 +32,32 @@
...
       language = "de"
       region = "DE"
       codeCoverageEnabled = "YES"
       onlyGenerateCoverageForSpecifiedTargets = "YES">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Run Script"
+               scriptText = "# Ensure the hardware keyboard is DISCONNECTED from the simulators to prevent issues where the software keyboard is not shown.&#10;# Solution stolen from: https://stackoverflow.com/a/57269479/194585&#10;&#10;killall Simulator&#10;&#10;defaults write com.apple.iphonesimulator ConnectHardwareKeyboard -bool false&#10;&#10;">
```
marcussc, line 57: <3

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -122,11 +122,10 @@ extension ExposureSubmissionCoordinator {
 	/// and returns an ExposureSubmissionTestResultViewController.
 	/// Option 3: (default) return the ExposureSubmissionIntroViewController.
 	private func getInitialViewController(with result: TestResult? = nil) -> UIViewController {
-		#if UITESTING
-		if ProcessInfo.processInfo.arguments.contains("-negativeResult") {
+		#if DEVELOP
```
marcussc, line 125: Can you list `DEVELOP` in COMPILATION_CONDITIONS.md

marcussc (commented):
Not sure in which file it would appear; but why is the execution Order Alphabetical and not Random?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: Gnaa… nope! Because this should be `DEBUG` and I'm always confusing this shit. 
Will update this in a moment!

whiskey:
> Not sure in which file it would appear; but why is the execution Order Alphabetical and not Random?

Good catch. It should be in the test plans, which is not the case. Will be added in a moment.

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionServiceFactory.swift
```diff
@@ -38,14 +39,23 @@ extension ExposureSubmissionServiceFactory {
...
 extension ExposureSubmissionServiceFactory {
 	static func create(diagnosiskeyRetrieval: DiagnosisKeysRetrieval, client: Client, store: Store) -> ExposureSubmissionService {
 
+		// default implementation if in non UI testing mode
+		guard isUITesting else {
+			return ENAExposureSubmissionService(
+				diagnosiskeyRetrieval: diagnosiskeyRetrieval,
+				client: client,
+				store: store
+			)
```
nickguendling, line 56: I see a problem with the `isUITesting` approach here (and in other places). You copied this code form above, and now in unit testing this return is tested, while in production, the return in line 31 is executed. That means we can't test the actual implementation in our unit test. It seems that in general, you had to copy code in multiple instances and the production code is not testable anymore in these places.

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -49,8 +49,8 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 		let window = UIWindow(windowScene: windowScene)
 		self.window = window
 
-		#if UITESTING
-		if let isOnboarded = UserDefaults.standard.string(forKey: "isOnboarded") {
+		#if DEBUG
+		if isUITesting, let isOnboarded = UserDefaults.standard.string(forKey: "isOnboarded") {
```
nickguendling: Is it on purpose that only the `isOnboarded` launch argument is exclusive to the UI tests now, and the others are also available in `DEBUG`?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfigurationMock.swift
```diff
@@ -0,0 +1,43 @@
...
+// under the License.
+//
+
+import Foundation
+
+
+final class CachedAppConfigurationMock {
+	init() {
+		#if !DEBUG
+		preconditionFailure("Do not use this mock in a non-DEBUG environment!")
```
nickguendling, line 26: I think you could just wrap the whole class in `#if DEBUG`

src/xcode/fastlane/SnapshotHelper.swift
```diff
@@ -177,14 +179,27 @@ open class Snapshot: NSObject {
...
+                try image.pngData()?.write(to: path, options: .atomic)
             } catch let error {
                 NSLog("Problem writing screenshot: \(name) to \(screenshotsDir)/\(simulator)-\(name).png")
                 NSLog(error.localizedDescription)
             }
         #endif
     }
 
+    class func fixLandscapeOrientation(image: UIImage) -> UIImage {
+        if #available(iOS 10.0, *) {
```
nickguendling, line 191: Do we need this, since our deployment target is iOS 13.6?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
whiskey: That's to distinguish unit from ui tests which has previously been via `UITESTING`. Without the `isUITesting` a unit test would use this `if` branch in case "isOnboarded" is set in the UserDefaults. This was not the case in the previous implementation.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfigurationMock.swift
whiskey: might be. I mostly changed the existing `#if UITESTING` statements to `DEBUG`. I'll review this, thanks!

src/xcode/fastlane/SnapshotHelper.swift
whiskey: SnapshotHelper is imported/copied from Fastlane. We should ignore this file at all :)

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionServiceFactory.swift
whiskey: I see the point and this problem (and maybe others) are a result of the current interpretation of the previous implementation. 

Previously there were `UITESTING` blocks for… well… ui testing only and the 'normal' code for prod, dev and whatever use cases. This results in such glorious chunks of code such as this ServiceFactory. This piece of code (and maybe others) is a 1:1 refactoring with the new `isUITesting` flag. Normally I would expect something like this:

```swift
// ignore the indentation! 
static func create(diagnosiskeyRetrieval: DiagnosisKeysRetrieval, client: Client, store: Store) -> ExposureSubmissionService {
  #if DEBUG
    if isUITesting {
      // UI test code
      return // skip rest of the code - if applicable
    } // `else` would lead to custom code for unit tests - if needed
  #endif
  […] // 'normal' code
}
```

I chose the default `DEBUG` flag to indicate 'test' or 'non-production' code. No need for `UITESTING`, `TESTING` or whatever. My biggest pain point with the extra `UITESTING` flag was, that we needed two targets/schemes for unit and ui testing. Both could be run sequentially in fastlane's `run_tests` (we just did the unit testing) but when it came to the code coverage [`xcov`](https://docs.fastlane.tools/actions/xcov/) only accepted one scheme to use the coverage from - which one to choose? This PR combines these two schemes and replaces an extra precompiler flag with a well known (DEBUG) one.

One final note. There are tons of alternatives how to control the UI in ui testing and how to inject mock data. Using [`Current`](https://www.pointfree.co/blog/posts/21-how-to-control-the-world) is a very nice one but requires a way bigger refactoring than this.

whiskey: I might refactor this whole file to this proposed structure. I skipped this during development to keep the amount of 'noise' from this PR as low as possible.

src/xcode/ENA/ENA/Source/SceneDelegate.swift
nickguendling: Thats true, my concern was more about the two following launch argument, that weren't available outside of UI tests before, but are now.

whiskey: Which ones? There are tons of **undocumented** arguments and environment variables buzzing around. This is also a problem that needs to be tackled but in a different story. Previously there where two schemes (ENATests & ENAUITests) that drifted apart. Now there is one common place to test things. If we screw up with launch parameters, we hopefully get inconsistencies within the tests.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfigurationMock.swift
whiskey: I refactored this and 1-2 other places.

nickguendling (approved):

PR #1323, opened 2020-10-08
Feature/2843 refactor logging

ArturFriesen:
## Description
This PR refactors logging to use os_log.
The logs for the debug menu are saved into a file instead of the NSUserDefaults.
Its now possible to select between log types at the debug menu.

Simulator Screen Shot - iPhone 8 - 2020-10-08 at 14 09 26

Simulator Screen Shot - iPhone 8 - 2020-10-08 at 14 09 08


src/xcode/ENA/ENA/Source/Workers/Logging.swift
```diff
@@ -0,0 +1,149 @@
+import Foundation
+import os.log
+
+extension OSLog {
+
+	private static var subsystem = Bundle.main.unwrappedBundleIdentifier
+
+    /// API interactions
+    static let api = OSLog(subsystem: subsystem, category: "api")
+    /// Exoplanet UI
```
marcussc, line 10: Exoplanet? 😁

src/xcode/ENA/ENA/Source/Workers/Logging.swift
```diff
@@ -0,0 +1,149 @@
...
+    static func warning(_ message: String, log: OSLog = .default) {
+        Self.log(message: message, type: .default, log: log, error: nil)
+    }
+
+    static func error(_ message: String, log: OSLog = .default, error: Error? = nil) {
+        Self.log(message: message, type: .error, log: log, error: error)
+    }
+
+	private static func log(message: String, type: OSLogType, log: OSLog, error: Error?) {
+		os_log("%{private}@", log: log, type: type, message)
```
marcussc, line 35: AFAIK we are not allowed to even to os_log in Production

marcussc (changes requested):
Very nice PR, only got two small remarks

src/xcode/ENA/ENA/Source/Workers/Logging.swift
whiskey: YES! :D

src/xcode/ENA/ENA/Source/Workers/Logging.swift
whiskey: But everything is "private", i.e. log output of the whole `message` will be:
`DATE_FOO <private>`

marcussc: What I am wondering currently, how do we differentiate between TestFlight and AppStore builds?

whiskey: `RELEASE` seems to be set for release builds. So:

```
#if !RELEASE
// log log log!
#endif
```

marcussc: Wouldn't this be the same for TestFlight Builds?

ArturFriesen: TestFlight builds are build with "TestFlight" configuration. And this configuration does not have a RELEASE flag configured in the build settings. So my guess is, logs would work for TestFlight releases.

Bildschirmfoto 2020-10-09 um 10 01 25

ArturFriesen: I have "hidden" os_log call behind !RELEASE

marcussc (approved):

whiskey (approved):

PR #1337, opened 2020-10-12
fix/3174 Empty checks in Exposure Check Log

ArturFriesen:
## Description
This PR fixes the ExposureManager so it is not starting an exposure detection when there is already one running.
Starting exposure detection while another one is running leads to empty logs like described in the story.

## Link to Jira
[3174](https://jira.itc.sap.com/browse/EXPOSUREAPP-3174)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
@@ -120,6 +120,7 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	private weak var exposureManagerObserver: ENAExposureManagerObserver?
 	private var statusObservation: NSKeyValueObservation?
 	@objc private var manager: Manager
+	private var progress: Progress?
 
 	// MARK: Creating a Manager
 
@@ -207,9 +208,21 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	/// Wrapper for `ENManager.detectExposures`
 	/// `ExposureManager` needs to be activated and enabled
 	func detectExposures(configuration: ENExposureConfiguration, diagnosisKeyURLs: [URL], completionHandler: @escaping ENDetectExposuresHandler) -> Progress {
-		manager.detectExposures(configuration: configuration, diagnosisKeyURLs: diagnosisKeyURLs) { summary, error in
+
+		// An exposure detection is currently running. Return current progress.
+		if let progress = progress, !progress.isCancelled && !progress.isFinished {
+			return progress
+		}
+
+		let _progress = manager.detectExposures(configuration: configuration, diagnosisKeyURLs: diagnosisKeyURLs) { [weak self] summary, error in
+			guard let self = self else { return }
+			self.progress = nil
 			completionHandler(summary, error)
 		}
+
+		progress = _progress
+
+		return _progress
 	}
 
 	// MARK: Diagnosis Keys
--- a/src/xcode/ENA/ENA/Source/Models/Exposure/__tests__/ExposureManagerTests.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Exposure/__tests__/ExposureManagerTests.swift
@@ -1,0 +1,112 @@
+//
+// Corona-Warn-App
+//
+// SAP SE and all other contributors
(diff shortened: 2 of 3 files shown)


marcussc (approved):

marcussc:
Very Nice

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -207,9 +208,21 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	/// Wrapper for `ENManager.detectExposures`
 	/// `ExposureManager` needs to be activated and enabled
 	func detectExposures(configuration: ENExposureConfiguration, diagnosisKeyURLs: [URL], completionHandler: @escaping ENDetectExposuresHandler) -> Progress {
-		manager.detectExposures(configuration: configuration, diagnosisKeyURLs: diagnosisKeyURLs) { summary, error in
+
+		// An exposure detection is currently running. Return current progress.
+		if let progress = progress, !progress.isCancelled && !progress.isFinished {
```
mlenkeit, line 213: @ArturFriesen did you also verify that this works through manual tests? IIRC, the `Progress` object returned by ENF is not fully functional, i.e. we don't get a real progress of the exposure detection (which we do not need here), I think calling `cancel()` to interrupt the exposure detection is not supported and a few other strange behaviors. So I'm just wondering if `isCancelled` and `isFinished` work reliably on a real device.

ArturFriesen: Yes I tested it on device. Basically it is working with the optionality of progress. It is set to nil on completion.
Additionally as far as I know, cancel() works and needs to be taken into account here. Christian fixed a bug by calling cancel() on timeout, see [here](https://github.com/corona-warn-app/cwa-app-ios/blob/e022c608cc4512019d581a811225fef1d2182a44/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift#L280) . "isFinished" is not working, because the progress is not tracked inside of ENF. I have added it for the case it could start work in the future.

src/xcode/ENA/ENA/Source/Models/Exposure/__tests__/ExposureManagerTests.swift
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		_ = exposureManager.detectExposures(configuration: ENExposureConfiguration(), diagnosisKeyURLs: []) { _, _ in
+		}
+
+		XCTAssertEqual(managerSpy.numberOfDetectionCalls, 1)
+	}
+}
+
+private final class ManagerSpy: NSObject, Manager {
```
nickguendling, line 85: Is this a new fancy name for a mock manager? 🕵🏼‍♂️😎

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -207,9 +208,21 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
+			return progress
+		}
+
+		let _progress = manager.detectExposures(configuration: configuration, diagnosisKeyURLs: diagnosisKeyURLs) { [weak self] summary, error in
```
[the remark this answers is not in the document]
ArturFriesen: Returning self.progress would fail because it's optional. This way I don't need to unwrap self.progress.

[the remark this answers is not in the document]
nickguendling: Okay, understood, it's because `self.progress` is optional.

nickguendling (approved):

src/xcode/ENA/ENA/Source/Models/Exposure/__tests__/ExposureManagerTests.swift
ArturFriesen: It's just a more precise naming for this Double. I first read about it from Martin Fowler: https://martinfowler.com/articles/mocksArentStubs.html

Here is another article from objc.io: 
https://www.objc.io/issues/15-testing/mocking-stubbing/

PR #1340, opened 2020-10-14
Refactoring To Get Rid of All Swiftlint Error Messages

tcfos:
## Description
Refactoring, code improvements and a few modified swiftlint rules to get rid of all swiftlint error messages.

## Link to Jira
https://jira.itc.sap.com/browse/EXPOSUREAPP-2901


src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
```diff
@@ -206,64 +207,78 @@ extension ExposureNotificationSettingViewController {
...
+				(cell as? ActionCell)?.configure(for: enState, delegate: self)
+			}
+		case .descriptionCell:
+			cell.configure(for: enState)
+		}
+
+		return cell
+	}
+
+	private func euTracingCell(for indexPath: IndexPath, in tableView: UITableView) -> UITableViewCell {
```
KaiTeuber, line 242: Since this has been moved to private methods, I think it would be good to use the MARKs

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionCoordinatorModelTests.swift
```diff
@@ -63,6 +65,7 @@ class ExposureSubmissionCoordinatorModelTests: XCTestCase {
 		XCTAssertFalse(model.exposureSubmissionServiceHasRegistrationToken)
 	}
 
+	// MARK: -
```
KaiTeuber, line 68: Why are empty marks used here?
The description of the test method should be sufficient here.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfigurationMock.swift
```diff
@@ -20,19 +20,29 @@
...
+final class CachedAppConfigurationMock: AppConfigurationProviding {
 
-extension CachedAppConfigurationMock: AppConfigurationProviding {
+	// MARK: - Init
+
+	init(appConfigurationResult: Result<SAP_ApplicationConfiguration, Error> = .success(SAP_ApplicationConfiguration())) {
+		self.appConfigurationResult = appConfigurationResult
+	}
+
+	// MARK: - Protocol AppConfigurationProviding
```
KaiTeuber, line 31: 👍

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
tcfos: How shall we use MARKS to structure a file with
1 protocol, 1 class, and 6 extensions? Any suggestions?

KaiTeuber: Basically the question arises why all this is in one file.
Within the class I would set the MARKs and then of course distribute them in the extensions.
Superfluous marks can then be removed.
Keep the protocol on top - I guess we will refactor most of them later.

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationCellModel.swift
```diff
@@ -0,0 +1,26 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import Foundation
+import UIKit
+
+struct AppInformationCellModel {
```
whiskey, line 22: According to the search we have `AppInformationLegalCell` and `AppInformationDynamicCell`. We should consider good 'ol inheritance and put this model as inline struct into a common `AppInformationCell`. Less files, better code.

I can help with this!

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/__tests__/RiskCalculationTests.swift
```diff
@@ -21,7 +21,8 @@
 import ExposureNotification
 import XCTest
 
-final class RiskCalculationTests: XCTestCase {
+// swiftlint:disable:next type_body_length
+class RiskCalculationTests: XCTestCase {
```
whiskey, line 25: Is `final` an option here?

```suggestion
final class RiskCalculationTests: XCTestCase {
```

src/xcode/ENA/.swiftlint.yml
```diff
@@ -240,7 +241,7 @@ type_body_length:
   error: 400
   
 function_parameter_count:
-  warning: 6
+  warning: 7
```
whiskey, line 244: Personally, I think even 5 parameters is too much for normal code. It may be valid in linear algebra and graphics programming. Here we should consider configuration structs/models to be passed into bigger functions.

whiskey (approved):
Some remarks & comments but no real show stopper.

KaiTeuber (approved):

src/xcode/ENA/ENA.xcodeproj/project.pbxproj
```diff
@@ -4597,7 +4602,7 @@
 			buildSettings = {
 				CODE_SIGNING_ALLOWED = NO;
 				CODE_SIGNING_REQUIRED = NO;
-				CODE_SIGN_IDENTITY = "Apple Development";
```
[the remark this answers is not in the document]
nickguendling: No, I'm not sure, Xcode suggested this change for the DEBUG build. Do you know if we definitely do not want this? Otherwise we might have to try it out, if it breaks something in our build process. So far the CI has not been unhappy about it. What do you think?

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
nickguendling: I added a private MARK for now, but this class needs an additional structural cleanup later

src/xcode/ENA/ENA.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
marcussc: Okay, if it comes from the Xcode suggested changes, lets give it a try :D

PR #1344, opened 2020-10-14
Fix/exposure detection executor completion callbacks

ArturFriesen:
## Description
This PR streamlines the persistence of the keys and calls completion once when all keys are persisted.
This way the risk detection starts only when all keys are persisted.

## Link to Jira
[3242](https://jira.itc.sap.com/browse/EXPOSUREAPP-3242)


whiskey (approved):
Ok for a quick fix 🧰🛠

SwissMountainFrog (approved):

PR #1345, opened 2020-10-14
fix/3185 Reset Deadman Switch in Foreground

marcussc:
## Description
The Deadman Switch should also be reseted if a Risk-Calculation was done in the Foreground.

## Link to Jira
https://jira.itc.sap.com/browse/EXPOSUREAPP-3185

### DIFF
--- a/src/xcode/ENA/ENA/Source/Extensions/UNUserNotificationCenter+DeadManSwitch.swift
+++ b/src/xcode/ENA/ENA/Source/Extensions/UNUserNotificationCenter+DeadManSwitch.swift
@@ -1,0 +1,75 @@
+//
+// Corona-Warn-App
+//
+// SAP SE and all other contributors
+// copyright owners license this file to you under the Apache
+// License, Version 2.0 (the "License"); you may not use this
+// file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+import UserNotifications
+
+extension UNUserNotificationCenter {
+
+	// MARK: - Internal
+	
+	/// Schedules a local notification to fire 36 hours from now, if there isn´t a notification already scheduled
+	func scheduleDeadmanNotificationIfNeeded() {
+
+		// Check if Deadman Notification is already scheduled
+		getPendingNotificationRequests(completionHandler: { notificationRequests in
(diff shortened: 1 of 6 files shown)


SwissMountainFrog (approved):

src/xcode/ENA/ENA/Source/Extensions/UNUserNotificationCenter+DeadManSwitch.swift
```diff
@@ -0,0 +1,75 @@
...
+		getPendingNotificationRequests(completionHandler: { notificationRequests in
+			if notificationRequests.contains(where: { $0.identifier == UNUserNotificationCenter.deadManNotificationIdentifier }) {
+				// Deadman Notification already setup -> return
+				return
+			} else {
+				// No Deadman Notification setup, contiune to setup a new one
+				let content = UNMutableNotificationContent()
+				content.title = AppStrings.Common.deadmanAlertTitle
+				content.body = AppStrings.Common.deadmanAlertBody
+				content.sound = .default
```
whiskey, line 39: We NEED a default sound!

src/xcode/ENA/ENA/Source/Extensions/UNUserNotificationCenter+DeadManSwitch.swift
```diff
@@ -0,0 +1,75 @@
...
+	
+	/// Reset the Deadman Notification, should be called after a successfull risk-calculation.
+	func resetDeadmanNotification() {
+		cancelDeadmanNotification()
+		scheduleDeadmanNotificationIfNeeded()
+	}
+	
+	// MARK: - Private
+	
+	private static let deadManNotificationIdentifier = (Bundle.main.bundleIdentifier ?? "") + ".notifications.cwa-deadman"
```
whiskey, line 69: Is dead<i>man</i> still a valid term?

whiskey (approved):

src/xcode/ENA/ENA/Source/Extensions/UNUserNotificationCenter+DeadManSwitch.swift
ArturFriesen: deadHumanNotificationIdentifier?

whiskey: A dead person does not tap/click. How about something with 'inactivity'?

marcussc: I think this was just a plain translation from: https://de.wikipedia.org/wiki/Totmanneinrichtung "Totmannschalter" This whole Feature is called "Deadman Switch" so I think renaming this would create a lot of confusion.

whiskey: Same for [Master/Slave](https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns) , etc.
Just mentioning… I already approved 🙃🏳️‍🌈

PR #1355, opened 2020-10-15
Fix/3246 Correct Flow Logic in Onboarding

tcfos:
## Description
Fixes a bug in the onboarding flow logic. Once the user has enabled logging of contacts then the button "Activate Exposure Logging" is not displayed any more when navigating back and forth in the onboarding process.

### How to Test
1 Onboarding Screen 3 - How to make exposure logging possible: Select "Do Not Activate".
2 on the next screen tap "Back".
3 Onboarding Screen 3: both buttons are displayed. Scroll down to the bottom of the view and check the content of screen 3.
4 Tap on "Activate Exposure Logging" (Screenshot 1).
5 On the next screen tap "Back" (Screenshot 2).
6 Onboarding Screen 3: only one button is displayed (Screenshot 3).
7 Scroll down to the bottom of the view and verify the content of screen 3 is still OK.

## Link to Jira
Story: https://jira.itc.sap.com/browse/EXPOSUREAPP-1692 
Task: https://jira.itc.sap.com/browse/EXPOSUREAPP-3246

### Screenshot 1
IMG_0037
### Screenshot 2
IMG_0038
### Screenshot 3
IMG_0039

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
@@ -19,24 +19,6 @@
 
 import UIKit
 
-
-enum OnboardingPageType: Int, CaseIterable {
-	case togetherAgainstCoronaPage = 0
-	case privacyPage = 1
-	case enableLoggingOfContactsPage = 2
-	case howDoesDataExchangeWorkPage = 3
-	case alwaysStayInformedPage = 4
-
-	func next() -> OnboardingPageType? {
-		OnboardingPageType(rawValue: rawValue + 1)
-	}
-
-	func isLast() -> Bool {
-		(self == OnboardingPageType.allCases.last)
-	}
-}
-
-
 extension OnboardingInfoViewController {
 
 	func addPanel(
--- a/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
@@ -19,6 +19,24 @@
 
 import UIKit
 
+
+enum OnboardingPageType: Int, CaseIterable {
+	case togetherAgainstCoronaPage = 0
+	case privacyPage = 1
+	case enableLoggingOfContactsPage = 2
+	case howDoesDataExchangeWorkPage = 3
+	case alwaysStayInformedPage = 4
+
+	func next() -> OnboardingPageType? {
+		OnboardingPageType(rawValue: rawValue + 1)
+	}
+
+	func isLast() -> Bool {
+		(self == OnboardingPageType.allCases.last)
+	}
+}
+
+
 extension OnboardingInfoViewController {
 
 	func addPanel(
@@ -32,7 +50,6 @@ extension OnboardingInfoViewController {
 		itemSpacing: CGFloat = 10
 	) {
 
-		// MARK: - Title label.
 		let titleLabel = ENALabel()
 		titleLabel.style = titleStyle
 		titleLabel.text = title
@@ -41,7 +58,6 @@ extension OnboardingInfoViewController {
 		titleLabel.lineBreakMode = .byWordWrapping
 		titleLabel.numberOfLines = 0
 
-		// MARK: - Text label.
 		let textLabel = ENALabel()
 		textLabel.style = bodyStyle
 		textLabel.text = body
@@ -50,7 +66,6 @@ extension OnboardingInfoViewController {
 		textLabel.lineBreakMode = .byWordWrapping
 		textLabel.numberOfLines = 0
 
-		// MARK: - Stackview for labels.
 		let labelStackView = UIStackView(arrangedSubviews: [titleLabel, textLabel])
 		labelStackView.translatesAutoresizingMaskIntoConstraints = false
 		labelStackView.axis = .vertical
@@ -58,7 +73,6 @@ extension OnboardingInfoViewController {
 		labelStackView.distribution = .equalSpacing
 		labelStackView.spacing = itemSpacing
 
-		// MARK: - Container for entire view.
 		let containerView = UIView()
 		containerView.addSubview(labelStackView)
 		containerView.layer.cornerRadius = 14.0
@@ -102,7 +116,6 @@ extension OnboardingInfoViewController {
 
 		let containerView = UIView()
 
-		// MARK: - Create country table.
 		let countryListView = UIStackView()
 		countryListView.translatesAutoresizingMaskIntoConstraints = false
 		countryListView.axis = .vertical
@@ -111,7 +124,6 @@ extension OnboardingInfoViewController {
 		countryListView.spacing = 7.5
 		containerView.addSubview(countryListView)
 
-		// MARK: - Create title label.
 		let titleLabel = ENALabel()
 		titleLabel.style = .headline
 		titleLabel.text = title
@@ -122,18 +134,15 @@ extension OnboardingInfoViewController {
 		containerView.addSubview(titleLabel)
 
 		for (index, country) in countries.enumerated() {
-			// MARK: - Create stack view to hold label and flag for country.
 			let stackView = UIStackView()
 			stackView.axis = .horizontal
 			stackView.alignment = .center
 			stackView.spacing = 13
 
-			// MARK: - Country name.
 			let label = ENALabel()
 			label.style = .body
 			label.text = country.localizedName
 
-			// MARK: - Country flag.
 			let image = UIImageView(image: country.flag)
 			image.widthAnchor.constraint(equalToConstant: 28).isActive = true
 			image.contentMode = .scaleAspectFit
@@ -142,7 +151,6 @@ extension OnboardingInfoViewController {
 			stackView.addArrangedSubview(label)
 			countryListView.addArrangedSubview(stackView)
 
-			// MARK: - Add separator inbetween each.
 			if index == countries.count - 1 { break }
 			let separator = UIView()
 			separator.backgroundColor = .enaColor(for: .hairline)
@@ -185,5 +193,26 @@ extension OnboardingInfoViewController {
 		let containerView = createCountrySection(title: title, countries: countries)
 		stackView.addArrangedSubview(containerView)
 	}
+
+}
+
+
+extension OnboardingInfoViewController: UITextViewDelegate {
+	func textView(_ textView: UITextView, shouldInteractWith url: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+		LinkHelper.openLink(withUrl: url, from: self)
+		return false
+	}
+}
+
+
+extension OnboardingInfoViewController: NavigationBarOpacityDelegate {
+	var preferredNavigationBarOpacity: CGFloat {
+		let alpha = (scrollView.adjustedContentInset.top + scrollView.contentOffset.y) / scrollView.adjustedContentInset.top
+		return max(0, min(alpha, 1))
+	}
+}
+
+
+extension OnboardingInfoViewController: RequiresAppDependencies {
 
 }
--- a/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
@@ -178,6 +178,7 @@ extension OnboardingInfoViewController {
 
 }
 
+// MARK: - Protocol UITextViewDelegate
 
 extension OnboardingInfoViewController: UITextViewDelegate {
 	func textView(_ textView: UITextView, shouldInteractWith url: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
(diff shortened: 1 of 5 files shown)


src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -227,11 +228,19 @@ final class OnboardingInfoViewController: UIViewController {
 		stateStateLabel.text = exposureNotificationsEnabled ? onboardingInfo.stateActivated : onboardingInfo.stateDeactivated
 
 		guard !pageSetupDone else {
+			// Back button tapped
+			// If "logging of contacts" has been enabled then display alternative text on buttom
+			if pageType == .enableLoggingOfContactsPage && showStateView {
+				nextButton.setTitle(onboardingInfo.alternativeActionText, for: .normal)
+			}
```
nickguendling, line 235: Seems like you set the alternative title independent of whether `pageSetupDone` is `true` or `false`. I think you could just move this if statement above the guard and remove it from the switch below.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -227,11 +228,19 @@ final class OnboardingInfoViewController: UIViewController {
 		stateStateLabel.text = exposureNotificationsEnabled ? onboardingInfo.stateActivated : onboardingInfo.stateDeactivated
 
 		guard !pageSetupDone else {
```
ArturFriesen, line 230: It's getting hard to understand the overall logic.
In generell I think we should, if possible, separate creating the UI and updating the UI to different funcs.

I would propose a quick fix for better readable. Use "if else" instead of "guard". What do you think?

if pageSetupDone {
 ....
} else {
 ....
}

pageSetupDone = true

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
ArturFriesen: And maybe move it to, or combine it with, the already existing line:
"nextButton.setTitle(onboardingInfo.actionText, for: .normal)"

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: refactored the code. Introduced new method pageSetup()

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: Thanks, I've done some refactoring.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -108,20 +108,24 @@ final class OnboardingInfoViewController: UIViewController {
 		if pageType == .togetherAgainstCoronaPage { loadCountryList() }
 	}
 
+
```
whiskey, line 111: Is 'double empty line' the new style for the code?

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: Separated updateUI() and pageSetup(); setting the alternative button text is now independent of pageSetupDone

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: Do we have a rule for this? I prefer two empty lines between funcs.

whiskey: Keep it as it is! :) We can discuss solutions here in the next Dev Camp.

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
```diff
@@ -19,6 +19,24 @@
 
 import UIKit
 
+
+enum OnboardingPageType: Int, CaseIterable {
```
nickguendling, line 23: I think it makes more sense to move this to a separate file instead of this extension collection

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
```diff
@@ -122,18 +134,15 @@ extension OnboardingInfoViewController {
 		containerView.addSubview(titleLabel)
 
 		for (index, country) in countries.enumerated() {
-			// MARK: - Create stack view to hold label and flag for country.
 			let stackView = UIStackView()
 			stackView.axis = .horizontal
 			stackView.alignment = .center
 			stackView.spacing = 13
 
-			// MARK: - Country name.
```
nickguendling, line 131: I agree these marks are too much, maybe we can have a better variable naming here instead, like `countryNameLabel`

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
```diff
@@ -122,18 +134,15 @@ extension OnboardingInfoViewController {
...
 			let label = ENALabel()
 			label.style = .body
 			label.text = country.localizedName
 
-			// MARK: - Country flag.
```
nickguendling, line 136: Same here, maybe `countryFlagImageView`

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
```diff
@@ -187,3 +195,24 @@ extension OnboardingInfoViewController {
 	}
 
 }
+
+
+extension OnboardingInfoViewController: UITextViewDelegate {
```
nickguendling, line 200: We have a `// MARK: - Protocol <Protocol Name>` in our set of marks. Could you add the protocol conformances directly into the class and use the marks?

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -19,29 +19,40 @@ import UIKit
...
-	}
 
-	func isLast() -> Bool {
-		(self == OnboardingPageType.allCases.last)
-	}
-}
+final class OnboardingInfoViewController: UIViewController {
 
-extension OnboardingInfoViewController: RequiresAppDependencies {
+	@IBOutlet var imageView: UIImageView!
```
nickguendling, line 25: All of the `@IBOutlet`s should go to the `// MARK: - Internal` section. Or, even better, make them private and put them into the `// MARK: - Private` section.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -19,29 +19,40 @@ import UIKit
...
-}
+	@IBOutlet var scrollView: UIScrollView!
+	@IBOutlet var stackView: UIStackView!
+	@IBOutlet var stateView: UIView!
+	@IBOutlet var innerStackView: UIStackView!
+	@IBOutlet var footerView: UIView!
 
-final class OnboardingInfoViewController: UIViewController {
-	// MARK: Creating a Onboarding View Controller
+	private var pageType: OnboardingPageType
```
nickguendling, line 42: These should go into the `// MARK: - Private` section

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
nickguendling: We only have on line between funcs basically everywhere else, so it would be better for the sake of consistency to only have one in this file as well

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -19,29 +19,40 @@ import UIKit
...
+	private var onboardingInfo: OnboardingInfo?
+	private var supportedCountries: [Country]?
+	private var client: Client
 
+	private var pageSetupDone = false
+	private var onboardingInfos = OnboardingInfo.testData()
+	private var exposureManagerActivated = false
+	
+	
+	// MARK: - Init
```
nickguendling, line 55: Since a mark is the headline of a whole section and not just documentation for one single var or func, I think there should be an empty line after the mark

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -108,20 +108,24 @@ final class OnboardingInfoViewController: UIViewController {
...
+	
 	override func viewDidLayoutSubviews() {
 		super.viewDidLayoutSubviews()
 
 		scrollView.contentInset.bottom = footerView.frame.height - scrollView.safeAreaInsets.bottom
 		scrollView.verticalScrollIndicatorInsets.bottom = scrollView.contentInset.bottom
 	}
 
+	
+	// MARK: - Public
```
nickguendling, line 128: These are actually `Internal`, not `Public` by default. And maybe you can quickly check if they could be set to private or are actually used from outside the class :)

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -226,10 +301,19 @@ final class OnboardingInfoViewController: UIViewController {
 		stateTitleLabel.text = onboardingInfo.stateTitle
 		stateStateLabel.text = exposureNotificationsEnabled ? onboardingInfo.stateActivated : onboardingInfo.stateDeactivated
 
-		guard !pageSetupDone else {
+		if pageType == .enableLoggingOfContactsPage && showStateView {
+			nextButton.setTitle(onboardingInfo.alternativeActionText, for: .normal)
+		}
```
nickguendling, line 306: I think Artur initially suggested to combine this with the original `nextButton.setTitle(onboardingInfo.actionText, for: .normal)` and I like that suggestion. You could move it up before line `nextButton.isHidden = onboardingInfo.actionText.isEmpty` and write it like this:

```
if pageType == .enableLoggingOfContactsPage && showStateView {
    nextButton.setTitle(onboardingInfo.alternativeActionText, for: .normal)
} else {
    nextButton.setTitle(onboardingInfo.actionText, for: .normal)
}
```

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
tcfos: OK, new file for enum

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: moved @IBAction funcs to private section

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: moved private vars to privat section

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: Empty line after //MARK.
Changed to 1 empty line between funcs. To enforce this we could set the swiftlint rule:

- vertical_whitespace:  max_empty_lines: 1

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: Good point. All these funcs can be declared as private. Question: how do you sort methods in a section? Alphabetically?

nickguendling: There is no strict rule, but my feeling is that I usually put methods that belong together next to each other, and often the ones that are called earlier are more at the top. But I guess everyone has their own preferences there. I think the order should somehow make sense and not just be alphabetically.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
tcfos: You're right, this avoids confusion.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController+Extension.swift
tcfos: New MARKs for extensions conforming to protocols

nickguendling (approved):

PR #1359, opened 2020-10-16
Clear the key packages and app config on ENError = 2

ArturFriesen:
## Description
Resets the key packages and app config when ENError = 2 (badParameter) occurs.

## Link to Jira
[3297](https://jira.itc.sap.com/browse/EXPOSUREAPP-3297)

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate.swift
@@ -23,7 +23,7 @@ import UIKit
 protocol CoronaWarnAppDelegate: AnyObject {
 	var client: HTTPClient { get }
 	var downloadedPackagesStore: DownloadedPackagesStore { get }
-	var store: Store { get }
+	var store: Store & AppConfigCaching { get }
 	var appConfigurationProvider: AppConfigurationProviding { get }
 	var riskProvider: RiskProvider { get }
 	var exposureManager: ExposureManager { get }
@@ -99,7 +99,7 @@ extension AppDelegate: ExposureSummaryProvider {
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
-	let store: Store
+	let store: Store & AppConfigCaching
 	let serverEnvironment: ServerEnvironment
 	
 	private let consumer = RiskConsumer()
@@ -109,11 +109,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		// use a custom http client that uses/recognized caching mechanisms
 		let appFetchingClient = CachingHTTPClient(clientConfiguration: client.configuration)
 
-		// we currently use the store as common place for temporal persistency
-		guard let store = store as? AppConfigCaching else {
-			preconditionFailure("Ensure to provide a proper app config cache")
-		}
-		
 		return CachedAppConfiguration(client: appFetchingClient, store: store)
 	}()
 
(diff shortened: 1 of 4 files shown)


src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -109,11 +109,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 		// use a custom http client that uses/recognized caching mechanisms
 		let appFetchingClient = CachingHTTPClient(clientConfiguration: client.configuration)
 
-		// we currently use the store as common place for temporal persistency
-		guard let store = store as? AppConfigCaching else {
```
marcussc, line 113: 👍

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
```diff
@@ -168,11 +168,26 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
...
+		// Clear the key packages and app config on ENError = 2 = .badParameter
+		// For more details, see: https://jira.itc.sap.com/browse/EXPOSUREAPP-3297
+		func clearCacheOnError2(error: Error) {
+			if let enError = error as? ENError, enError.code == .badParameter {
+				// Clear the key packages
+				downloadedPackagesStore.reset()
+				downloadedPackagesStore.open()
+
+				// Clear the app config
+				store.appConfig = nil
```
marcussc, line 181: Just to make sure, every change to the store gets instantly persisted to Disk?

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
ArturFriesen: Its reseting the app config cache. Also added etag and date reset.

ArturFriesen: Its persisted here: https://github.com/corona-warn-app/cwa-app-ios/blob/3c8416d309651783b2c1dafff446661bcac14e9d/src/xcode/ENA/ENA/Source/Workers/SQLiteKeyValueStore.swift#L138

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
```diff
@@ -168,11 +168,28 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
 			writtenPackages: WrittenPackages,
 			completion: @escaping (Result<ENExposureDetectionSummary, Error>) -> Void
 	) -> Progress {
+
+		// Clear the key packages and app config on ENError = 2 = .badParameter
+		// For more details, see: https://jira.itc.sap.com/browse/EXPOSUREAPP-3297
+		func clearCacheOnError2(error: Error) {
```
KaiTeuber, line 174: why call this func ..OnError2 and explain the error code in the comment?
I think we should use a better name here, like
clearCacheOnErrorBadParameter

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
ArturFriesen: :) Renamed!

PR #1370, opened 2020-10-20
fix/flaky_unit_test

KaiTeuber:
## Description
fixed a flaky unit test, success callbacks only work in stored test results
.expired state is no longer stored

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
+++ b/src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
@@ -19,6 +19,8 @@
 import ExposureNotification
 import XCTest
 
+
+// swiftlint:disable file_length
 // swiftlint:disable:next type_body_length
 class ExposureSubmissionServiceTests: XCTestCase {
 	let expectationsTimeout: TimeInterval = 2
--- a/src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
+++ b/src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
@@ -401,7 +401,8 @@ class ExposureSubmissionServiceTests: XCTestCase {
 		let client = ClientMock()
 		store.registrationToken = "dummyRegistrationToken"
 
-		let testResult = TestResult.allCases.randomElement() ?? TestResult.positive
+		let storedResults: [TestResult] = [.pending, .negative, .positive, .invalid]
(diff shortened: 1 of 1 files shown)


whiskey:
So shouldn't we then create a test to expect an error response in case of `.expired` results?

src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
```diff
@@ -401,7 +401,8 @@ class ExposureSubmissionServiceTests: XCTestCase {
 		let client = ClientMock()
 		store.registrationToken = "dummyRegistrationToken"
 
-		let testResult = TestResult.allCases.randomElement() ?? TestResult.positive
+		let storedResults: [TestResult] = [.pending, .negative, .positive, .invalid]
```
whiskey, line 404: So shouldn't we then create a test to expect an error response in case of .expired results?

src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
```diff
@@ -401,7 +401,8 @@ class ExposureSubmissionServiceTests: XCTestCase {
...
+		let testResult = storedResults.randomElement() ?? TestResult.positive
```
nickguendling, line 405: Wouldn't it be great to have a separate test for every test result case, so we can immediately see that something is wrong and don't have to wait that `.randomElement()` chooses a certain case? I think we've just experienced the drawback of this random approach.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
KaiTeuber: Yes, maybe - I only fixed the behavior I found

src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
KaiTeuber: Added all cases as a test

nickguendling (approved):

src/xcode/ENA/ENA/Source/Services/__tests__/ExposureSubmissionServiceTests.swift
KaiTeuber: I have included all other cases in tests to make it more reliable

marcussc (approved):

PR #1372, opened 2020-10-20
Fix/3275 reset riskcalculation after config change

nickguendling:
## Description
Invoke the risk detection immediately after a changed app configuration is loaded to get an updated risk score. If the detection fails, the previous summary is kept and the calculation is still applied on the cached summary.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3275

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
@@ -104,24 +104,31 @@ final class ExposureDetection {
 	typealias Completion = (Result<ENExposureDetectionSummary, DidEndPrematurelyReason>) -> Void
 
 	func start(completion: @escaping Completion) {
-		self.completion = completion
 		Log.info("ExposureDetection: Start downloading packages.", log: .riskDetection)
 
+		self.completion = completion
 		activityState = .downloading
 
 		downloadKeyPackages { [weak self] in
 			guard let self = self else { return }
+
 			Log.info("ExposureDetection: Completed downloading packages.", log: .riskDetection)
 			Log.info("ExposureDetection: Start writing packages to file system.", log: .riskDetection)
+
 			self.writeKeyPackagesToFileSystem { [weak self] writtenPackages in
 				guard let self = self else { return }
+
 				Log.info("ExposureDetection: Completed writing packages to file system.", log: .riskDetection)
+
 				self.activityState = .detecting
 
 				if let exposureConfiguration = self.exposureConfiguration {
-          Log.info("ExposureDetection: Start detecting summary.", log: .riskDetection)
+					Log.info("ExposureDetection: Start detecting summary.", log: .riskDetection)
+
 					self.detectSummary(writtenPackages: writtenPackages, exposureConfiguration: exposureConfiguration)
 				} else {
+					Log.error("ExposureDetection: End prematurely.", log: .riskDetection, error: DidEndPrematurelyReason.noExposureConfiguration)
+
 					self.endPrematurely(reason: .noExposureConfiguration)
 				}
 			}
--- a/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
@@ -27,7 +27,7 @@ final class ExposureDetection {
 	private var completion: Completion?
 	private var progress: Progress?
 	private var countryKeypackageDownloader: CountryKeypackageDownloading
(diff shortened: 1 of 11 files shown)


KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
```diff
@@ -126,12 +117,12 @@ final class ExposureDetection {
...
-					guard let self = self else { return }
-					Log.info("ExposureDetection: Completed loading exposure configuration.", log: .riskDetection)
-					Log.info("ExposureDetection: Start detecting summary.", log: .riskDetection)
-					self.detectSummary(writtenPackages: writtenPackages, exposureConfiguration: configuration)
+
+				if let exposureConfiguration = self.exposureConfiguration {
+          Log.info("ExposureDetection: Start detecting summary.", log: .riskDetection)
+					self.detectSummary(writtenPackages: writtenPackages, exposureConfiguration: exposureConfiguration)
+				} else {
+					self.endPrematurely(reason: .noExposureConfiguration)
```
marcussc, line 125: Can we log this as well please? :=)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -17,7 +17,7 @@
 // under the License.
 //
 
-import Foundation
+import UIKit
```
marcussc, line 20: Hmm not sure if I understand why.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
```diff
@@ -27,7 +27,7 @@ final class ExposureDetection {
 	private var completion: Completion?
 	private var progress: Progress?
 	private var countryKeypackageDownloader: CountryKeypackageDownloading
-	private let appConfigurationProvider: AppConfigurationProviding
+	private let appConfiguration: SAP_ApplicationConfiguration
```
marcussc, line 30: Why are we losing the protocol here?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -105,68 +107,52 @@ extension RiskProvider: RiskProviding {
...
-			activeTracingHours: store.tracingStatusHistory.activeTracing().inHours,
-			lastExposureDetectionDate: store.summary?.date
-		)
-		if !enoughTimeHasPassed || !self.exposureManagerState.isGood {
-			completion(
-				.init(
-					previous: nil,
-					current: store.summary
-				)
+		if !ignoreCachedSummary {
```
marcussc, line 123: How do we make sure that we don't run into the rate-limit?

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
nickguendling: Done 👍

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
nickguendling: Reverted that change now, was needed for a previous approach that we abandoned.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
nickguendling: Because we can directly inject the actual App Configuration instead of the provider that tries to get it from the cache. The only caller of the exposure detection's initializer retrieves a cached app configuration beforehand anyway. That way we can ensure that all involved instances work with the same configuration. And we were able to make the exposure detection's dependency on the app configuration more clear by passing it in.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
nickguendling: Because we assume an app configuration change is a rare event, and we are currently not using the api close to it's rate limit, we don't take any precautions here. We just give it a try, and if it fails, we keep the old summary until the next detection happens by itself.

marcussc: Okay makes sense, thanks for the explanation 👍

marcussc (approved):

PR #1373, opened 2020-10-20
Remove buggy display of nextTriggerDate from the DM

marcussc:
## Description
`nextTriggerDate` does not return the `nextTriggerDate` but the currentDate + the timeInterval with which the notification was setup with.  Since this bug seems not to get fixed by Apple soon (https://stackoverflow.com/questions/51618620/nexttriggerdate-doesnt-return-the-expected-value-is-there-another-way-to-o) we have to remove the display of the `nextTriggerDate` because it causes more confusion than it does good.
## Screenshot
Screen Shot 2020-10-20 at 15 27 36

### DIFF
--- a/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMNotificationsViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMNotificationsViewController.swift
@@ -60,7 +60,7 @@ final class DMNotificationsViewController: UITableViewController {
 		
 		cell.textLabel?.text = notificationRequest.identifier
 		
-		// Dear future Develeoper that thinks it would be nice to display the nextTriggerDate of the notificationRequest
+		// Dear future developer that thinks it would be nice to display the nextTriggerDate of the notificationRequest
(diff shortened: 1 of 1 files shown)


KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMNotificationsViewController.swift
```diff
@@ -60,14 +60,9 @@ final class DMNotificationsViewController: UITableViewController {
...
 		
-		guard let trigger = notificationRequest.trigger as? UNTimeIntervalNotificationTrigger, let triggerDate = trigger.nextTriggerDate() else {
-			return cell
-		}
-		
-		let df = DateFormatter()
-		df.dateFormat = "yyyy-MM-dd hh:mm:ss"
-
-		cell.detailTextLabel?.text = df.string(from: triggerDate)
+		// Dear future Develeoper that thinks it would be nice to display the nextTriggerDate of the notificationRequest
```
KaiTeuber, line 63:
```suggestion
		// Dear future developer that thinks it would be nice to display the nextTriggerDate of the notificationRequest
```

tcfos (approved):

PR #1394, opened 2020-10-22
Feature/3392 central protobuf files

marcussc:
## Description
This updates our project to run with the new central managed protobuf files.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3392


proto/convert_proto_to_swift.sh
```diff
@@ -2,10 +2,8 @@
 
 ## This script converts *.proto files to *.pb.swift files.
 ## In order to work properly protoc should be installed.
-cd resources
 
-set -euo pipefail
```
whiskey, line 7: Should be part of most scripts here. It is default in CircleCI but not when you run it locally

```suggestion
set -euo pipefail
```

further reading: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

proto/resources/external/exposurenotification/diagnosis_key_batch.proto
```diff
@@ -15,14 +17,13 @@ message DiagnosisKey {
   ReportType reportType = 7; // set by backend
   // Number of days elapsed between symptom onset and the TEK being used.
   // E.g. 2 means TEK is 2 days after onset of symptoms.
-  uint32 days_since_onset_of_symptoms = 8;
+  sint32 days_since_onset_of_symptoms = 8;
```
whiskey, line 20: Just a comment but `sint32` makes no sense (although no difference) here.

whiskey (changes requested):
(Blocking because of the `pipefail`.)

So, no Swift-Packages then?

proto/convert_proto_to_swift.sh
marcussc: Ah thanks did not know about this 👍

proto/resources/external/exposurenotification/diagnosis_key_batch.proto
marcussc: Since we don't create these files there is nothing I can do at this moment. 
But we might be able to change it here: https://github.com/corona-warn-app/cwa-protocol-buffers/tree/main/src/proto/external/exposurenotification

marcussc:
> (Blocking because of the `pipefail`.)
> 
> So, no Swift-Packages then?

This PR is just to get us aligned with the central protobuf files and make it easer for upcoming changes. We can still create a Swift-Package for this.

whiskey (approved):

ArturFriesen (approved):

PR #1402, opened 2020-10-23
Feature/2971 decrease ci time

whiskey:
This PR is the first step to decrease CI build & test time. It 

1. restructured workflows to start with a common operations and split up for build, test and screenshot jobs (see image)
1. uses CircleCI's caching to keep the dependencies until the `Gemfile.lock` will change (has still room for improvement)
1. uses faster animation speeds to reduce UI test time (thanks @marcussc)

All current workflows run at once: (only for illustration)
CI - all workflows


ArturFriesen (approved):
🚀 Like it!

marcussc (approved):

whiskey:
@marcussc @kaddaSz If this PR is accepted we need to tweak the GH actions to check on `build-and-test` instead of the old `build` and `test-and-scan` 😃

marcussc (approved):

PR #1398, opened 2020-10-23
Fix/3311 app crash after qr scanning

KaiTeuber:
## Description
Fixed a problem that caused a crash when the QR Code Scanner read unexpected data

## Screenshot
Screen Shot 2020-10-23 at 09 57 18

## Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3311

PR #1407, opened 2020-10-26
Chore/Unit Test for Legal Texts

tcfos:
This PR contains two unit tests to ensure that the legal texts for PL, RO, and BG are equal to EN.

There is a **dependency** to https://github.com/corona-warn-app/cwa-app-ios/pull/1398. Please merge #1398 before #1407

No Jira item. No screenshots.


src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
```diff
@@ -0,0 +1,63 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import Foundation
+import XCTest
+@testable import ENA
+
+class AppInformationLegalTextsTest: XCTestCase {
```
ArturFriesen, line 23: I can't see the file in Xcode. Only in the file system.

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
```diff
@@ -0,0 +1,63 @@
...
+			XCTAssertEqual(targetText, sourceText)
+		}
+	}
+	
+	private func getText(fromFile: String, withExtension: String, localization: String) throws -> String {
+		let directory = localization + ".lproj"
+		guard let url = Bundle.main.url(forResource: fromFile, withExtension: withExtension) else {
+			return ""
+		}
+		let data = try Data(contentsOf: url.deletingLastPathComponent()
```
ArturFriesen, line 55: Why is removing the path components and adding a directory needed?
Wouldn't it be enough to just Data(contentsOf: url) ?

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
tcfos: Fixed. New file AppInformationLegalTextsTest.swift added to Xcode

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
tcfos: The tests iterates over all target languages, and it has to read the localized files from different directories.

tcfos:
SonarCloud check failed because of low coverage of AppInformationLegalTextsTest. But this is a unit test!

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
ArturFriesen: I would propose to use subdirectory api:

```
private func getText(fromFile: String, withExtension: String, localization: String) throws -> String {
 let directory = localization + ".lproj"
 guard let url = Bundle.main.url(forResource: fromFile, withExtension: withExtension, subdirectory: directory) else {
  return ""
 }
 let data = try Data(contentsOf: url)

 return String(data: data, encoding: .utf8) ?? ""
}
```

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationLegalTextsTest.swift
tcfos: Thanks for the feedback. Changed func getText() accordingly.

tcfos:
**Caution** – please merge #1398 first because of dependencies.

marcussc (approved):

tcfos:
PR cancelled due to dependencies. New PR will be created.

PR #1421, opened 2020-10-28
Fix/3230 Formatting Terms of Use (EN)

tcfos:
## Desription
This PR fixes some formatting issues in the EULA.

## Jira 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3230

## Screenshots

Terms of Use EN


src/xcode/ENA/ENA/Resources/Localization/en.lproj/usage.html
```diff
@@ -174,45 +174,19 @@
...
+		<p>4. The App does not protect you against infection</p>
+		<p>5. Retrieval of test results</p>
+		<p>6. Activating the warning feature</p>
+		<p>7. Requirements for using the App</p>
+		<p>8. Limitations of the App</p>
+		<p>9. Rights of use</p>
+		<p>10. Important information about availability and changes</p>
+		<p>11. No warranty</p>
+		<p>12. Special terms for the iOS version of the App</p>
+		<p>13. Final provisions</p>
```
whiskey, line 189: We really need a sophisticated solution here. Not only will, for example an `<ol>`, create cleaner semantics but also help with accessibility for screen readers!

whiskey (approved):
Approved. Just added my 0.02€.

src/xcode/ENA/ENA/Resources/Localization/en.lproj/usage.html
tcfos: Converting the MSWord document to html is a pain. There is still manual formatting required, and the CSS styles  are not working reliably.

janetback (approved):
UA reviewed

PR #1468, opened 2020-11-05
Feature/Task 3583 Download Hour Packages on WiFi Only

tcfos:
## Description
Hourly packages shall be downloaded on WiFi network only

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3583

## Screenshot
n/a


src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
```diff
@@ -33,7 +33,21 @@ extension URLSessionConfiguration {
 	}
 
 
-	/// Session config with caching policy `.useProtocolCachePolicy`
+	/// CWA Session config but only valid in WiFi environments
+	/// - Returns: the same as `.coronaWarnSessionConfiguration` but only for WiFi connections
+	class func coronaWarnSessionConfiguration_WiFi_only() -> URLSessionConfiguration {
```
marcussc, line 38:
```suggestion
	class func coronaWarnSessionConfigurationWifiOnly() -> URLSessionConfiguration {
```

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
tcfos: OK, done.

hansemannn:
Shouldn't this be up to the user? Either with a prompt or app setting?

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -128,7 +128,9 @@ final class HTTPClient: Client {
 		completion completeWith: @escaping HourCompletionHandler
 	) {
 		let url = configuration.diagnosisKeysURL(day: day, hour: hour, forCountry: country)
-
+		let config = URLSessionConfiguration.coronaWarnSessionConfigurationWifiOnly()
+		session.configuration.allowsConstrainedNetworkAccess = config.allowsConstrainedNetworkAccess
```
ArturFriesen, line 132: Please correct me if I'm wrong, but here you are changing the session settings of the HTTPClient.
This settings are never changed back.
In the next request (also a day request) the wifi settings will be used.
The HTTPClient is only initialized once.

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
```diff
@@ -33,7 +33,21 @@ extension URLSessionConfiguration {
...
+	/// CWA Session config but only valid in WiFi environments
+	/// - Returns: the same as `.coronaWarnSessionConfiguration` but only for WiFi connections
+	class func coronaWarnSessionConfigurationWifiOnly() -> URLSessionConfiguration {
+		let config = coronaWarnSessionConfiguration()
+		// no download over expensive network (cellular)
+		config.allowsExpensiveNetworkAccess = false
+		// no download in case user has activated the low data mode
+		config.allowsConstrainedNetworkAccess = false
+		// session shall not wait for connectivity to become available; fail immediately
+		config.waitsForConnectivity = false
```
ArturFriesen, line 45: I think this is not used and can be removed.

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
tcfos: Shall we reset the configuration after loading the hour packages?

tcfos:
> Shouldn't this be up to the user? Either with a prompt or app setting?

The day packages are essential for risk calculation and the download is not restricted to WiFi networks. The hour packages are not essential in that sense, and therefore the download is skipped in case of expensive networks or low data mode. 
Any user input here would complicate the situation unnecessarily and would required more documentation.

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
marcussc: Nice catch Artur.
IMHO we should have two different Sessions. This would allow us later to have the two downloads run in parallel and we don't have to reset anything after the hour packages were downloaded.

ArturFriesen: Hmm I'm not sure if this is a good solution.
Because the requests are dispatched async and another request (maybe a day request) could be dispatched (with the wrong session config) while we are waiting for the hour request to come back.

ArturFriesen: Solutions could be: 

- We use 2 separate sessions.
- We reconfigure the session for every request.

@marcussc  Any other suggestions?

marcussc: @ArturFriesen 
Nope but I would prefer 2 separate sessions for the reasons you stated. :)

whiskey: I put this already on my list. The Client is currently very inflexible. We already have a dedicated `CachingHTTPClient` to handle similar issues in other places. Maybe we create another WiFiOnlyClient for the moment and refactor them altogether?

src/xcode/ENA/ENA/Source/Extensions/URLSessionConfiguration+Default.swift
whiskey: Even `allowsConstrainedNetworkAccess`?

src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -167,6 +168,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 	let downloadedPackagesStore: DownloadedPackagesStore = DownloadedPackagesSQLLiteStore(fileName: "packages")
 
 	let client: HTTPClient
+	let wifiClient: HTTPClientWifiOnly
```
ArturFriesen, line 171: To stay consistent I would propose the name: WifiOnlyHTTPClient. 
Its like CachingHTTPClient.

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/AppDelegate.swift
KaiTeuber: good point - I changed it

marcussc (approved):

PR #1479, opened 2020-11-09
Feature/task 3613 fix downloading detecting state

ArturFriesen:
## Description
This PR fixes the downloading and detecting state on the home screen.

[3613](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3613)


src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
```diff
@@ -290,6 +311,7 @@ class KeyPackageDownload: KeyPackageDownloadProtocol {
 		switch downloadMode {
 		case .daily:
 			client.availableDays(forCountry: country) { result in
+				print("availableServerData: \(result)")
```
whiskey, line 314: Please use proper logging or remove `print`

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
```diff
@@ -33,7 +33,7 @@ final class RiskConsumer: NSObject {
 	var didCalculateRisk: ((Risk) -> Void) = { _ in }
 
 	/// Called when the risk calculation fails.
-	var didFailCalculateRisk: (RiskCalculationError) -> Void = { _ in }
+	var didFailCalculateRisk: (RiskProviderError) -> Void = { _ in }
```
whiskey, line 36: can we please get rid of this anti-pattern and `throw` errors!?

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
```diff
@@ -22,7 +22,6 @@ import Foundation
 final class ExposureDetection {
 
 	// MARK: Properties
-	@Published var activityState: RiskProvider.ActivityState = .idle
```
whiskey, line 25: Sad to see this being removed…

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -131,54 +133,12 @@ extension RiskProvider: RiskProviding {
...
-			if !enoughTimeHasPassed || !self.exposureManagerState.isGood || !shouldDetectExposures {
-				if let summary = store.summary {
-					completion(.success(summary))
-				} else {
-					completion(.failure(.missingCachedSummary))
-				}
-				return
+			// Erase optionality.
+			let completion: Completion = { result in
+				completion?(result)
```
whiskey, line 138: What's happening here?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -456,46 +483,57 @@ extension RiskProvider: RiskProviding {
...
 #if DEBUG
 extension RiskProvider {
 	private func _requestRiskLevel_Mock(userInitiated: Bool, completion: Completion? = nil) {
-		let risk = Risk.mocked
-
-		targetQueue.async {
-			completion?(.success(.mocked))
+		// Erase optionality.
+		let completion: Completion = { result in
+			completion?(result)
```
whiskey, line 531: Why leave the completion handler then optional?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
ArturFriesen: Its just a renaming.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: Resolving optionality for the completion closure for the rest of the code.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: Why not? It can be used or not by the other parts of the app.
There is a scenario where you can request risk and the result comes from the consumer.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetection.swift
ArturFriesen: Why are you sad? The ExposureDetection has no different states anymore. Why do we need it?

whiskey: Must have been confused with the `.idle` in `ExposureDetectionViewController+State` then.

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
ArturFriesen: removed

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
KaiTeuber: throw errors / current used (anti) pattern might be a good discussion for dev camp?

ArturFriesen: In the RiskConsumer we can use Result-Type. Don't see a possibility to use throws in RiskConsumer because its "async".

whiskey: We've discussed that. A potential fix for a future refactoring might be a `Result`.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
whiskey: Already bad enough to get results on two possible ways but how about turning the completion block of `_requestRiskLevel` and `calculateRiskLevel` into optionals then? Also, the `_requestRiskLevel_Mock` has an optional completion handler. This looks inconsistent.

Suggestion at the corresponding place:
`private func _requestRiskLevel(userInitiated: Bool, ignoreCachedSummary: Bool, completion: Completion? = nil)`

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
whiskey: Merci. Should we add a swiftlint rule for this? (cc @marcussc) 
I might have one laying around from an old project…

marcussc: Nice idea 👍
But please in an extra PR :)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Model/RiskConsumer.swift
KaiTeuber: > We've discussed that. A potential fix for a future refactoring might be a `Result`.

Unfortunately I can't remember it - but maybe I was simply not there?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: Getting results over 2 possible ways is not something I have added in this PR. It was already there and I also don't like it. Needs to be refactored.

"but how about turning the completion block of _requestRiskLevel and calculateRiskLevel into optionals then?"
Now we are getting to the point. Thats exactly what I want to prevent with the wrapped completion.

I don't think, that wrapping the completion to erase optionality is a "bad" thing.
But I can understand, that its kind of hard to understand.
I will change every completion to optional again.

marcussc (approved):

whiskey (approved):

PR #1486, opened 2020-11-10
Feature/3627 invalidate cached files

whiskey:
As part of the 'update risk calculation more often' story, this PR checks for the ETags given in the app configuration's

* iosKeyDownloadParameters.cachedDayPackagesToUpdateOnEtagMismatch
* iosKeyDownloadParameters.cachedHourPackagesToUpdateOnEtagMismatch

If ETags are given, they will be matched with the list of currently stored key packages and eventually removed.

This solves Jira ticket [EXPOSUREAPP-3627](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3627)


src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
```diff
@@ -188,7 +188,7 @@ final class DownloadedPackagesSQLLiteStoreTests: XCTestCase {
 
 		let countries = ["DE", "IT"]
 		let days = ["2020-11-03", "2020-11-02"]
-		let hours = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 141, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
+		let hours = [Int].init(1...24)
```
tcfos, line 191: 👍

tcfos (approved):

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/PackageKeyStore+CacheInvalidation.swift
```diff
@@ -0,0 +1,33 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+import Foundation
+
+
+extension SAP_Internal_ApplicationConfiguration {
```
ArturFriesen, line 23: I would propose the naming "ApplicationConfiguration+CacheInvalidation.swift" instead of "PackageKeyStore+CacheInvalidation.swift"

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V2/DownloadedPackagesStoreV2.swift
```diff
@@ -56,6 +56,15 @@ protocol DownloadedPackagesStoreV2: AnyObject {
 	/// - Throws: An error of type `SQLiteStoreError`
 	func delete(packages: [SAPDownloadedPackage]) throws
 
+	// MARK: - Package Invalidation
+
+	/// A list of invalid ETags to be removed/ignored. Defined in the AppConfiguration.
+	var revokationList: [String] { get set }
```
ArturFriesen, line 62: Why is is the revokationList part of the PackagesStore?
Shouldn't it be enough to just clean them from cache every time we get them?

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/Doubles/DownloadedPackagesStoreErrorStub.swift
```diff
@@ -76,4 +76,6 @@ class DownloadedPackagesStoreErrorStub: DownloadedPackagesStoreV2 {
 
 	var keyValueStore: Store?
 
+	var revokationList: [String] = []
```
ArturFriesen, line 79: I don't see that .revokationList is ever written besides of the tests.
Can we write the test a different way to prevent this list?

ArturFriesen (changes requested):
"revokationList" only used in tests. Maybe not needed.

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/Doubles/DownloadedPackagesStoreErrorStub.swift
whiskey: I thought it was also written in the real app on fetching a new app config. The idea was to prevent storing invalid packages. That is an edge case but covers a case where the CDN still provides invalid packages that have already been flagged in the app config.
I'll update the code.

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/PackageKeyStore+CacheInvalidation.swift
whiskey: The initial code contained more protocols but was changed due to implementation details. Will change this.

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V2/DownloadedPackagesStoreV2.swift
whiskey: Checking just on AppConfig update leaves a gap in which packages might arrive that are on the 'bad list' . Keeping this list at hand for future save (`set(:)`) operations will cover cases where a download happens after the cache validation. 

I have to admit, it's an edge case but by the price of this property we cover all states regardless the order of fetch and update app config.

PR #1489, opened 2020-11-10
Feature/task 3667 dm disable hourly packages download

KaiTeuber:
## Description
Added switch to simulate the behavior for downloading the hourly keys without Wifi

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3667

## Screenshot
Simulator Screen Shot - iPhone 11 - 2020-11-10 at 16 02 35
Simulator Screen Shot - iPhone 11 - 2020-11-10 at 16 02 05


whiskey (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/WifiOnlyHTTPClient.swift
```diff
@@ -60,6 +61,12 @@ final class WifiOnlyHTTPClient: ClientWifiOnly {
 			}
 		}
 
+		#if !RELEASE
+		guard !disableHourlyDownload else {
+			return
```
marcussc, line 66: Don't we have to call the completion with the same error as if we are connected over cellular and `allowsCellularAccess` is set to `false`?

KaiTeuber: exactly, the defer paragraph farther up makes that

marcussc: Hm but `responseError` never gets set and then no completion is called?

marcussc (approved):

PR #1506, opened 2020-11-13
Fix/3711 refactor cached app config

ArturFriesen:
## Description
This PR prevents the race condition which is leading to trigger risk detection more then the timeframe (24h) allows.
The meta data and the cached app config are persisted at once, preventing data collisions when the cache called multiple times. 

## Link to Jira
links follows. Cant access jira atm.


src/xcode/ENA/ENA/Source/Client/__tests__/Mocks/CachingHTTPClientMock.swift
```diff
@@ -42,6 +42,22 @@ final class CachingHTTPClientMock: CachingHTTPClient {
 		return config
 	}()
 
+	static let staticAppConfigMetadata: AppConfigMetadata = {
+		let bundle = Bundle(for: CachingHTTPClientMock.self)
+		// there is a test for this (`testStaticAppConfiguration`), let's keep it short.
+		guard
+			let fixtureUrl = bundle.url(forResource: "de-config-int-2020-09-25", withExtension: nil),
```
whiskey, line 49: Just FYI, the app configuration format will change in the next two sprints which will likely break the deserialization. However, with the default app configuration on the horizon we might have this covered!

Unless we have a need for a bespoke config, we could use the static default configuration, right?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -47,63 +47,75 @@ final class CachedAppConfiguration {
...
 				// update revokation list
-				let revokationList = self.store.appConfig?.revokationEtags ?? []
+				let revokationList = self.store.appConfigMetadata?.appConfig.revokationEtags ?? []
 				self.packageStore?.revokationList = revokationList // for future package-operations
 				// validate currently stored key packages
 				do {
 					try self.packageStore?.validateCachedKeyPackages(revokationList: revokationList)
 				} catch {
-					Log.error("Error while removing invalidated key packages.", log: .localData, error: error)
+					Log.error("[App Config] Error while removing invalidated key packages.", log: .localData, error: error)
```
whiskey, line 96: We could create custom OSLog keys for `.cache` and `.appConfig`!? This would allow nicer filtering for the different components.

whiskey (approved):

src/xcode/ENA/ENA/Source/Client/__tests__/Mocks/CachingHTTPClientMock.swift
ArturFriesen: Good point. Made use if CachingHTTPClientMock.staticAppConfig

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
ArturFriesen: Added .appConfig and used it.

marcussc (approved):

PR #1510, opened 2020-11-13
Fix/dont show allready running alert

ArturFriesen:
## Description
This PR prevents the alert for already running error.

## Link to Jira
[3769](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3769)


marcussc:
Labels please :)

marcussc (approved):

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -97,7 +97,12 @@ extension AppDelegate {
 				secondaryActionTitle: AppStrings.Common.errorAlertActionMoreInfo,
 				secondaryActionCompletion: openFAQ
 			)
-		default:
+		} else if let exposureDetectionError = noSummaryError as? ExposureDetectionError {
+			switch exposureDetectionError {
+			case .isAlreadyRunning:
+				return nil
```
nickguendling, line 103: Doesn't the error also trigger a different cell on the home view? Should we throw this error at all?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/AppDelegate.swift
ArturFriesen: Very good point! 
Currently it cannot happen, because RiskProvider "guards" this case for the key download and the risk detection.
But I think it is still a valid point from the perspective of code readability.
I have added error handling to HomeInteractor which explicitly ignores the allready running errors.

nickguendling (approved):

PR #1511, opened 2020-11-13
Fix/3775 No increased frequency for exposure checks visible though new hourly packages available

marcussc:
## Description
- Adds a Switch in the DM to disable Device-Time-Check.
- Fixes an Issue where Device-Time-Check is not done if the Etag of the AppConfig does not change
- Fixes an Edge-Case in manual mode where the user gets requested to update his risk-score but nothing happens.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3775

## Screenshot
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-13 at 16 49 29

### DIFF
--- a/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewController.swift
@@ -1,0 +1,59 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+#if !RELEASE
+
+import UIKit
+
+final class DMDeviceTimeCheckViewController: UITableViewController {
+
+	// MARK: - Init
+
+	init(store: Store) {
+		self.viewModel = DMDeviceTimeCheckViewModel(store: store)
+		super.init(style: .insetGrouped)
+	}
+
+	required init?(coder: NSCoder) {
+		fatalError("not supported")
+	}
+
+	// MARK: - Overrides
+
+	override func viewDidLoad() {
+		super.viewDidLoad()
+		setupTableView()
+		title = "Device Time Check 📱 🕰"
+	}
+
+	// MARK: - Protocol UITableViewDataSource
+
+	override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+		return viewModel.itemsCount
+	}
+
+	override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+		guard let cell = tableView.dequeueReusableCell(withIdentifier: "DMSwitchTableViewCell") as? DMSwitchTableViewCell else {
+			let dummy = UITableViewCell(style: .default, reuseIdentifier: "DummyFallBackCell")
+			dummy.textLabel?.text = "Fallback cell"
+			return dummy
+		}
+		cell.configure(cellViewModel: viewModel.cellViewModel(for: indexPath))
+		return cell
+	}
+
+	// MARK: - Public
+
+	// MARK: - Internal
+
+	// MARK: - Private
+
+	private let viewModel: DMDeviceTimeCheckViewModel
+
+	private func setupTableView() {
+		tableView.register(UINib(nibName: "DMSwitchTableViewCell", bundle: nil), forCellReuseIdentifier: "DMSwitchTableViewCell")
+	}
+}
+
+#endif
--- a/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewModel.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewModel.swift
@@ -1,0 +1,56 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+#if !RELEASE
+
+import Foundation
+import UIKit
+
+final class DMDeviceTimeCheckViewModel {
+
+	// MARK: - Init
+
+	init(store: Store) {
+		self.store = store
+	}
+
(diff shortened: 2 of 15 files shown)


src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+		guard let item = menuItems(rawValue: indexPath.row) else {
+			fatalError("failed to create cellViewModel")
+		}
+		switch item {
+
+		case .killDeviceTimeCheck:
+			return DMSwitchCellViewModel(
+				labelText: "Disable device time check",
+				isEnabled: { [store] in
+					store.dmKillDeviceTimeCheck ?? false
```
ArturFriesen, line 38: Will the button be disabled if the store.dmKillDeviceTimeCheck is nil? I think the store.dmKillDeviceTimeCheck is nil initially and the button will never be enabled? Or did I miss something?

ArturFriesen: Maybe it makes sense to put default "false" into the store property?

marcussc: The `isEnabled` refers to the state of the Switch, is it switched on or ist it switched off.

ArturFriesen: Ok.. then the name is misleading. Should be "isOn"

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -252,6 +252,11 @@ final class SecureStore: Store {
 		get { kvStore["fakeSQLiteError"] as Int32? }
 		set { kvStore["fakeSQLiteError"] = newValue }
 	}
+	
+	var dmKillDeviceTimeCheck: Bool? {
```
ArturFriesen, line 256: Maybe it makes sense to put default "false" into the store property?

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+		guard let cell = tableView.dequeueReusableCell(withIdentifier: "DMSwitchTableViewCell") as? DMSwitchTableViewCell else {
+			let dummy = UITableViewCell(style: .default, reuseIdentifier: "DummyFallBackCell")
+			dummy.textLabel?.text = "Fallback cell"
+			return dummy
+		}
+		cell.configure(cellViewModel: viewModel.cellViewModel(for: indexPath))
+		return cell
+	}
+
+	// MARK: - Public
```
SwissMountainFrog, line 46: Shouldn't we remove the unused ones?

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+
+	init(store: Store) {
+		self.store = store
+	}
+
+	// MARK: - Overrides
+
+	// MARK: - Public
+
+	// MARK: - Internal
```
SwissMountainFrog, line 22: Same MARK question as before

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
marcussc: Didn't won't to write something into the store that only gets set if you want to debug something. Maybe I should also set it back to nil if the Toogle gets set to Off again.

ArturFriesen: You wouldn't write it to the store. The default would come into place if the flag is read but nothing was written to the store before. At the moment you are handling the nil case with default of false at different places instead of a central one.

For example isAllowedToPerformBackgroundFakeRequests:
```
	var isAllowedToPerformBackgroundFakeRequests: Bool {
		get { kvStore["shouldPerformBackgroundFakeRequests"] as Bool? ?? false }
		set { kvStore["shouldPerformBackgroundFakeRequests"] = newValue }
	}
```

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewModel.swift
marcussc: Done ✅

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
marcussc: Ah sweet, I don't know that.
Refactored :)

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDeviceTimeCheck/DMDeviceTimeCheckViewController.swift
marcussc: Hm feels like we don't have no clear decision yet. I personally like to keep them in for future me, if I have to add something in an currently empty Mark

PR #1516, opened 2020-11-16
Fix/3773 Show unkown risk cell when the risk is not available.

ArturFriesen:
## Description
This PR fixes the missing download and detection state on the home screen during the initial launch of the app.

If the App is launched initially there is no Risk calculated.
To show the download and detection state there needs to be a risk card.
This PR adds the unknown initial risk card initially to make the progress visible.

Additionale there could be allready a risk detection running when the home screen is initialized.
For this case the PR adds an update to the init.

## Link to Jira
[3773](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3773)


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -20,6 +20,8 @@ final class HomeInteractor: RequiresAppDependencies {
 		self.homeViewController = homeViewController
 		self.state = state
 		self.exposureSubmissionService = exposureSubmissionService
+
+		updateAndReloadRiskCellState(to: riskProvider.activityState)
 		observeRisk()
```
KaiTeuber, line 25: I wonder if observing the risk after starting updateAndReloading might lead to a race condition?
Isn't it better to first observerRisk() and after that trigger the update?

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
ArturFriesen: I changed it to setting just the state instead of reloading the cell.
The cell gets loaded anyways after init.

PR #1519, opened 2020-11-16
Fix/3806 no key download between detections

ArturFriesen:
## Description
This PR fixes the date formatter for hour packages and adds a lot of logging.

## Link to Jira
[3806](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3806)


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
```diff
@@ -326,8 +334,8 @@ class KeyPackageDownload: KeyPackageDownloadProtocol {
 			case .daily:
 				downloadedPackagesStore.deleteDayPackage(for: package, country: countryId)
 			case .hourly(let keyDay):
-				// hourly packages for a day are deleted when the day package is stored. See func
-				// DownloadedPackagesSQLLiteStoreV1.set(  country: Country.ID,	day: String, package: SAPDownloadedPackage )
+				// Hourly packages for the last day are deleted when the last day package is stored. See func
+				// DownloadedPackagesSQLLiteStore.set(country: Country.ID, day: String, package: SAPDownloadedPackage)
```
tcfos, line 338: Just to be a bit more precise: func DownloadedPackagesSQLLiteStoreV2.set(country: Country.ID, day: String, etag: String?, package: SAPDownloadedPackage) deletes the hour packages for the respective day. This works for all days, not only the last one.

tcfos (approved):

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
ArturFriesen: Changed the comment to: 
"// Hourly packages for a day are deleted when the day package is stored. See func
// DownloadedPackagesSQLLiteStore.set(country: Country.ID, day: String, package: SAPDownloadedPackage)"

I would like to use the macro DownloadedPackagesSQLLiteStore, because the versioned name would get out of date.

PR #1523, opened 2020-11-17
Protocol Buffer update for release 1.9.x (EXPOSUREAPP-3411)

Corona-Warn-App-Technical-User:


mlenkeit:
Protobuf is added so that the CWA Server team can already start. I converted this to draft, as the client activities are not 100% ready.

proto/resources/internal/stats/key_figure_card.proto
```diff
@@ -0,0 +1,42 @@
...
+  enum Trend {
+    NONE = 0;
+    STABLE = 1;
+    INCREASING = 2;
+    DECREASING = 3;
+  }
+
+  enum TrendSemantic {
+    NONE = 0;
+    NEUTRAL = 1;
```
whiskey, line 32: What's the difference?

proto/resources/internal/stats/key_figure_card.proto
```diff
@@ -14,21 +14,21 @@ message KeyFigureCard {
...
-    NONE = 0;
+    UNSPECIFIED_TREND = 0;
     STABLE = 1;
     INCREASING = 2;
     DECREASING = 3;
   }
 
   enum TrendSemantic {
-    NONE = 0;
+    UNSPECIFIED_TREND_SEMANTIC = 0;
```
whiskey, line 31: No need to suffix the value here, as it has context from the enum name.
``` suggestion
    UNSPECIFIED = 0;
``

mlenkeit: They actually do have to be unique. That's super annoying, otherwise it doesn't compile:

Error message:
> Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "NONE" must be unique within "app.coronawarn.server.common.protocols.internal.stats", not just within "TrendSemantic".


see extensive commit message https://github.com/corona-warn-app/cwa-protocol-buffers/commit/24233fcb351668f1cdbecf972ada616acd22bb7d

whiskey: 🙄  oh well, this is so 2000…
Anyways, thanks for the hint :)

PR #1524, opened 2020-11-17
Fix/3809 No reminder to share a positive test result if user first register QR code before test result positive

SwissMountainFrog:
This PR fixes an issue if the test result changes while the user has the app in the background (so the test result was `pending` and changed to `positive` on the server.

Also, this PR fixes an issue with "updating" an existing `pending` result and the updated state is `positive`.

Now in both situations, the two "warn others reminder" get scheduled as expected.

The PR is [based on this JIRA Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3809)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
@@ -125,7 +125,7 @@ class ExposureSubmissionTestResultViewModel {
 	}
 	
 	func updateWarnOthers() {
-		self.warnOthersReminder.evaluateNotificationState(testResult: self.testResult)
+		warnOthersReminder.evaluateNotificationState(testResult: testResult)
 	}
 	
 	private func updateButtons() {
@@ -163,6 +163,7 @@ class ExposureSubmissionTestResultViewModel {
 				self?.error = error
 			case let .success(testResult):
 				self?.testResult = testResult
+				self?.updateWarnOthers()
 			}
 
 			completion()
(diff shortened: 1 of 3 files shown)


nickguendling (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
```diff
@@ -125,7 +125,7 @@ class ExposureSubmissionTestResultViewModel {
 	}
 	
 	func updateWarnOthers() {
-		self.warnOthersReminder.evaluateNotificationState(testResult: self.testResult)
+		warnOthersReminder.evaluateNotificationState(testResult: testResult)
 	}
```
KaiTeuber, line 129: inside the prive section - make it private too or move?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
```diff
@@ -163,6 +163,7 @@ class ExposureSubmissionTestResultViewModel {
 				self?.error = error
 			case let .success(testResult):
 				self?.testResult = testResult
+				self?.updateWarnOthers()
```
KaiTeuber, line 166: if this is the only caller why use a separte func?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
SwissMountainFrog: Moved. Thanks.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultViewModel.swift
SwissMountainFrog: The reason is to keep it logically separated

KaiTeuber (approved):

PR #1541, opened 2020-11-18
Feature/3780 exposure window risk calculation v2 unit tests

KaiTeuber:
## Description
unit tests added for new risk calculation

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3780


tcfos (approved):

src/xcode/ENA/ENA/Source/View Helpers/__tests__/CountdownTimerTests.swift
```diff
@@ -11,6 +11,7 @@ class CountdownTimerTests: XCTestCase {
 	private var countdownTimerTarget: CountdownTimerTarget!
 
 	override func setUp() {
+		super.setUp()
```
whiskey, line 14: I was wondering if this is required. It is :)
https://developer.apple.com/forums/thread/113314?answerId=348312022#348312022

src/xcode/ENA/ENATests/Helper/TimeInterval+Convenience.swift
```diff
@@ -14,11 +14,13 @@ extension TimeInterval {
 		self = Double(days * 24 * 60 * 60)
 	}
 
-	/// One second
+	/// two seconds
```
whiskey, line 17: I like the increase. 👍
Just for discussions sake, why just 2,5 and 10? and not 10, 20, 30? These numbers seem to be arbitrary 🤷‍♂️ If something is time-critical we should consider performance tests and if not, a general timeout of n seconds (for synchronous tasks) would do the trick.

whiskey (approved):
Nothing to complain, just a few thoughts for further discussions.

nickguendling (approved):

PR #1555, opened 2020-11-23
Describe origin of public keys

mlenkeit:
This PR adds a comment to `PublicKeyStore.swift` describing the format of the hard-coded public keys.

Reason for this is that we were wondering why they are different than on Android. To answer this, we had to reach out to Engineers from the early phase of the project. That's why I though it would might be a good idea to add the finding as a comment for our future selves and and other Engineers 😅


marcussc (approved):

whiskey (approved):
Good note

PR #1564, opened 2020-11-24
TASK 3761 Changed Illustration on Screen "Vielen Dank"

tcfos:
## Description
Replaced the illustration on the "Vielen Dank" screen. Adjusted the layout – the illustration stretches out from left to right over the complete width of the card.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3761

### 1
Vielen Dank Home 1

### 2
Vielen Dank Home 2

### 3
Vielen Dank Home 3


nickguendling (commented):
The Illustration does not go all the way to the edges of the card as requested by the new design. Is that something we can easily fix?

whiskey (approved):

tcfos:
> The Illustration does not go all the way to the edges of the card as requested by the new design. Is that something we can easily fix?

Alignment has been fixed accordingly.

src/xcode/ENA/ENA/Source/Views/Home Screen/Cells/Risk/RiskThankYouCollectionViewCell.xib
```diff
@@ -66,9 +85,26 @@
                                         </userDefinedRuntimeAttributes>
                                     </label>
                                     <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" translatesAutoresizingMaskIntoConstraints="NO" id="BEI-EV-nwa">
-                                        <rect key="frame" x="0.0" y="296" width="341" height="173"/>
+                                        <rect key="frame" x="16" y="412" width="396" height="57"/>
+                                        <constraints>
+                                            <constraint firstAttribute="width" constant="396" id="0DF-hU-67w"/>
```
nickguendling, line 90: Please don't set fixed `width` constraints on all these. We support multiple screen sizes and get auto layout warnings because of these. Your solution should work without them

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Views/Home Screen/Cells/Risk/RiskThankYouCollectionViewCell.xib
KaiTeuber: I had a shot discussion with Andreas about that and we came to a solution with a proportional width constraint used

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Views/Home Screen/Cells/Risk/RiskThankYouCollectionViewCell.xib
nickguendling: I tried it out before I commented, there already were trailing and leading constraints, you just would have needed to delete the redundant width constraints, that's all. Would be a cleaner solution than involving the superview's width as it is now.

tcfos: removed the fixed width. No more layout warnings at runtime.

tcfos: Discussed with Nick; removed the width, added constraints for horizontal layout.

nickguendling (approved):

PR #1575, opened 2020-11-25
Fix/fix for leave crash

ArturFriesen:
## Description
This PR fixes a possibility for the leave crash to occur.
The `Set` in `RiskProvider` prevents that the same consumer is registered more than once and then called more than once.
Possible multiple calls to the `RiskConsumer` lead to multiple calls of `.leave()` in `ENATaskExecutionDelegate`.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
@@ -53,11 +53,11 @@ final class RiskProvider: RiskProviding {
 	}
 
 	func observeRisk(_ consumer: RiskConsumer) {
-		consumers.append(consumer)
+		consumers.insert(consumer)
 	}
 
 	func removeRisk(_ consumer: RiskConsumer) {
-		consumers.removeAll(where: { $0 === consumer })
+		consumers.remove(consumer)
 	}
 
 	/// Called by consumers to request the risk level. This method triggers the risk level process.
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENA/Source/Services/Risk/Provider/__tests__/RiskProviderTests.swift
```diff
@@ -407,9 +407,82 @@ final class RiskProviderTests: XCTestCase {
...
+		}
+
+		let appDelegate = AppDelegate()
+		appDelegate.riskProvider = riskProvider
+
+		for _ in 0...numberOfExecuteENABackgroundTask - 1 {
+			appDelegate.executeENABackgroundTask { _ in }
+		}
+
+		waitForExpectations(timeout: 20)
```
marcussc, line 451: Can we use something like `.extraLong` here?

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -101,8 +101,8 @@ final class RiskProvider: RiskProviding {
 
 	private var subscriptions = [AnyCancellable]()
 	
-	private var _consumers: [RiskConsumer] = []
-	private var consumers: [RiskConsumer] {
+	private var _consumers: Set<RiskConsumer> = Set<RiskConsumer>()
+	private var consumers: Set<RiskConsumer> {
```
KaiTeuber, line 105: good idea to better use a set here

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/__tests__/RiskProviderTests.swift
ArturFriesen: changed

PR #1591, opened 2020-11-28
fix/removed configuration inside exposure windows inside debug menu

KaiTeuber:
## Description
configuration was exported on every ExposureWindow, what made the JSON long
Removed the configuration inside each exposure window

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4052


src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
```diff
@@ -101,12 +86,14 @@ class DMDebugRiskCalculationViewController: UIViewController {
 		encoder.dateEncodingStrategy = .iso8601
 
 		textView.text += "\n\nConfiguration:\n\n"
-		if let data = try? encoder.encode(mostRecentRiskCalculationConfiguration), let mostRecentRiskCalculationConfiguration = String(data: data, encoding: .utf8) {
+		if let data = try? encoder.encode(mostRecentRiskCalculationConfiguration),
```
whiskey, line 89: We should consider proper error handling, even for dev code. This might bring up potential de-/serialization issues earlier.

whiskey (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/RiskCalculationExposureWindow.swift
```diff
@@ -49,11 +49,11 @@ final class RiskCalculationExposureWindow: Codable {
 		self.init(exposureWindow: exposureWindow, configuration: configuration)
 	}
 
+	// encode data to json - used for debugging help
 	func encode(to encoder: Encoder) throws {
 		var container = encoder.container(keyedBy: CodingKeys.self)
 
 		try container.encode(exposureWindow, forKey: .exposureWindow)
-		try container.encode(configuration, forKey: .configuration)
```
nickguendling, line 56: I tried this before. If we don't encode the configuration, the decoding fails because of the missing configuration and we're not able to get the most recent risk calculation from the store. It always returns `nil`. I couldn't come up with a quick solution last time, that's why the configuration is still encoded with every window.

nickguendling (changes requested):

KaiTeuber:
After Nick pointed out, that we can't remove the configuration that easy I tried some different approaches. But unfortunately it's not that easy, so we decided to combine configuration & most recent risk calculation into a more export friendly formt

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
```diff
@@ -82,32 +67,26 @@ class DMDebugRiskCalculationViewController: UIViewController {
 			view.bottomAnchor.constraint(equalTo: textView.bottomAnchor)
 		])
 
-		guard let mostRecentRiskCalculationConfiguration = store.mostRecentRiskCalculationConfiguration,
-			  let mostRecentRiskCalculation = store.mostRecentRiskCalculation else {
-			textView.text = "No risk calculation run yet."
+		guard let mostRecentRiskCalculationConfiguration = store.mostRecentRiskCalculationConfiguration else {
+			textView.text = "missing configuration - stop here"
```
nickguendling, line 71: Why not leave the old guard here that checks for both values and otherwise shows "No risk calculation run yet."?

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
```diff
@@ -119,4 +98,9 @@ class DMDebugRiskCalculationViewController: UIViewController {
 
 }
 
+struct  RiscCalculationDebugHelper: Encodable {
```
nickguendling, line 101:
```suggestion
struct  RiskCalculationDebugHelper: Encodable {
```

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
```diff
@@ -82,32 +67,26 @@ class DMDebugRiskCalculationViewController: UIViewController {
...
-		guard let mostRecentRiskCalculationConfiguration = store.mostRecentRiskCalculationConfiguration,
-			  let mostRecentRiskCalculation = store.mostRecentRiskCalculation else {
-			textView.text = "No risk calculation run yet."
+		guard let mostRecentRiskCalculationConfiguration = store.mostRecentRiskCalculationConfiguration else {
+			textView.text = "missing configuration - stop here"
 			return
 		}
 
-		textView.text = "Most recent risk calculation: "
+		let riscCalculationDebugHelper = RiscCalculationDebugHelper(
```
nickguendling, line 75:
```suggestion
		let riskCalculationDebugHelper = RiskCalculationDebugHelper(
```

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
nickguendling: Then `mostRecentRiskCalculation` in `RiskCalculationDebugHelper` wouldn't have to be optional, they are always both there or none of them is anyway.

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMDebugRiskCalculationViewController.swift
```diff
@@ -82,32 +67,26 @@ class DMDebugRiskCalculationViewController: UIViewController {
...
+			configuration: mostRecentRiskCalculationConfiguration,
+			mostRecentRiskCalculation: store.mostRecentRiskCalculation)
 
 		let dateFormatter = DateFormatter()
 		dateFormatter.dateStyle = .short
 		dateFormatter.timeStyle = .short
```
nickguendling, line 81: I guess this is not needed anymore

nickguendling (changes requested):

nickguendling (approved):

PR #1594, opened 2020-11-29
force app config update in onboarding

whiskey:
## Description
The (cached) app configuration was not 'reset' on app reset. This PR fixes this by force-loading the app configuration on first usage in onboarding. The configuration is loaded, regardless the current cache state.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3978


src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -145,7 +145,7 @@ final class OnboardingInfoViewController: UIViewController {
 	}
 
 	private func loadCountryList() {
-		appConfigurationProvider.appConfiguration().sink { [weak self] configuration in
+		appConfigurationProvider.appConfiguration(forceFetch: true).sink { [weak self] configuration in
```
nickguendling, line 148: Can we add a comment here on why we need to force fetch?

nickguendling (approved):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
whiskey: Sure. I'd like to get a step further and find (later™) a way to make sure that the first access to the app config is ensured to be forced. Currently this bug will reappear if we add something like `appConfigurationProvider.appConfiguration()` before this part of the code is called.

For time reasons I didn't check this in depth.

marcussc: Could you be so kind a create a refactor story for this, so we don't loose track of this issue? :)

whiskey: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4060

marcussc (approved):

PR #1618, opened 2020-12-02
Fix/4100 fix 24 hour countdown

ArturFriesen:
## Description
This PR fixes the countdown on the home screen.
The Problem occurred when the app was started without internet connection and the app stuck to the default config for 5 minutes. Resulting in showing the 24 hour countdown.

The fix changes the logic of the cached app config the way, that following criteria is fulfilled:

- Don't overwrite a previously from server fetched config with the default config. 
- Don't stick to the default config or a cached config for 5 minutes.

## Link to Jira
[4100](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4100)


src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -55,19 +55,10 @@ final class CachedAppConfiguration {
...
-		getAppConfig(with: store.appConfigMetadata?.lastAppConfigETag)
-			.sink { response in
-				self.store.appConfigMetadata = AppConfigMetadata(
-					lastAppConfigETag: response.etag ?? "\"ReloadMe\"",
-					lastAppConfigFetch: Date(),
-					appConfig: response.config
-				)
-			}
-			.store(in: &subscriptions)
+		getAppConfig(with: store.appConfigMetadata?.lastAppConfigETag).sink(receiveValue: { _ in }).store(in: &subscriptions)
```
marcussc, line 61: Is this still needed if we don't do anything with the result?

ArturFriesen: Yes, if the request succeeds we are saving it to the store.

ArturFriesen: Inside of getAppConfig result handling

nickguendling: Then maybe you could mark `getAppConfig()` with `@discardableResult` and remove `.sink(receiveValue: { _ in }).store(in: &subscriptions)` from this line?

ArturFriesen: As far as I know, .sink needs to be called on the Future to execute the fetch.

nickguendling: Makes sense 👍

marcussc (approved):

nickguendling (approved):

PR #1628, opened 2020-12-04
Fix/4107 key package revocation

ArturFriesen:
## Description
This PR fixes the handling of the etag header by using `.value(forHTTPHeaderField:` which handles the header case insensitive.
Furthermore the PR fixes an issue where packages where not downloaded after they where revoked and deleted.

## Link to Jira
[4107](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4107)


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V2/DownloadedPackagesStoreV2.swift
```diff
@@ -65,31 +65,50 @@ protocol DownloadedPackagesStoreV2: AnyObject {
...
 		try dayPackages.forEach { day, bucket in
-			try self.set(country: country, day: day, etag: bucket.etag, package: bucket.package)
+			do {
+				try self.set(country: country, day: day, etag: bucket.etag, package: bucket.package)
+			} catch DownloadedPackagesSQLLiteStore.StoreError.revokedPackage {
+				revokedPackageError = DownloadedPackagesSQLLiteStore.StoreError.revokedPackage
+			}
+		}
+		if let revokedPackageError = revokedPackageError {
+			throw revokedPackageError
```
KaiTeuber, line 83: I think we might make this a little bit more compact by:
```
	func addFetchedDays(_ dayPackages: [String: PackageDownloadResponse], country: Country.ID) throws {
		try dayPackages.forEach { day, bucket in
			do {
				try self.set(country: country, day: day, etag: bucket.etag, package: bucket.package)
			} catch DownloadedPackagesSQLLiteStore.StoreError.revokedPackage {
				throw DownloadedPackagesSQLLiteStore.StoreError.revokedPackage
			}
		}
	}
```

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V2/DownloadedPackagesStoreV2.swift
```diff
@@ -65,31 +65,50 @@ protocol DownloadedPackagesStoreV2: AnyObject {
...
-			try self.set(country: country, hour: hour, day: day, etag: bucket.etag, package: bucket.package)
+			do {
+				try self.set(country: country, hour: hour, day: day, etag: bucket.etag, package: bucket.package)
+			} catch DownloadedPackagesSQLLiteStore.StoreError.revokedPackage {
+				revokedPackageError = DownloadedPackagesSQLLiteStore.StoreError.revokedPackage
+			}
+		}
+
+		if let revokedPackageError = revokedPackageError {
+			throw revokedPackageError
```
KaiTeuber, line 98: See above

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/V2/DownloadedPackagesStoreV2.swift
ArturFriesen: It's not the same result with you suggested code.
Think of a package in the middle of the list which gets revoked.
What we want, is that the rest of the list is still persisted.
This is why "collect" the revokedPackageError and throw it later.

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
```diff
@@ -413,15 +422,15 @@ final class DownloadedPackagesSQLLiteStoreTests: XCTestCase {
 		XCTAssertEqual(store.allDays(country: "DE").count, 2)
 
 		// add some more data
-		try store.set(country: "DE", day: "2020-06-03", etag: revokedEtag, package: package)
+		try? store.set(country: "DE", day: "2020-06-03", etag: revokedEtag, package: package)
```
whiskey, line 425: why optional? if it fails, the test function (re-)throws the error. No need to omit a potential issue here!
```suggestion
		try store.set(country: "DE", day: "2020-06-03", etag: revokedEtag, package: package)
```

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
```diff
@@ -413,15 +422,15 @@ final class DownloadedPackagesSQLLiteStoreTests: XCTestCase {
...
 		try store.set(country: "IT", day: "2020-06-06", etag: nil, package: package)
 		try store.set(country: "DE", day: "2020-06-07", etag: nil, package: package)
 
 		XCTAssertEqual(store.allDays(country: "DE").count, 3)
 		XCTAssertEqual(store.allDays(country: "IT").count, 1)
 
 		// 2 different 'set' implementations, so let's cover the 2nd as well
-		try store.set(country: "IT", hour: 1, day: "2020-06-03", etag: revokedEtag, package: package)
+		try? store.set(country: "IT", hour: 1, day: "2020-06-03", etag: revokedEtag, package: package)
```
whiskey, line 433: see above
```suggestion
		try store.set(country: "IT", hour: 1, day: "2020-06-03", etag: revokedEtag, package: package)
```

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/HourKeyPackagesDownloadTests.swift
```diff
@@ -293,7 +293,7 @@ class HourKeyPackagesDownloadTests: XCTestCase {
 	func test_When_PersistHourPackagesToDatabaseFails_Then_unableToWriteDiagnosisKeysErrorReturned() {
 		let store = MockTestStore()
 
-		let packagesStore = DownloadedPackagesStoreErrorStub(error: .unknown)
+		let packagesStore = DownloadedPackagesStoreErrorStub(error: DownloadedPackagesSQLLiteStore.StoreError.sqliteError(.unknown))
```
whiskey, line 296: (not blocking) maybe we can declare the `DownloadedPackagesStoreErrorStub ` error type as `StoreError` to simplify the code again or was there a reason to go fully verbose here?

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
whiskey: Oh and IF this throws we can test for that case with `XCTAssertThrow`.

ArturFriesen: It will definitely throw because with this change we throw an error when the package is on the revoke list.
Don't need to test for the throw. Only need to test, wether the package was persisted or not.

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/HourKeyPackagesDownloadTests.swift
ArturFriesen: Lets tackle this with the next change.

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
whiskey: We expect the error to be thrown, this SHOULD be tested. The test will move on after the assert.

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/HourKeyPackagesDownloadTests.swift
whiskey: Just mentioning. I'm not insisting on this one yet.

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
ArturFriesen: done

src/xcode/ENA/ENA/Source/Services/DownloadedPackagesStore/__tests__/DownloadedPackagesSQLLiteStoreTests.swift
ArturFriesen: done

whiskey (approved):

PR #1635, opened 2020-12-07
Fix ExposureDetectionRiskCell background color

30mar:
## Description
Wrong background color on Risk detail card in light mode

the bug seems to be produced because of the "always true" safe unwrapping of the backgroundColor property before the return of the end of the cellForRow function in **DynamicTableViewController**

```
	func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
		if dynamicTableViewModel.section(at: indexPath).isHidden(for: self) {
			return UITableViewCell()
		}

		let section = dynamicTableViewModel.section(at: indexPath)
		let content = dynamicTableViewModel.cell(at: indexPath)

		let cell = tableView.dequeueReusableCell(withIdentifier: content.cellReuseIdentifier, for: indexPath)

		content.configure(cell: cell, at: indexPath, for: self)

		cell.removeSeparators()

		if section.separators != .none {
			let isFirst = indexPath.row == 0
			let isLast = indexPath.row == section.cells.count - 1

			if isFirst && section.separators == .all { cell.addSeparator(.top) }
			if isLast && section.separators == .all { cell.addSeparator(.bottom) }
			if !isLast { cell.addSeparator(.inBetween) }
		}

		if let cellBackgroundColor = cellBackgroundColor {
			cell.backgroundColor = cellBackgroundColor
		}

		return cell
	}
```
the property is already initialized by a default value at the top of the ViewController:
```
	@IBInspectable var cellBackgroundColor: UIColor? = .clear
```
So whatever color was assigned to the cell it is replaced by this default color at this part.

I already smoke tested the app and I don't see this safe unwrapping used affecting any parts of the app and all the screens seems fine now, please let know if it is used in some case that I'm not aware of.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4144

## Screenshot
Simulator Screen Shot - iPhone 11 Pro - 2020-12-08 at 00 16 47
Simulator Screen Shot - iPhone 11 Pro - 2020-12-08 at 00 17 45


whiskey (approved):

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -262,10 +262,6 @@ extension DynamicTableViewController {
 			if !isLast { cell.addSeparator(.inBetween) }
 		}
 
-		if let cellBackgroundColor = cellBackgroundColor {
-			cell.backgroundColor = cellBackgroundColor
-		}
-
```
KaiTeuber, line 268: if cellBackground gets no longer set over here - do we still need the property on top
```
	@IBInspectable var cellBackgroundColor: UIColor? = .clear
```
There are at least 6 classes using that and now it won't get set to the view? I Thinkl all of them use .clear color, so this might be a good moment to remove the property, use .clear as a default color and remove setting the property inside the classes as well?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
30mar: I agree, will do that

30mar:
@KaiTeuber could you please check the last commit when it is possible? thanks

KaiTeuber (approved):

PR #1651, opened 2020-12-09
Feature/4171 diary overview screen

nickguendling:
## Description
Implements the overview screen for the contact diary.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4171

## Screenshots
Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-09 at 11 10 24
Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-09 at 12 18 17


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/DiaryService.swift
```diff
@@ -110,12 +110,14 @@ class MockDiaryStore: DiaryStoring {
 
 	func removeContactPerson(id: Int) {
 		contactPersons.removeAll { $0.id == id }
+		contactPersonEncounters.removeAll { $0.contactPersonId == id }
```
marcussc, line 113: Since I can't comment at the top of the File I have to do it here.
Nice to see a Mock, would be even nicer, if it lived inside its own File :D

marcussc (changes requested):

janetback (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/DiaryService.swift
nickguendling: We will move these models all into separate files once the actual store implementation is done :)

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/DiaryOverviewTableViewController.swift
```diff
@@ -16,14 +17,20 @@ class DiaryOverviewTableViewController: UITableViewController {
 		onEditContactPersonsButtonTap: @escaping () -> Void,
 		onEditLocationsButtonTap: @escaping () -> Void
 	) {
-		self.viewModel = DiaryOverviewViewModel(diaryService: diaryService)
+		self.diaryService = diaryService
```
ArturFriesen, line 20: Maybe we should rename our services to ViewModel?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/DiaryOverviewTableViewController.swift
```diff
@@ -35,41 +42,124 @@ class DiaryOverviewTableViewController: UITableViewController {
...
+			UINib(nibName: String(describing: DiaryOverviewDayTableViewCell.self), bundle: nil),
+			forCellReuseIdentifier: String(describing: DiaryOverviewDayTableViewCell.self)
+		)
+
+		tableView.separatorStyle = .none
+		tableView.rowHeight = UITableView.automaticDimension
+		tableView.estimatedRowHeight = 60
+	}
+
+	private func descriptionCell(at indexPath: IndexPath) -> UITableViewCell {
```
ArturFriesen, line 118:
```suggestion
	private func makeDescriptionCell(at indexPath: IndexPath) -> UITableViewCell {
```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/DiaryOverviewTableViewController.swift
```diff
@@ -35,41 +42,124 @@ class DiaryOverviewTableViewController: UITableViewController {
...
+		guard let cell = tableView.dequeueReusableCell(withIdentifier: String(describing: DiaryOverviewDescriptionTableViewCell.self), for: indexPath) as? DiaryOverviewDescriptionTableViewCell else {
+			fatalError("Could not dequeue DiaryOverviewDescriptionTableViewCell")
+		}
+
+		return cell
+	}
+
+	private func dayCell(at indexPath: IndexPath) -> UITableViewCell {
```
ArturFriesen, line 126:
```suggestion
	private func makeDayCell(at indexPath: IndexPath) -> UITableViewCell {
```

ArturFriesen (approved):

ouboub:
Hi

I have been using the contact diaries that are meanwhile available in the App Store quite extensively in the last weeks. (these are Cluster Diary, Kontakt Tagebuch and Coronika).

I'd like to add some feature proposal about the screenshot presented.

1.  It would be great to have icons, indicating if masks were used and to which extend, also whether distance was kept (say in a lecture). Moreover an icon for a cluster event would be helpful.

2.  Optionally using GPS for finding out addresses of the event/meeting.

3.  Export: besides send a csv file of all contact in the last 14 days to the relevant institution (Gesundheitsamt), it would very useful to allow the export of single encounters, in case I also want to inform my contacts individually if I am positive. And for reasons of privacy I don't want to send the whole list.

Uwe Brauer

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/DiaryOverviewTableViewController.swift
nickguendling: Let's discuss this once we've used it everywhere, might actually make sense to call this a view model

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/DiaryOverviewTableViewController.swift
nickguendling: I was basically extracting code form the `cellForRowAt` func, and I think the naming without `make` is more consistent with the UITableViewDataSource protocol.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/Cells/DiaryOverviewDayTableViewCell.xib
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
nickguendling: Sorry 😅

christianneu:
@ouboub: Thanks for the feedback. Please create a wishlist entry for this, so the users and the stakeholder can track this in a more transparent way. A pull request should not be misused for this.

ouboub:
> @ouboub: Thanks for the feedback. Please create a wishlist entry for your proposal, so the users and the stakeholder can track this in a more transparent way. A pull request should not be misused for it.

ok, I will, sorry.

GisoSchroederSAP:
Dear @ouboub, 

I'd like to emphasize three points here:

1. As my colleagues already did, we have to insist that all improvement/enhancement request and discussions for an existing feature a future functions are documented in the [cwa-wishlist](https://github.com/corona-warn-app/cwa-wishlist) repository. 
2. The new functio of the contact diary is an early version and we are well aware there is room for improvement.
3. And no, you are wrong with the comment mentioned in the referred issue (https://github.com/corona-warn-app/cwa-app-ios/pull/1695) that was incorrectly placed as pull-request:

> in certain aspects but in this aspect seems not to care at all.

Data privacy was and still is an omnipresent topic during design, implementation and test of the CWA functions.  You might want to check the existing blog entry (https://www.coronawarn.app/de/blog/2020-12-28-corona-warn-app-version-1-10/) for the intention of the contact diary feature. So, if your intention is *not* to share certain contacts with the health authority, this is your responsibility. But obviously, there is no need to share your personal list of documented meetings with all of you contacts - again, this is your data and you select carefully what you share and who receives these information.

Side note: I'll copy this response to the above-mentioned wronlgy placed pull request.
Best,
Giso
Corona Warn-App Open Source Team

PR #1654, opened 2020-12-09
Fix: Remove the close button from the screen

30mar:
## Description
The close button is not working is the "Warn others" screen.

**UPDATE**: in the comments it was stated that the updated design in figma doesn't have the button so it should be removed from the app as well.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4181

## Screenshot
IMG_0005
IMG_0006


marcussc (approved):

tcfos (approved):

KaiTeuber:
But you still can pull it down to close the screen?

30mar:
@KaiTeuber no you have to go back to dismiss by pull-down, but that is the expected behavior right ?

KaiTeuber:
>  but that is the expected behavior right ?
I don't know. But after the change you still can pull the screen down and viewmodel.dissmis() is called - I think

nickguendling (changes requested):
There is a hack in `ExposureSubmissionTestResultConsentViewModel` that makes this change also change the text. See line 63-70.

marcussc:
> There is a hack in `ExposureSubmissionTestResultConsentViewModel` that makes this change also change the text. See line 63-70.

@whiskey Maybe you can help out here? :)

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -181,12 +181,7 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
 				self?.model.exposureSubmissionService.loadSupportedCountries(
 					isLoading: isLoading,
 					onSuccess: { supportedCountries in
-						self?.showTestResultSubmissionConsentScreen(
-							supportedCountries: supportedCountries,
-							onDismiss: {
-								self?.showTestResultAvailableCloseAlert()
-							}
-						)
+						self?.showTestResultSubmissionConsentScreen(supportedCountries: supportedCountries, onDismiss: nil)
```
KaiTeuber, line 184: please keep the line break style
```suggestion
self?.showTestResultSubmissionConsentScreen(
 							supportedCountries: supportedCountries,
 							onDismiss: nil
 						)						self?.showTestResultSubmissionConsentScreen(supportedCountries: supportedCountries, onDismiss: nil)
```

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
30mar: Ok sure

KaiTeuber (approved):

nickguendling (approved):

PR #1663, opened 2020-12-10
Update Localizable.legal.strings [EXPOSUREAPP-3829]

janetback:
## Description
Changes to legal Turkish texts to incorporate feedback from lawyers


tcfos (approved):

ArturFriesen (approved):

src/xcode/ENA/ENA/Resources/Localization/tr.lproj/Localizable.legal.strings
```diff
@@ -1,4 +1,4 @@
-/* ExposureSubmissionWarnOthers */
+ExposureSubmissionQRInfo_acknowledgement_2_2/* ExposureSubmissionWarnOthers */
```
marcussc, line 1: This is not the right format and the reason why the build is not going through :/

marcussc (changes requested):

PR #1669, opened 2020-12-11
Fix/4236 positive homescreen flow

whiskey:
## Description
This PR skips the test results view for known (positive) tests and jumps right into the warn others view. It also added a convenience publisher for supported countries (see `AppConfigurationProviding`).

A proper UI test will follow soon™

## Link to Jira
[EXPOSUREAPP-4236](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4236)

## Screenshots
(the 'double separator' is also fixed)
4236-fixed


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -267,6 +273,45 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
 		)
 	}
 
+	private func createWarnOthersViewController() -> ExposureSubmissionWarnOthersViewController {
+		var vc: ExposureSubmissionWarnOthersViewController?
```
marcussc, line 277: I think this can be a let, which would resolve the force unwrap in the return.

whiskey: not because of the closure (`sink`), but this was also my first attempt.

marcussc (approved):

SwissMountainFrog (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -267,6 +273,45 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
...
+							}
+						}
+					)
+					group.leave()
+				})
+			.store(in: &subscriptions)
+
+		group.wait()
+		// swiftlint:disable:next force_unwrapping
+		return vc!
```
ArturFriesen, line 312: Instead of force unwrap and disable the linter, you could use a guard with fatalError. This way you could also provide an meaningful error message.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/AppConfigurationProviding.swift
```diff
@@ -17,6 +17,9 @@ protocol AppConfigurationProviding: AnyObject {
 
 	/// App configuration publisher that provides the latest app config or a locally stored default config.
 	func appConfiguration() -> AnyPublisher<SAP_Internal_V2_ApplicationConfigurationIOS, Never>
+
+	/// The list of partner coutries provided by the app config, or the default country.
```
ArturFriesen, line 21:
```suggestion
	/// The list of partner countries provided by the app config, or the default country.
```

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -267,6 +273,45 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
 		)
 	}
 
+	private func createWarnOthersViewController() -> ExposureSubmissionWarnOthersViewController {
```
nickguendling, line 276: I would prefer to just use the current implementation from showWarnOthersScreen() for such a pressing high priority bug and just extract it to a create func.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -147,14 +147,12 @@ extension CachedAppConfiguration: AppConfigurationProviding {
...
 			// use the cached version
 			return Just(cachedVersion)
-				.receive(on: DispatchQueue.main)
 				.eraseToAnyPublisher()
 		} else {
 			Log.debug("fetching fresh app configuration. forceFetch: \(forceFetch), force: \(force)", log: .appConfig)
 			// fetch a new one
 			return getAppConfig(with: store.appConfigMetadata?.lastAppConfigETag)
 				.map({ $0.config })
-				.receive(on: DispatchQueue.main)
```
ArturFriesen, line 157: Wont this make Ui-Tests fail 🤔 Just asking.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -147,14 +147,12 @@ extension CachedAppConfiguration: AppConfigurationProviding {
 			Log.debug("fetching cached app configuration", log: .appConfig)
 			// use the cached version
 			return Just(cachedVersion)
-				.receive(on: DispatchQueue.main)
```
nickguendling, line 150: Not sure we should make this change on this pressing PR, do we know if anyone is relying on this being returned on the main queue currently?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -267,6 +273,45 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
...
+								isLoading(false)
+								if let error = error {
+									self?.showErrorAlert(for: error)
+								} else {
+									self?.showThankYouScreen()
+								}
+							}
+						}
+					)
+					group.leave()
```
ArturFriesen, line 306: This is perfect for the completion closure.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: Nope. This is not asynchronous as it `Just` returns a value unlike the other branch for fetching a config. Also, the whole Combine flow [would not work](https://forums.swift.org/t/combine-receive-on-runloop-main-loses-sent-value-how-can-i-make-it-work/28631/6) or with a lot of rework – too much for a high bug.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: How? `showWarnOthersScreen` does not return anything and the existing flow in the coordinator expects a view controller to be returned. See `getInitialViewController`

nickguendling: Well, as I said, by extracting it, just as we did for the other two existing cases.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: Obviously not. 
Not working on the main thread or in the main queue does not imply automatic disaster. If his is enabled, the `supportedCountries` Publisher never finishes - try it for yourself.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
nickguendling: I don't think the overcomplicated combine solution is necessary here, we already have working solutions that could be reused in this coordinator, but I'm not blocking here and trusting you that this works for now.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: This is what `getInitialViewController` unfortunately expects. I did not come up with this mess and just copy the general behavior until we refactor the coordinator.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: If a force unwrap fails it's a developer error. No need to adopt mechanisms that are user-targeted here.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: I wanted to refactor the `showWarnOthersScreen` to use THIS controller but did not dare to not break this controller-hell.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
ArturFriesen: What aredoyou mean by user-targeted?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
ArturFriesen: I don't get it.
Whats the problem with moving `group.leave()` to completion closure of the sink?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
ArturFriesen: Hmm not returning on the main thread after calling the CachedAppConfiguration resulted in a UI-Test disaster in the past. What is different here? On which thread is the closure of the sink called now?

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: I hope somebody wrote a test for that disaster!?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: It made no difference in this situation. But it's moved now.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
ArturFriesen: We had a lot disaster in the past by changing code and not returning on the main thread. The solution so far, was to change back and return on main thread again. Because everything else would explode in effort.
Thats why I'm a little but concerned about this change. If it sometimes does not return on main, we might get problems somewhere.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
ArturFriesen: Still cleaner and safer for changes. Thanks

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
whiskey: `fatalError` is to create proper crash logs. This code here would 'just fail' but during local and CI testing. Given a proper test coverage (✅  and still improving), there is no need to write extra guards.

ArturFriesen: Sorry, I don't agree. If it fails on the CI, it would be a good help for the developer in charge to have a proper crash log. No need to put ugly swiftlint disable lines, if you can handle it clean with a guard + error message and as a side effect create a nicer crash log.

PR #1677, opened 2020-12-14
fix the missing close button

puneetmahali1:
## Description
This PR is fix for the missing button on the test result screen.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4268

## Screenshots
IMG_2086855F8062-1


tcfos (approved):

nickguendling (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultConsentViewController.swift
```diff
@@ -27,10 +27,14 @@ class ExposureSubmissionTestResultConsentViewController: DynamicTableViewControl
 	override func viewDidLoad() {
 		super.viewDidLoad()
 		self.title = AppStrings.AutomaticSharingConsent.consentTitle
-		self.navigationController?.navigationItem.rightBarButtonItem = nil
 		setupView()
 	}
 	
+	override func viewDidAppear(_ animated: Bool) {
+		super.viewDidAppear(animated)
+		self.navigationItem.rightBarButtonItem = nil
```
whiskey, line 35: Why is this ADDING a cancel button??? 🤯

nickguendling: Instead of removing it for every following view controller it now just removes it from this single view controller. This code is not in the view controller where the button was missing 😉

puneetmahali1: Yes exactly as Nick said, we pass the nil to the Navigation Controller which effect some other VC that's why this viewDidAppear makes nil only in TestResultConsentVC. Therefore that cancel button missing because of this mess.

puneetmahali1: > Instead of removing it for every following view controller it now just removes it from this single view controller. This code is not in the view controller where the button was missing 😉

Thanks Nick for the quick reply.

whiskey: And all because people can't dismiss modal views properly… Thanks guys!

PR #1691, opened 2020-12-16
Fix/4191 add more screenshots

whiskey:
## Description
This PR adds more views and states to the list of general app screenshots, i.e. the ones for testers and review.
As this only adds code to tests and accessibility identifiers to app code, I'd merge this into `1.10.x`.

It also fixes tons of issues on the road I never wanted to touch…

## Link to Jira
[EXPOSUREAPP-4191](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4191)

## Screenshots
[choose wisely](https://17705-268042230-gh.circle-artifacts.com/0/src/xcode/screenshots/screenshots.tar)
knight


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
```diff
@@ -143,7 +143,9 @@ class ENAUITests_04_ExposureSubmissionUITests: XCTestCase {
 		
 		// This may fail in future because of the Thank You card. So, just check the home screen
 		XCTAssertTrue(app.collectionViews.buttons["AppStrings.Home.submitCardButton"].waitForExistence(timeout: .long))
-		waitForNonExistence(of: app.buttons["AppStrings.ExposureSubmission.primaryButton"], timeout: .long)
+
+		// Back to homescreen
+		XCTAssertTrue(app.collectionViews.buttons["AppStrings.Home.submitCardButton"].waitForExistence(timeout: .long))
```
nickguendling, line 148: Puneet and I invented the `waitForNonExistence` because due to the new modal presentation I think this check will pass even if the submission flow is not dismissed.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
```diff
@@ -311,11 +315,9 @@ class ENAUITests_04_ExposureSubmissionUITests: XCTestCase {
 		// Click secondary button to skip symptoms screens and immediately go to warn others screen.
 		XCTAssertTrue(app.buttons["AppStrings.ExposureSubmission.secondaryButton"].waitForExistence(timeout: .medium))
 		app.buttons["AppStrings.ExposureSubmission.secondaryButton"].tap()
-
-		app.alerts["Sie wollen keine Warnung verschicken?"].buttons["Nicht warnen"].tap()
-		
-		waitForNonExistence(of: app.buttons["AppStrings.ExposureSubmission.primaryButton"], timeout: .long)
```
nickguendling, line 317: Is this dismissed by `addUIInterruptionMonitor(withDescription: "System Dialog")`? How do we distinguish between two different alert actions? Which one is triggered?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
```diff
@@ -401,20 +414,67 @@ class ENAUITests_04_ExposureSubmissionUITests: XCTestCase {
...
+
+		// Open Intro screen.
+		XCTAssertTrue(app.collectionViews.buttons["AppStrings.Home.submitCardButton"].waitForExistence(timeout: .long))
+		app.collectionViews.buttons["AppStrings.Home.submitCardButton"].tap()
+		snapshot("tan_submissionflow_qr_\(String(format: "%04d", (screenshotCounter.inc() )))")
+
+		// Overview Screen: click TAN button.
+		XCTAssertTrue(app
+						.buttons["AppStrings.ExposureSubmissionDispatch.qrCodeButtonDescription"]
+						.waitForExistence(timeout: .medium)
```
nickguendling, line 459:
```suggestion
			.buttons["AppStrings.ExposureSubmissionDispatch.qrCodeButtonDescription"]
			.waitForExistence(timeout: .medium)
```

src/xcode/ENA/ENA/Source/View Helpers/AccessibilityIdentifiers.swift
```diff
@@ -288,6 +292,10 @@ enum AccessibilityIdentifiers {
 	enum General {
 		static let exposureSubmissionNavigationControllerTitle = "ExposureSubmissionNavigationController"
 		static let image = "ExposureSubmissionIntroViewController.image"
+
+		static let primaryFooterButton = "General.primaryFooterButton"
+		static let secondaryFooterButton = "General.secondaryFooterButton"
+		static let cancelButton = "General.cancelButton"
```
nickguendling, line 298: Are these used anywhere already or just set so far?

src/xcode/ENA/ENAUITests/Helper/AccessibilityLabels.swift
```diff
@@ -0,0 +1,94 @@
+////
+// 🦠 Corona-Warn-App
+//
+// helper functions fur UITests
+
+import Foundation
+import XCTest
+
+enum accessibilityLabels {
```
nickguendling, line 9:
```suggestion
enum AccessibilityLabels {
```

src/xcode/ENA/TestPlans/AllTests.xctestplan
```diff
@@ -43,6 +43,15 @@
...
       }
+    },
+    {
+      "enabled" : false,
+      "id" : "23BB78C4-766D-4EE2-87E6-E0C288E37D26",
+      "name" : "fake-Klingon",
+      "options" : {
+        "language" : "IDELaunchSchemeLanguageRLO",
+        "region" : "AS"
+      }
```
nickguendling, line 54: What? 😅

nickguendling (changes requested):
More screenshots 🥳

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
nickguendling: I just quickly checked, for example the test `test_SubmitTAN_CancelOnTestResultScreen()` has this check but never actually arrives back at the home screen.

whiskey: I know you both worked on this but I oversaw some major flaws in that design during my review.
That implementation never passed the thrown error to the test. All I got was a grey notification that the expectation was not triggered but the test passed green - bad thing.

Also design wise I think it's still very weird to check for the non-existance of something and not the existence of the thing that should be there. Not having something in the `app` hierarchy does not mean it's ok. The view could be broken and this check would result in a success, like a check for the non-existence of zombies ;)
corona is fine

About the `test_SubmitTAN_CancelOnTestResultScreen`, I wonder why the test passes then? If we really need this 'non-existence' thing, we could bring it back but then in a consistent way and not the old implementation. I'm happy to help with that.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
whiskey: This way of using `addUIInterruptionMonitor` by just returning `false` triggers the default implementation which uses the `.default` action on the alert. We can tweak the handler, make our own logic, e.g. by using identifiers to find the right button on the right alert and return `true` as in 'yes, we got this covered'.

I consider this monitor as a general handler for all kind of non-deterministic interruptions and try to keep it simple. If I know for sure that a certain alert will trigger AND I need to handle that alert in a very particular way as in `test_ThankYouScreen_withWarnOthers` I target the (expected) alert directly and handle it in place.

There is [another resource on this topic](https://useyourloaf.com/blog/handling-system-alerts-in-ui-tests/) and we can cover more questions in the dev meeting if needed as this feature is generally really cool to handle privacy prompts.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/__tests__/ExposureSubmissionUITests.swift
whiskey: Thanks, I had tons of white space issues here in Github and considered them only as graphics glitches.
This is what you get by using tabs instead of spaces… 😜

src/xcode/ENA/ENA/Source/View Helpers/AccessibilityIdentifiers.swift
whiskey: Are live and kicking in `ENANavigationFooterView` as default for the footer buttons. I haven't touched the current custom settings of these in existing code to reduce noise in this PR.

The `ExposureSubmissionQRScannerViewController` uses the cancel button but not in any test yet. I planned something for the screenshots and postponed it until further working on screenshots.

src/xcode/ENA/TestPlans/AllTests.xctestplan
whiskey: Absolutely needed to ensure locale-agnostic tests. All previous and current tests might pass for ze Germanz but not for any other locale 😱! See all the issues above.
If we write future tests and refactor the current failing ones, this might even become the new standard configuration to be used in the CI tests. If THIS thing passes, ANYTHING will.

tcfos (approved):

nickguendling (approved):

PR #1692, opened 2020-12-16
Fix: show correct screen based on test result when testResultNotifica…

30mar:
## Description
Show appropriate screen when user taps on the "test result is available notification"

**According to Figma the following should happen based on the possible cases for the test result:**

- pending or expired: do **nothing**
- Positive: show **TestResultAvailableViewController**
- Negative: show **ExposureSubmissionTestResultViewController**
- invalid: there is no scenario so we show **HomeViewController** for now.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4262

### Important: please check the video attached to the ticket comments to see the whole flow

IMG_157D1D48A72F-1


src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -118,15 +118,32 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
...
+			guard let testResultRawValue = response.notification.request.content.userInfo[ActionableNotificationIdentifier.testResult.identifier] as? Int,
+				  let testResult = TestResult(rawValue: testResultRawValue) else {
+				showHome(animated: true)
+				return
+			}
+			
+			switch testResult {
+			case .positive, .negative:
+				showTestResultFromNotification(with: testResult, animated: true)
+
```
ArturFriesen, line 139: New line can be removed ✌️

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -118,15 +118,32 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
...
+			}
+			
+			switch testResult {
+			case .positive, .negative:
+				showTestResultFromNotification(with: testResult, animated: true)
+				
+			case .invalid:
+				showHome(animated: true)
+				
+			default:
```
ArturFriesen, line 143: Did you consider, to explicitly handling all cases? This way, if a new case is added, the compiler will fail here and the future-developer has to think about showing some UI or not for the new case.

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -223,13 +240,17 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
 		coordinator.showHome(enStateHandler: enStateHandler)
 	}
 	
-	private func showPositiveTestResultFromNotification(animated _: Bool = false) {
+	private func checkIfPositiveTestResultWasShown() {
```
ArturFriesen, line 243: I would propose to name it: "showPositiveTestResultIfNeeded()"

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -223,13 +240,17 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
...
 		let warnOthersReminder = WarnOthersReminder(store: store)
 		guard warnOthersReminder.positiveTestResultWasShown else {
 			return
 		}
+		showTestResultFromNotification(with: .positive, animated: true)
+	}
 
-		coordinator.showPositiveTestResultFromNotification(with: .positive)
+	private func showTestResultFromNotification(with testResult: TestResult, animated _: Bool = false) {
```
ArturFriesen, line 251: `animated` is not used and can be removed.

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/SceneDelegate.swift
```diff
@@ -118,15 +118,32 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate, RequiresAppDepend
...
 			 ActionableNotificationIdentifier.deviceTimeCheck.identifier:
 			showHome(animated: true)
 
 		case ActionableNotificationIdentifier.warnOthersReminder1.identifier,
 			 ActionableNotificationIdentifier.warnOthersReminder2.identifier:
-			showPositiveTestResultFromNotification(animated: true)
+			checkIfPositiveTestResultWasShown()
 
+		case ActionableNotificationIdentifier.testResult.identifier:
+			guard let testResultRawValue = response.notification.request.content.userInfo[ActionableNotificationIdentifier.testResult.identifier] as? Int,
```
whiskey, line 130: Wow, `response.notification.request.content.userInfo[ActionableNotificationIdentifier.testResult.identifier]` is just…
6a3b1480-d93d-11e9-82c8-3780f102a035

whiskey (approved):
Normally I would ask for a UI test but I know for sure, it's currently very hard to achieve with the actual home view controller.

src/xcode/ENA/ENA/Source/SceneDelegate.swift
30mar: so make a fatal error for the default case? I thoguht about it actually but i didn't see it alot in the code here. ok will do it

src/xcode/ENA/ENA/Source/SceneDelegate.swift
30mar: I used two screens for that :'D
Will try to make it more readable

PR #1693, opened 2020-12-16
Fix/4229 unexpected popup

SwissMountainFrog:
This PR fixes [the issue 4229](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4229) with the "Bitte geben Sie Ihre Zufalls-IDs frei, um andere zu warnen." popup appears if the OS authorization for exposure submission was not granted.


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -283,7 +283,13 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
 				self?.model.exposureSubmissionService.getTemporaryExposureKeys { error in
 					isLoading(false)
 					if let error = error {
-						self?.showErrorAlert(for: error)
+						if case .notAuthorized = error {
+							self?.model.exposureSubmissionService.isSubmissionConsentGiven = false
```
KaiTeuber, line 287: both cases do set ```isSubmissionConsentGiven = false``` - better move out of the if?
I also wonder why not use ```if error == .notAuthorized``` here? It's better readable in my opinion.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
SwissMountainFrog: I have moved the line out if the if/else. 
The second proposal won't work as we have `if case .notAuthorized` (`if case` syntax).

KaiTeuber:
```
self?.model.exposureSubmissionService.isSubmissionConsentGiven = false
if error == .notAuthorized {
	Log.info("ExposureSubmissionCoordinator: Submission consent reset to false after OS authorization was not given.")
} else {
	self?.showErrorAlert(for: error)
}
```
if case syntax isn't needed, there is no associated value inside that case?
We can have a short talk about that, if you want to.

SwissMountainFrog: Refactored

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/SAPDownloadedPackageTests.swift
```diff
@@ -28,7 +28,7 @@ final class SAPDownloadedPackageTests: XCTestCase {
 		let bytes = [0xA, 0xB, 0xC, 0xD]
 		// The bin and signature were  made for different data sets
 		let package = try SAPDownloadedPackage.makePackage(bin: Data(bytes: bytes, count: 4),
-														   signature: try SAPDownloadedPackage.makeSignature(data: Data(bytes: bytes, count: 3), key: signingKey).asList() //swiftlint:disable:this vertical_parameter_alignment_on_call
+														   signature: try SAPDownloadedPackage.makeSignature(data: Data(bytes: bytes, count: 3), key: signingKey).asList() // swiftlint:disable:this vertical_parameter_alignment_on_call
```
nickguendling, line 31: The linter warnings are already fixed on the release branch, it's probably the easiest to just revert that commit and update your branch from the release branch.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -205,17 +205,20 @@ class ExposureSubmissionCoordinator: NSObject, ExposureSubmissionCoordinating, R
...
-							self.showErrorAlert(for: error)
-						}
-					} else {
+					
+					guard let error = error else {
 						self.showTestResultScreen(with: testResult)
+						return
+					}
+					
+					self.model.exposureSubmissionService.isSubmissionConsentGiven = false
```
nickguendling, line 214: Do we really want to retract the consent if the framework throws an error other than `.notAuthorized`?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/SAPDownloadedPackageTests.swift
SwissMountainFrog: There was only this change marked as conflict ;-) so easy merging ;-)

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
SwissMountainFrog: Very good point. No, only on `.notAuthorized`. Changed 👍

nickguendling (approved):

PR #1716, opened 2020-12-18
Get last exposure detection date from store

nickguendling:
## Description
Get the last exposure detection date from the store so it's also correct for the inactive state

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4370

## Screenshots
IMG_0035

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
@@ -215,7 +215,6 @@ extension HomeInteractor {
 // MARK: - Action section setup helpers.
 
 extension HomeInteractor {
-	private var riskDetails: Risk.Details? { state.riskDetails }
 
 	func setupRiskConfigurator() -> CollectionViewCellConfiguratorAny? {
 		let detectionIsAutomatic = detectionMode == .automatic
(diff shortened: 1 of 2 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -219,7 +219,7 @@ extension HomeInteractor {
 
 	func setupRiskConfigurator() -> CollectionViewCellConfiguratorAny? {
 		let detectionIsAutomatic = detectionMode == .automatic
-		let dateLastExposureDetection = riskDetails?.exposureDetectionDate
+		let dateLastExposureDetection = store.riskCalculationResult?.calculationDate
```
ArturFriesen, line 222: Is there a reason why riskDetails is not uptodate?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
```diff
@@ -219,7 +219,7 @@ extension HomeInteractor {
```
ArturFriesen, line 219: Can riskDetails now be completely removed from the state, because I don't see any usage.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
nickguendling: `riskDetails` is actually nil for the `RiskState`s `.inactive` and `.detectionFailed`, because they don't have an associated `Risk`. That's why the date was also nil.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeInteractor.swift
nickguendling: Eagle eyes! 🦅 Have removed it, it was not used anymore.

ArturFriesen (approved):

PR #1718, opened 2020-12-18
Task/4399–rescue on fail

ArturFriesen:
## Description
This PR adds rescue logic to the initialization of the contact diary store.
If the init doesn't work, the store folder is removed and the store is recreated.

## Link to Jira
[4399](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4399)


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -592,44 +600,26 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
 			return .failure(error)
 		}
 
-		openAndSetup()
+		let openAndSetupResult = openAndSetup()
+		if case .failure = openAndSetupResult {
+			return openAndSetupResult
+		}
 
+		var updateDiaruesResult: DiaryStoringVoidResult?
```
nickguendling, line 608:
```suggestion
		var updateDiaryDaysResult: DiaryStoringVoidResult?
```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -854,26 +888,74 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
 	}
 }
 
+// MARK: Creation
+
 extension ContactDiaryStoreV1 {
-	convenience init(fileName: String) {
+
+	static func make(filename: String) -> ContactDiaryStoreV1 {
```
nickguendling, line 895: Can we turn the naming the other way around, so that we can still use `init(filename:)` as the initializer used in AppDelegate?

ArturFriesen: Was my first attempt. Compiler says no "A non-failable initializer cannot delegate to failable initializer"
Do you know how to do it?

nickguendling: Okay, I understand. Spontaneously I do not have a solution. But I quickly tried and realized on another note, that the `filename` does not seem to be used anymore.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -24,7 +24,7 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
 
 	// MARK: - Init
 
-	init(
+	init?(
```
KaiTeuber, line 27: if init is only used by make method, why isn't it private?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -854,26 +888,74 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
...
+		if let secondTryStore = ContactDiaryStoreV1(fileName: filename) {
+			Log.info("[ContactDiaryStore] Successfully rescued contact diary store", log: .localData)
+			return secondTryStore
+		} else {
+			Log.info("[ContactDiaryStore] Failed to rescue contact diary store.", log: .localData)
+			fatalError("[ContactDiaryStore] Could not create contact diary store after second try.")
+		}
+	}
+
+	convenience init?(fileName: String) {
```
KaiTeuber: This one is only used from the make method as well - still only internal?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
KaiTeuber: Why move this into an extension instead inside the MARKs section?

nickguendling (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
ArturFriesen: I moved it to an extension, because all this stuff is in context of "convenience creation". It does not really belongs to the store. Just logic to have a simpler creation process.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
ArturFriesen: Its used by testing.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
ArturFriesen: Its still a possible, convenience way to create the store. One could use it in testing to avoid the rescue logic in make.

PR #1719, opened 2020-12-18
Fix/4226 dark mode glitches

puneetmahali1:
## Description
This PR fix the all dark mode glitches.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4226#anchor

## Screenshots
IMG_0020
IMG_0021
IMG_0022
IMG_0023
IMG_0024
IMG_0025
IMG_0026
IMG_0027

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
@@ -133,7 +133,7 @@ extension DynamicCell {
 		.identifier(CellReuseIdentifier.space) { _, cell, _ in
 			guard let cell = cell as? DynamicTableViewSpaceCell else { return }
 			cell.height = height
-			cell.backgroundColor = color
+			cell.backgroundColor = .clear
 		}
 	}
 	
--- a/src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
@@ -129,7 +129,7 @@ extension DynamicCell {
 		)
 	}
 
-	static func space(height: CGFloat, color: UIColor? = .clear) -> Self {
+	static func space(height: CGFloat, color: UIColor = .clear) -> Self {
 		.identifier(CellReuseIdentifier.space) { _, cell, _ in
 			guard let cell = cell as? DynamicTableViewSpaceCell else { return }
 			cell.height = height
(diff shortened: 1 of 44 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultConsentViewModel.swift
```diff
@@ -75,7 +75,8 @@ class ExposureSubmissionTestResultConsentViewModel {
 									)
 								}
 							},
-						.space(height: 20)
+						.space(height: 20,
+							   color: .clear)
```
nickguendling, line 79: Can't we set this to `.clear` by default?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionTestResultConsentViewModel.swift
puneetmahali1: Thanks Nick. Yes we can & fixed it.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
```diff
@@ -133,7 +133,7 @@ extension DynamicCell {
 		.identifier(CellReuseIdentifier.space) { _, cell, _ in
 			guard let cell = cell as? DynamicTableViewSpaceCell else { return }
 			cell.height = height
-			cell.backgroundColor = color
+			cell.backgroundColor = .clear
```
nickguendling, line 136: Would be a better idea to leave it as it was and change the function to `static func space(height: CGFloat, color: UIColor = .clear) -> Self {`

puneetmahali1: Thanx! Done.

KaiTeuber:
Is a clear background color really the solution we want here?
Transparency is slower in rendering and we have had side effects with it in the past. e.g. the red tile once shone through in another section.
Wouldn't it be more effective to set the expected color directly?

puneetmahali1:
> Is a clear background color really the solution we want here?
> Transparency is slower in rendering and we have had side effects with it in the past. e.g. the red tile once shone through in another section.
> Wouldn't it be more effective to set the expected color directly?

I was not aware of the side effects and didn't see any problems there. Yes, we can set the expected colors then we need those colors. Also, there is no figma screens for the dark mode. So if this is the case then need to ask from the designers and set the expected colors.

nickguendling (approved):

ArturFriesen (approved):

tcfos (approved):

PR #1737, opened 2020-12-23
fix Prefixed zeros not being saved correctly in Contact Journal

30mar:
## Description
Saving numbers starting with zeros is incorrect, when you save "001" or 01231" and then you fetch you get: "1" and "1231"

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4380

## Screenshots
IMG_0040


nickguendling (commented):
What about people that already have the app installed with the old database schema. Should we do a migration?

whiskey (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/__tests__/ContactDiaryStoreV1Tests.swift
```diff
@@ -51,7 +51,47 @@ class ContactDiaryStoreV1Tests: XCTestCase {
 
 		XCTAssertEqual(name, "Hinterm Mond")
 	}
+	
+	func test_When_addZeroPrefixedLocation_Then_LocationIsPersistedCorrectly() {
+		let databaseQueue = makeDatabaseQueue()
+		let store = makeContactDiaryStore(with: databaseQueue)
+
+		let result = store.addLocation(name: "0043")
```
tcfos, line 59: Why not using constants instead of string literals?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -50,10 +50,48 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
...
+		migrateColumns(inTable: "ContactPerson")
+		migrateColumns(inTable: "Location")
 
 		registerToDidBecomeActiveNotification()
 	}
-
+	
+	func migrateColumns(inTable tableName: String) {
+		databaseQueue.inDatabase { database in
+			let queryResult = database.prepare("PRAGMA table_info(" + tableName + ")" )
```
whiskey, line 62: What's the deal with string subsitution?

```suggestion
			let queryResult = database.prepare("PRAGMA table_info(\(tableName))" )
```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -50,10 +50,48 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
...
 		}
+		
+		migrateColumns(inTable: "ContactPerson")
+		migrateColumns(inTable: "Location")
 
 		registerToDidBecomeActiveNotification()
 	}
-
+	
+	func migrateColumns(inTable tableName: String) {
```
marcussc, line 60: Can we please create the migration similar to the DownloadedPackageStore?
This would mean creating: ContactDiaryStoreV1 a Migration that implements the `Migration` Protocol

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -50,10 +50,48 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
...
+				if name == "name" && type == "STRING" {
+					finalSQL = """
+				    ALTER TABLE \(tableName) RENAME TO tmp;
+				    CREATE TABLE \(tableName) (
+				    id INTEGER PRIMARY KEY,
+				    name TEXT NOT NULL CHECK (LENGTH(name) <= 250)
+				    );
+				    INSERT INTO \(tableName)(id, name)
+				    SELECT id, name
+				    FROM tmp;
```
whiskey, line 79: Needed?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
30mar: it is the same, i will rewrite it

30mar: I will try the other way

30mar: Screenshot 2020-12-29 at 13 31 19
Screenshot 2020-12-29 at 13 31 49

30mar: as you see in the screenshots the .next() returns false with the normal String interpolation

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
30mar: To copy the data from the tmp table to the new one

whiskey: 🤦‍♂️  I was expecting a semicolon in line 76, which does not exist. Indenting these two lines might help to understand the statement.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
whiskey: Thanks for the verification. I still wonder why this is the case 🤨

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -688,7 +687,9 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
 		if case let .failure(error) = schemaCreateResult {
 			return .failure(.database(error))
 		}
-
+		databaseQueue.inDatabase { database in
+			print(database.tableExists("Location"))
```
marcussc, line 691: Still needed? :D

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
```diff
@@ -648,7 +647,7 @@ class ContactDiaryStoreV1: DiaryStoring, DiaryProviding {
 
 	private let databaseQueue: FMDatabaseQueue
 
-	private func openAndSetup() -> DiaryStoringVoidResult {
+	func openAndSetup() -> DiaryStoringVoidResult {
```
marcussc, line 650: really no longer private?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -0,0 +1,995 @@
...
+//		let schema = ContactDiaryStoreSchemaV1(
+//			databaseQueue: databaseQueue
+//		)
+//
+//		self.init(
+//			databaseQueue: databaseQueue,
+//			schema: schema,
+//			key: ContactDiaryStoreV2.encryptionKey
+//		)
+//	}
```
whiskey, line 933: Needed?

```suggestion
```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -0,0 +1,995 @@
...
+	// MARK: - Protocol DiaryProviding
+
+	var diaryDaysPublisher = CurrentValueSubject<[DiaryDay], Never>([])
+
+	func export() -> Result<String, SQLiteErrorCode> {
+		var result: Result<String, SQLiteErrorCode>?
+
+		databaseQueue.inDatabase { database in
+			Log.info("[ContactDiaryStore] export entries.", log: .localData)
+			var contentHeader = "Kontakte der letzten 14 Tage (%@ - %@)\nDie nachfolgende Liste dient dem zuständigen Gesundheitsamt zur Kontaktnachverfolgung gem. § 25 IfSG."
```
whiskey, line 57: Shouldn't we localize this – even if some people think there are only German speaking entities in this country?

```suggestion
			var contentHeader = NSLocalizedString("some.key", tableName: nil, bundle: nil, value: "Kontakte der letzten 14 Tage (%@ - %@)\nDie nachfolgende Liste dient dem zuständigen Gesundheitsamt zur Kontaktnachverfolgung gem. § 25 IfSG.", comment: "Currently not translated!")

```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/__tests__/ContactDiaryStoreV2Tests.swift
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			guard queryResult.next() else {
+				return
+			}
+
+			result = queryResult
+		}
+
+		return result
```
whiskey, line 103: So
much
white
space

:)

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
whiskey, line 691: Please use `Log`!

```suggestion
			Log.debug(database.tableExists("Location"), log: .localData)

```

whiskey (approved):
This migration mechanism is… 🙄

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreV1.swift
30mar: Was testing inhering from V1 approach, forgot to change it back, will update the PR soon

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
30mar: nope, forgot to delete it before committing, will push a new commit soon

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
30mar: I guess we are not localizing because this whole export feature is only available for Germany "at least for now", same goes for date formatting we are only using German Local

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/__tests__/ContactDiaryStoreV1Tests.swift
```diff
@@ -51,7 +51,7 @@ class ContactDiaryStoreV1Tests: XCTestCase {
 
 		XCTAssertEqual(name, "Hinterm Mond")
 	}
-
+
```
ArturFriesen, line 54: The old implementation of the store can be removed. In case of the packages store, the old implementation was only used for "faster" testing purpose. After discussing about that, its better to remove the old implementation and only use the Schema for migration testing.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/__tests__/ContactDiaryStoreV2Tests.swift
```diff
@@ -0,0 +1,101 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import XCTest
+import FMDB
+import Combine
+@testable import ENA
+
+class ContactDiaryStoreV2Tests: XCTestCase {
```
ArturFriesen, line 10: Please add a migration test.

1. Create the old schema
2. Write data into the database created by the old schema
3. Create new store
4. Execute migration
5. Test if the data is still there after the migration.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -0,0 +1,977 @@
...
+			
+			guard database.open() else {
+				Log.error("[ContactDiaryStore] Database could not be opened", log: .localData)
+				errorResult = .failure(dbError(from: database))
+				return
+			}
+			// We need to check for the following:
+			// (userVersion = 0) means: no database was created and stored previously by the user --> this is the first database --> create new updated schema
+			// if a database was stored but not the latest version --> migrate the old schema
+			if database.userVersion != 0 && database.userVersion < latestVersion {
```
ArturFriesen, line 662: I think this is not needed if you place the migration after the schema creation.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -0,0 +1,977 @@
...
+				// we create the migration here and not in the init() because the database need to be open and decrypted before migration or it will fail
+				// we added the migration twice because there is a bug here; initial database version should be 0 not 1, this is a temp fix
+				let migrations: [Migration] = [ContactDiaryMigration1To2(database: database), ContactDiaryMigration1To2(database: database)]
```
ArturFriesen, line 665: Adding the migration twice doesn't fix the root problem.
Better fix the root problem by decouple access to the migration list from the version inside of the migrator implementation.

src/xcode/ENA/ENA/Source/Workers/Keychain/KeychainHelper.swift
```diff
@@ -105,7 +105,7 @@ struct KeychainHelper {
 		}
 
 		let key = "x'\(Data(bytes).hexEncodedString())'"
-		try saveToKeychain(key: ContactDiaryStoreV1.encriptionKeyKey, data: Data(key.utf8))
+		try saveToKeychain(key: ContactDiaryStoreV2.encriptionKeyKey, data: Data(key.utf8))
```
ArturFriesen, line 108: I think we can remove the versioning from the store name. With the decoupled schema implementations, we should not need the old store versions anymore.

ArturFriesen (changes requested):
Migration is hard, thanks for your fix! ⭐️

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/StoreV2/ContactDiaryStoreV2.swift
30mar: Turns out we need to place the migration before the schema creation, the schema will change the database version, so if we have an old database and we initialize the store --. > at the schema creation step we will change the database version ---> now if we try to migrate the migrator won't execute as the userDatabase version is now the latest because of the scema step

ArturFriesen (approved):
Nice! ⭐️

PR #1741, opened 2020-12-28
Fix/4469 wrong text

KaiTeuber:
## Description
Remove image description from Contact Diary illustration

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4469

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Info/DiaryInfoViewModel.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Info/DiaryInfoViewModel.swift
@@ -25,6 +25,7 @@ struct DiaryInfoViewModel {
 						UIImage(
 							imageLiteralResourceName: "Illu_ContactDiary-Information"
 						),
+						accessibilityLabel: AppStrings.ContactDiary.Information.imageDescription,
 						accessibilityIdentifier: AccessibilityIdentifiers.ContactDiaryInformation.imageDescription
 					),
 				cells: [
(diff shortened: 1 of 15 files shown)


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Info/DiaryInfoViewModel.swift
```diff
@@ -25,7 +25,6 @@ struct DiaryInfoViewModel {
 						UIImage(
 							imageLiteralResourceName: "Illu_ContactDiary-Information"
 						),
-						accessibilityLabel: AppStrings.ContactDiary.Information.imageDescription,
```
marcussc, line 28: Is the string used somewhere else or could it be removed from the `AppStrings`?

KaiTeuber: AccessibilityIdentifiers.ContactDiaryInformation.imageDescription points to it
But that one is only used here, shall we remove it too?

KaiTeuber: But accessibilityIdentifier is required for .image cells

nickguendling: I think the AccessibilityIdentifiers don't actually use the AppStrings. They just use "AppStrings." as prefix which is a bit confusing. I think we could change that to make this more clear and e.g. use "AccessibilityIdentifier." as a prefix. But for this PR I think it's save to just delete the string that is not used anymore from the Localizable.strings file.

KaiTeuber: @nickguendling Good point, I#ll remove them

nickguendling: I just took another look, I think the JIRA ticket is probably referring to the book illustration on the Home Screen, not the illustration on top of the information view controller. Because I think the hand holding a mobile, that is also mentioned, is referring to the test result/submission cell that also does not have an accessibility label 🤔

PR #1750, opened 2020-12-29
better use swipe up instead of coordinate system based touch

KaiTeuber:
## Description
Improved UITest

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4483


marcussc (approved):

whiskey (approved):
I'm still puzzled why we have to do this at all. AFAIK, a cell can just be checked for existence, even when 'further down a table'. It's terribly slow but works under normal conditions.

marcussc:
> I'm still puzzled why we have to do this at all. AFAIK, a cell can just be checked for existence, even when 'further down a table'. It's terribly slow but works under normal conditions.

@whiskey are you sure? How can a UI-Test "see" a cell that isn't even created, since it is that far off screen? Just curious :D

whiskey:
> > I'm still puzzled why we have to do this at all. AFAIK, a cell can just be checked for existence, even when 'further down a table'. It's terribly slow but works under normal conditions.
> 
> @whiskey are you sure? How can a UI-Test "see" a cell that isn't even created, since it is that far off screen? Just curious :D

@marcussc that's what I did in my app and I was selecting a cell further down the table. I would assume the systems does some kind of 'scrolling' under the hood or (hopefully) traverses the data source to check for a queued cell. Maybe I can create a quick test later this week.

This scrolling method we now have helps with snapshotting a whole view so it's useful in any case.

KaiTeuber:
I have not managed to select a cell that is not visible. Scrolling it into the visible area seems to be a common solution.

PR #1757, opened 2021-01-04
fix/4476-Loop_After_Reset

ArturFriesen:
## Description
Set the home view controller to nil if the coordinator navigates to onboarding.
Navigation to onboarding happens after reseting the app.

## Link to Jira
[4476](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4476)


marcussc (approved):

src/xcode/ENA/ENA/Source/Coordinator.swift
```diff
@@ -111,6 +111,8 @@ class Coordinator: RequiresAppDependencies {
 			],
 			animated: false
 		)
+
```
whiskey, line 114: Maybe tell why we are nilling it:

```suggestion

		// prevents a loop
```

whiskey (approved):

src/xcode/ENA/ENA/Source/Coordinator.swift
ArturFriesen: added comment

PR #1762, opened 2021-01-06
iOS 12.5 preparations 

marcussc:
## Description
This PR prepares support for iOS 12.5 and refactored the Home-Screen into a TableView

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4179
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-1958


src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -18,20 +19,296 @@ protocol CoronaWarnAppDelegate: AnyObject {
...
-extension AppDelegate: CoronaWarnAppDelegate {
-	// required - otherwise app will crash because cast will fail.
+	func requestUpdatedExposureState()
+
 }
 
-extension AppDelegate {
+@UIApplicationMain
+// swiftlint:disable:next type_body_length
+class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, RequiresAppDependencies, ENAExposureManagerObserver, CoordinatorDelegate, UNUserNotificationCenterDelegate, ExposureStateUpdating, ENStateHandlerUpdating {
```
whiskey, line 29: That's a great example why extensions for protocol implementations might be useful. Separate classes/structs for stuff like ENA observation or Notifications would be even nicer to reduce the size of this AppDelegate.

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -75,7 +75,7 @@ class CachingHTTPClient: AppConfigurationFetching {
 				// serialize config
 				do {
 					let config = try SAP_Internal_V2_ApplicationConfigurationIOS(serializedData: package.bin)
-					let eTag = response.httpResponse.value(forHTTPHeaderField: "ETag")
+					let eTag = response.httpResponse.value(forCaseInsensitiveHeaderField: "ETag")
```
whiskey, line 78: I would stick to the old name and 'just make it' case insensitive as the standard behaviour. They were also more meaningful. We only had issues because we use the wrong (case sensitive) way of accessing the fields.

ArturFriesen: `forHTTPHeaderField` is already case-insensitive. From the docs: "The name is case-insensitive."

nickguendling: The old name is part of Foundation of iOS 13+, how could I keep the old name and make it work for iOS 12 as well?

ArturFriesen (approved):
Wow! Breathtaking! How did you manage to pull this off in such a short time?! ⭐️⭐️⭐️

src/xcode/ENA/ENA/Source/Scenes/AppDisabled/AppDisabledViewModel.swift
```diff
@@ -0,0 +1,9 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import UIKit
+
+struct AppDisabledViewModel {
+	let text = "Sorry no App for you, please upgrade your OS 🤙"
```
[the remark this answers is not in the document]
marcussc: Of course not, but the screen will not be shown to customers :) Its just a placeholder.

nickguendling (approved):

pavshr:
@marcussc 

Hello, is there anyway to trigger execution of the managers setLaunchActivityHandler() from outside the app? E.g., for BGTaskScheduler it is possible to time travel for 6 hours in the future and it will trigger the execution of the background task. Anyway to do the same for iOS 12.5 way? Thanks.

Great work, btw.

marcussc:
@pavshr 
As far as we know there is no way to trigger this manually. The spec says that it should be triggered at least every 24 Hours. In my personal testing it feels like its getting called every 1-4 Hours. Jumping into the future is something I would not recommend since we check the device time against the CWA-Server time and if the difference is to big no risk calculation is done.

PR #1774, opened 2021-01-11
Admin/improve sonar coverage

marcussc:
## Description
This PR improves the generation of Codecoverage for SonarCloud by replacing the script with a small program which utilises more than one CPU.


tcfos (approved):

whiskey (approved):
@marcussc Sonar is still having `develop` as the main branch. As far as I remember you don't have permissions to change that but can you ping maybe @kaddaSz to change this to `main`, as it's the most stable branch.

PR #1790, opened 2021-01-13
Fix/load test result initially

nickguendling:
## Description
After starting the app from Xcode the test result was not loaded. So I added a loading of the test result right after the home view controller did load.

Also fixed some indentation.


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/NewHome/HomeTableViewController.swift
```diff
@@ -41,12 +41,12 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 		super.init(style: .plain)
 
 		viewModel.state.$testResult
-		   .sink { [weak self] _ in
-			DispatchQueue.main.async {
-			 self?.reload()
+			.sink { [weak self] _ in
+				DispatchQueue.main.async {
```
KaiTeuber, line 45: Would it be better to put the DispatchQueue.main.async in the reload method. Whoever calls it then doesn't have to worry about how it is correct.

src/xcode/ENA/ENA/Source/Scenes/NewHome/HomeTableViewController.swift
```diff
@@ -87,6 +87,8 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 		tableView.backgroundColor = .enaColor(for: .separator)
 
 		NotificationCenter.default.addObserver(self, selector: #selector(refreshUIAfterResumingFromBackground), name: UIApplication.didBecomeActiveNotification, object: nil)
+
+		viewModel.state.updateTestResult()
```
KaiTeuber, line 91: The ViewModel is built after super.init and started here. Maybe you could use a method SetupViewModel to combine both here?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/NewHome/HomeTableViewController.swift
nickguendling: I think it should be the responsibility of the caller coming from another thread and calling the ui code.

KaiTeuber: The caller doesn't know if a dispatch is required or not.

nickguendling: I would say the caller is the only one that knows 😄

src/xcode/ENA/ENA/Source/Scenes/NewHome/HomeTableViewController.swift
nickguendling: Neither are setting up the view model. One place is binding the view to the view model and this on is triggering an update of the state, which is updating the view model. I don't see a relation between them.

src/xcode/ENA/ENA/Source/Scenes/NewHome/HomeTableViewController.swift
KaiTeuber: The dispatch to the main queue is required because inside of reload() `tableView.reloadData()` gets called.
The called can't know that the method does work that requires a main thread. With taht solution the caller needs to know internal things about the method 'reload()'
I think there are two options herer:
* move `tableView.reloadData()` inside the dispatch of the sink and delete the method `reload()`
* move the dispatch main inside the the method `reload()`so it can get called from whom ever and does the correct thing

nickguendling: Well, I would argue that the HomeTableViewController of course knows the internals of its own implementations and knows that reload triggers UI code. The fact that we both know that `tableView.reloadData()` needs to be called on the main thread shows that Apple handles it the same way, even for their external API. Otherwise they would internally dispatch it to the main queue already, if the followed your proposed approach and my dispatch (that I just reindented here 😅) would not be needed here at all. I think my approach is consistent with how Apple handles it in their own framework and thereby would like to keep it.

KaiTeuber: `reload()` is internal, so it can get called from other classes as well. All these callers have to know that they need to make a dispatch to the man thread. I'm quite sure no all of them will know (if any)

KaiTeuber (approved):

PR #1798, opened 2021-01-14
Fix/3781 throttling app config request

pascal-brause:
## Description
The App fetched the app configuration about three or more time. This happened, because we didn't wait with a new call until we get a response from the server.
Artur started resolving this issue by gathering the promises and then resolving them at the end. I added additionally an async barrier queue to ensure, that the fetch is called only once.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3781

## Screenshots
Bildschirmfoto 2021-01-14 um 16 31 37


KaiTeuber (changes requested):
the jira ticket says target should be 1.11 ?

pascal-brause:
> the jira ticket says target should be 1.11 ?

I forgot to update the target version in the jira ticket.

whiskey (approved):
Would be nice to to see the removed tests back soon.
(This is also a self-reminder to myself ;) )

KaiTeuber (approved):

pascal-brause:
> Would be nice to to see the removed tests back soon.
> (This is also a self-reminder to myself ;) )

As I understood, the "new" Tests are doing the same as the old but with the refactored store. So I think, we do not need them anymore.

PR #1800, opened 2021-01-14
updated text on home screen tile

KaiTeuber:
## Description
Text on the home screen tile 'diary' changed
open journal button text changed

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4612

## Screenshots
Simulator Screen Shot - iPhone 12 Pro - 2021-01-14 at 17 54 09
Simulator Screen Shot - iPhone 12 Pro - 2021-01-14 at 17 54 15


janetback (approved):
UA approved

tcfos (approved):

Ein-Tim:
Isn't this text a bit confusing since "wann" implicates that the date/time will be shown, I would better say "wenn Sie Risiko-Begegnungen hatten".

janetback:
HI @Ein-Tim  , thanks for your feedback. 
I have considered your suggestion and also discussed with my UA colleagues and with the PO. Whereas ' wann' can refer to date and time, it can also be a reference to date only. As for replacing with 'wenn' (as in whether there were any risk encounters) - this is something the users already see on the risk card itself. 
I have therefore decided to leave the text as it is.
But thanks for making the suggestion, Janet

PR #1806, opened 2021-01-15
Fix/4602 Delta Onboarding Close Button Doesn't Work

tcfos:
## Description
The delta onboarding screen appears but can't be closed.

**Discussion**
In HomeTableViewController.viewDidLoad() we do
`NotificationCenter.default.addObserver(self, selector: #selector(refreshUIAfterResumingFromBackground),
name: UIApplication.didBecomeActiveNotification, object: nil)
`
which triggers also the Delta Onboarding when invoked.

Additionally the Delta onboarding is called in `HomeTableViewController.viewDidAppear() `
This leads to two instances of the deltaOnboardingCoordinator. As a consequence the close button on the DeltaOnboaring screen doesn't work any more.

There are two obvious possible solutions:
1. remove `showDeltaOnboardingAndAlertsIfNeeded()` from viewDidAppear()
2. use a flag to ensure that `showDeltaOnboardingIfNeeded()` is called only once

To avoid possible side effects option 2 has been implemented. Any other opinions?

## UITest
Displaying delta onboarding and returning to the Home Screen is tested.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4602

## Screenshots
n/a


SwissMountainFrog (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -429,6 +430,13 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 	}
 
 	private func showDeltaOnboardingIfNeeded(completion: @escaping () -> Void = {}) {
+		switch deltaOnboardingDone {
```
nickguendling, line 433: I think this was done by the delta onboarding internally, is that not the case?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
tcfos: No, I didn't find anything.

nickguendling (approved):

PR #1810, opened 2021-01-19
fix no new risk calculation in automatic mode

30mar:
## Description
In automatic mode, if 2 phones where next to each other for 30 min, then you switch one phone off, and submit a **Positive test result** with the other phone, then switch the first phone on again and open the app, the app doesn't detect the high risk exposure.
this is because the **detection mode is manual by default** and we should check the mode every time we enter the app from both background or from first load from memory.

PS: the risk exposure interval is 4 hours.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4564

## Screenshots
IMG_0070
IMG_0069

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate.swift
@@ -88,18 +88,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 	}
 
 	func applicationWillEnterForeground(_ application: UIApplication) {
+		let detectionMode = DetectionMode.fromBackgroundStatus()
+		riskProvider.riskProvidingConfiguration.detectionMode = detectionMode
 		checkForRisks()
 		appUpdateChecker.checkAppVersionDialog(for: window?.rootViewController)
 	}
 	
 	private func checkForRisks() {
-		let detectionMode = DetectionMode.fromBackgroundStatus()
-		riskProvider.riskProvidingConfiguration.detectionMode = detectionMode
-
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENA/Source/AppDelegate.swift
```diff
@@ -83,10 +83,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		NotificationCenter.default.addObserver(self, selector: #selector(isOnboardedDidChange(_:)), name: .isOnboardedDidChange, object: nil)
 		NotificationCenter.default.addObserver(self, selector: #selector(backgroundRefreshStatusDidChange), name: UIApplication.backgroundRefreshStatusDidChangeNotification, object: nil)
 
+		checkForRisks()
 		return true
 	}
 
 	func applicationWillEnterForeground(_ application: UIApplication) {
+		checkForRisks()
```
marcussc, line 91: Looks like this will try to trigger the Riskcalculation twice on a fresh Appstart. I think it would make sense to move/copy `let detectionMode = DetectionMode.fromBackgroundStatus()
 		riskProvider.riskProvidingConfiguration.detectionMode = detectionMode		 		 to the 
`	lazy var riskProvider: RiskProvider = {
		let keyPackageDownload = KeyPackageDownload(
			downloadedPackagesStore: downloadedPackagesStore,
			client: client,
			wifiClient: wifiClient,
			store: store
		)`

marcussc (approved):

nickguendling (approved):

PR #1817, opened 2021-01-19
build tests as well

whiskey:
Reminder to check if the tests build.


marcussc:
Not sure, where does this help? On the CI we see non building Tests since the Tests get well, build and executed. 
Is this for testing/convince on local dev machines?

KaiTeuber (approved):

KaiTeuber (approved):

whiskey:
> Not sure, where does this help? On the CI we see non building Tests since the Tests get well, build and executed.
Is this for testing/convince on local dev machines?

@marcussc I had several cases where the tests were not matching the frontend development and thus fail to even build. This tweak is a small reminder to not forget the tests when you change existing code. It also helps to prevent blockers when working into a common branch in parallel.

marcussc (approved):

PR #1821, opened 2021-01-19
Bump to next Version 1.11.0.6

Corona-Warn-App-Technical-User:
Bump to next version 1.11.0.6


marcussc (approved):

whiskey (approved):
*klugscheiß*


Actually it's version `1.11.0 (6)`, as `1.11.0.6` would break our current semantic version scheme.
🤓😉

PR #1827, opened 2021-01-20
Feature/4616 task database change for risklevel per date

pascal-brause:
## Description
These are the changes to store & fetch a new property in the contact diary database. For this we create a new table "RiskLevelPerDate". 
To run the existing tests, we must introduce a versioning of the existing `ContactDiaryStore` and the `DiaryDay` (and also `DiaryStoring` and `DiaryProviding`). The current versions are gathered and set in `DiaryStoringProviding`.
The tests for the migration from V2 to V3 and the tests for the new `ContactDiaryStoreV3` are in a separate task and are Todo.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4616

## Screenshots
comes later


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -52,20 +43,42 @@ class ContactDiaryStore: DiaryStoring, DiaryProviding {
...
-	var diaryDaysPublisher = CurrentValueSubject<[DiaryDay], Never>([])
+	var diaryDaysPublisher = CurrentValueSubject<[DiaryDayV2], Never>([])
 
 	func export() -> Result<String, SQLiteErrorCode> {
 		var result: Result<String, SQLiteErrorCode>?
 
 		databaseQueue.inDatabase { database in
 			Log.info("[ContactDiaryStore] export entries.", log: .localData)
-			var contentHeader = "Kontakte der letzten 14 Tage (%@ - %@)\nDie nachfolgende Liste dient dem zuständigen Gesundheitsamt zur Kontaktnachverfolgung gem. § 25 IfSG."
+			var contentHeader = "Kontakte der letzten 15 Tage (%@ - %@)\nDie nachfolgende Liste dient dem zuständigen Gesundheitsamt zur Kontaktnachverfolgung gem. § 25 IfSG."
```
KaiTeuber, line 81: does this string need to be localized?

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/DiaryStoringProviding.swift
```diff
@@ -55,10 +68,21 @@ protocol DiaryStoring {
 	func close()
 }
 
-protocol DiaryProviding {
+protocol DiaryStoringV3 {
```
whiskey, line 71: Why can't we just extend the existing protocols?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/DiaryDayV2.swift
```diff
@@ -0,0 +1,46 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+struct DiaryDayV2: Equatable {
```
whiskey, line 7: There is actually no need to create a new version. Just update the old struct and add a migration.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/__tests__/ContactDiaryStoreTests.swift
```diff
@@ -632,7 +632,7 @@ class ContactDiaryStoreTests: XCTestCase {
 		}
 
 		let expectedString = """
-			Kontakte der letzten 14 Tage (02.12.2020 - 15.12.2020)
+			Kontakte der letzten 15 Tage (01.12.2020 - 15.12.2020)
```
whiskey, line 635: This will fail in multi-language tests. Please localize OR (simple solution) skip this test if `locale != "de_DE"`.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/ContactDiaryStoreV2.swift
```diff
@@ -7,18 +7,9 @@ import UIKit
...
-
-struct DateProvider: DateProviding {
-	var today: Date {
-		Date()
-	}
-}
 
 // swiftlint:disable:next type_body_length
-class ContactDiaryStore: DiaryStoring, DiaryProviding {
+class ContactDiaryStoreV2: DiaryStoringV2, DiaryProvidingV2 {
```
nickguendling, line 12: I'm just writing this here as a general question: Why do we store the risk levels per day in the contact diary store at all? Aren't they contained in the normal riskCalculationResult? Isn't this just redundant data?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/DiaryStoringProviding.swift
pascal-brause: If we would extend the protocol, all ContactDiaryStores (V1,V2,V3) should implement the method. But we want it only be available for V3.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/DiaryDayV2.swift
pascal-brause: No we can't because otherwise the tests for the migration from V1 to V2 would not run anymore. But we want to keep the tests for migrate from V1 to V2. 
In these tests, a database V1 is created, data is written in there and then the migration is called. This would not work if we only have one struct with the logic in it for V3.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/ContactDiaryStoreV2.swift
KaiTeuber: RiskCalculationResult did not store that data before it was only a step inside the calculation. But the main issue is that we need this information inside the journal on a daily basis.
The tec-spec suggested to store these information as part of the contact journal and automatically delete after 15 days with the rest.

nickguendling: But we do store it now, don't we? I don't see why we should keep and update it redundantly if it's now saved by the risk calculation.
I can't find anything in the tech spec that suggests storing this data with the contact journal, can you help me out?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/DiaryStoringProviding.swift
whiskey: Are we using these in parallel? If yes, why? Else, let's merge this and have one protocol that's versioned via git and not copy/paste.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/StoreV2/DiaryDayV2.swift
whiskey: I had a similar discussion with Artur about this and in my opinion there is no reason to handle it this way. 

The migration should contain only the changes in the database structure. With that at hand, the representation objects like `DiaryDay` doesn't need to contain previous states. Same for any kind of 'logic', i.e. how data is handled within the app.

To cut it short, the current way of handling migrations is cumbersome at best. It does not follow any patterns for data migration I know of and produces tons of redundant code by copy/paste existing. The latter also prevents us from tracking changes in the 'root' class.

PR #1828, opened 2021-01-20
fix titles margins in iOS 12.5

30mar:
## Description
Some screens has large empty space at the top and other screens has almost no margin.
Tested and implemented on iPhone 6 12.5.1
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4661

## Screenshots
IMG_2038
IMG_2037
IMG_2036
IMG_2035


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
```diff
@@ -131,7 +131,13 @@ final class ExposureNotificationSettingViewController: UITableViewController, Ac
...
 
 		guard content.cellType == .euTracingCell else { return }
-
+		
+		if #available(iOS 13, *) {
+			navigationItem.largeTitleDisplayMode = .always
+		} else {
+			navigationItem.largeTitleDisplayMode = .never
+		}
+		navigationItem.largeTitleDisplayMode = .never
```
nickguendling, line 140: Shouldn't this line be gone? Right now it's always set to .never I think.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
30mar: yes you are right

nickguendling (approved):

PR #1832, opened 2021-01-21
[iOS] "No Test Result Notifications" does not work

freshking:
## Description

Fixes a missing check for user settings if presenting a notification is allowed.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4692


marcussc (approved):

marcussc:
@freshking 
Please add Milestone and Label.

src/xcode/ENA/ENA/Source/AppDelegate+ENATaskExecutionDelegate.swift
```diff
@@ -121,6 +121,11 @@ extension AppDelegate: ENATaskExecutionDelegate {
 				// Do not trigger notifications for pending or expired results.
 				Log.info("TestResult pending or expired", log: .api)
 			case .success(let testResult):
+				// Before presenting a local notification, first check if user activated notification setting
+				guard self.store.allowTestsStatusNotification else {
+					Log.info("Successfuly got TestResult: \(testResult.stringValue)", log: .api)
+					return
+				}
```
nickguendling, line 128: I think you could even move this `guard` to the top of this function, because we don't persist the test result, we just fetch it to schedule the notification.

nickguendling (dismissed):

nickguendling (approved):

PR #1841, opened 2021-01-22
Bugfix/repair cached configuration unit tests

pascal-brause:
## Description
Fixes the flaky unit tests for the `CachedAppConfigurationTests`. 

We are observing `supportedCountries()` in our unit tests. Our problem is here, that the onboarding triggers the observers in our unit tests, before the unit tests really started. So the unit tests fail, because they think they are executed but this is wrong.
This PR fixes the problem by waiting for the `onFetchExecution` and after this we start to observe the `supportedCountries()`.

On my opinion, the main problem is still that the onboarding should not be called for any unit test. We should refactor this, but that would be a huge thing.


KaiTeuber (approved):

marcussc:
Can you please create a ticket that we investigate how to make sure that the Onboarding is not triggered in the Unit-Tests?

marcussc (approved):

PR #1844, opened 2021-01-22
Fix/1721 remove trait

KaiTeuber:
## Description
removed accessibility traits on subheadline and image in Risklegend view

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-1721

## Screenshots
Simulator Screen Shot - iPhone 12 Pro - 2021-01-22 at 16 39 35
Simulator Screen Shot - iPhone 12 Pro - 2021-01-22 at 16 39 45
ug


src/xcode/ENA/ENA/Source/Scenes/RiskLegend/RiskLegendViewController.swift
```diff
@@ -79,7 +79,8 @@ class RiskLegendViewController: DynamicTableViewController {
 			.navigationSubtitle(
 				text: AppStrings.RiskLegend.subtitle,
 				insets: insets,
-				accessibilityIdentifier: AccessibilityIdentifiers.RiskLegend.subtitle),
+				accessibilityIdentifier: AccessibilityIdentifiers.RiskLegend.subtitle,
+				accessibilityTraits: .none),
```
nickguendling, line 83: Can you make the same change in the StatisticsInfoViewController? I know it's not mentioned in the ticket, because its new, but it's basically a copy of the RiskLegendViewController, so it probably has the same issue. Or maybe we could just change the default value to .none, because it seems that these are the only two places that use `.navigationSubtitle`.

nickguendling (changes requested):

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/RiskLegend/RiskLegendViewController.swift
KaiTeuber: I'll check and change it - like the idea.

nickguendling (approved):

src/xcode/ENA/ENA/Source/Scenes/RiskLegend/RiskLegendViewController.swift
pascal-brause: And it also works for the UIImage below, which is another bug ticket.  👍

pascal-brause (approved):

PR #1860, opened 2021-01-26
Statistical data are not present after restart of the mobile

freshking:
## Description

Always return statistical data if available and from the current day.

## Link to JIRA

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4736


whiskey (approved):
👍

tcfos (approved):

PR #1883, opened 2021-01-29
Feature/4874-TASK-time check

KaiTeuber:
## Description
extended device time check to 3 different states (correct, assumedCorrect and incorrect)

## Link to Jira
Story: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4778
Task: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4874

## Link to Tech Spec change
https://github.com/corona-warn-app/cwa-app-tech-spec/compare/release/1.11.x..proposal/privacy-preserving-analytics#diff-797fa889d039ff7fc826beaca77b68f4b0a9b5d2f27397a97675d8968e1657a9

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
@@ -5,8 +5,8 @@
 import Foundation
 
 protocol DeviceTimeCheckProtocol {
-	func updateDeviceTimeFlags(serverTime: Date, deviceTime: Date)
-	func resetDeviceTimeFlags()
+	func updateDeviceTimeFlags(serverTime: Date, deviceTime: Date, configUpdateSuccessful: Bool)
+	func resetDeviceTimeFlags(configUpdateSuccessful: Bool)
 
 	var isDeviceTimeCorrect: Bool { get }
 }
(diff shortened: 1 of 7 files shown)


pascal-brause (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
```diff
@@ -24,50 +23,53 @@ final class DeviceTimeCheck: DeviceTimeCheckProtocol {
...
+
+        if store.deviceTimeCheckResult == .correct {
+            store.wasDeviceTimeErrorShown = false
+        }
 	}
 
-	func resetDeviceTimeFlags() {
-		store.isDeviceTimeCorrect = true
+	func resetDeviceTimeFlags(configUpdateSuccessful: Bool) {
+		store.deviceTimeCheckResult = configUpdateSuccessful ? .correct : .assumedCorrect
```
ArturFriesen, line 54: I'm not sure about setting to correct if the date header was missing in the app config request.
I also can't find anything about that in the spec.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
KaiTeuber: good point, I'll ask Max about that - I was a bit confuse about that fact as well

KaiTeuber: Max said we should set it to assume_correct - I will change it.

ArturFriesen (approved):

PR #1890, opened 2021-02-01
Fix/4732 scrolling glitch

nickguendling:
## Description
Fixes scrolling glitches and performance issues on the home screen.

## Links to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4732
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4776


src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -72,17 +72,6 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 			}
 			.store(in: &subscriptions)
 
-		viewModel.state.$statistics
```
ArturFriesen, line 75: Is there another place where the statistics are updated?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -363,6 +361,10 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 	}
 
 	private func riskCell(forRowAt indexPath: IndexPath) -> UITableViewCell {
+		if let riskCell = riskCell {
```
ArturFriesen, line 364: Wouldn't this prevent the risk-cell to update with a new model and a new risk-state?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
nickguendling: The statistics data is updated in the HomeState which triggers the update directly inside the statistics cell, there's no need to update the table view around it. This was necessary when we did not show the cell at all if the statistics data was not there yet. But we later changed it to initially show the cell with a height of 0 and then let it update itself internally, so I removed this code.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
nickguendling: The risk cell is also updated from the HomeState, that's why we only need to set it up once.

ArturFriesen (approved):

marcussc (approved):

PR #1909, opened 2021-02-03
catching errors & added extra logging to some local data related func…

whiskey:
## Description
This PR adds extra logging to some parts of the app that might be affected.

## Link to Jira
[EXPOSUREAPP-4859](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4859)

## Screenshots
n/n


src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -66,54 +66,53 @@ final class CachedAppConfiguration {
 	private var promises = [(Result<CachedAppConfiguration.AppConfigResponse, Never>) -> Void]()
 	private var requestIsRunning: Bool { !promises.isEmpty }
 
+	private static let queue = DispatchQueue(label: "fetchAppConfiguration queue", qos: .userInitiated, attributes: .concurrent)
+
 	private func getAppConfig(with etag: String? = nil) -> Future<AppConfigResponse, Never> {
 		return Future { promise in
+			Self.queue.async(flags: .barrier) {
```
freshking, line 73: Would it be easier to make the `DispatchQueue` serial? Then you wouldn't need add the `.barrier` flag. Just asking.

freshking (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -66,54 +66,53 @@ final class CachedAppConfiguration {
 	private var promises = [(Result<CachedAppConfiguration.AppConfigResponse, Never>) -> Void]()
 	private var requestIsRunning: Bool { !promises.isEmpty }
 
+	private static let queue = DispatchQueue(label: "fetchAppConfiguration queue", qos: .userInitiated, attributes: .concurrent)
```
ArturFriesen, line 69: Please correct me, but in my understanding a .concurrent queue in combination with .barrier for every task in this queue, makes it a serial queue. Why not using a serial queue without .barrier?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -66,54 +66,53 @@ final class CachedAppConfiguration {
...
-			guard !self.requestIsRunning else {
-				Log.debug("Return immediately because request allready running.", log: .appConfig)
 				Log.debug("Append promise.", log: .appConfig)
 				self.promises.append(promise)
-				return
-			}
 
-			Log.debug("Append promise.", log: .appConfig)
-			self.promises.append(promise)
+				func defaultFailureHandler() {
```
ArturFriesen, line 84: Can we move the funcs defaultFailureHandler and resolvePromises outside of the getAppConfig func? getAppConfig is hard to read.

KaiTeuber (approved):

ArturFriesen (approved):
The app config implementation is getting out of hands. We should rethink this pattern.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
freshking: @ArturFriesen Check my comment above. Great minds think alike ;-)

marcussc: No it is not the same. 
Using the .barrier flag, lets the the running "jobs" in the Queue finish. Then the Queue is blocked until the "job" with the` .barrier` flag is completed. So while the job with the .barrier flag is not running the queue is concurrent.

freshking: But in this implementation it makes no difference, right?

marcussc: It makes a difference, since we don't bottle neck the queue while we are adding requests to the cachedappconfig. Those can run from n threads at once.

freshking: But if every operation on the queue is a barrier task then you could just make the queue a serial one. As far as I saw that was the only kind of operation executed on that queue.

marcussc: Oh I didn't read the whole changes, my bad, we had a call about the change and I thought only the resolving of the promises should be with the barrier flag 🤔

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: To be honest, I'm just 'expanding' the scope of the queue to the whole `Future` initializer as a result of a spontaneous meeting 😅

The main issue we have is that the app configuration was fetched multiple times on app launch. If we find a way to fetch that config (or do the fallback to a cached/default one) we would not need such stunts. This would normally my approach to tackle such obscure issues rather than fixing symptoms. Unfortunately it's the last day before release and I'm somewhat forced to _hack_ this.

If you want to explore such a solution ('wait for config on app launch') I'm happy to provide help if needed.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: You wrote `resolvePromises` but sure, I can clean this up.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
whiskey: Fun fact. It's `sync` now.

PR #1926, opened 2021-02-05
Feature/4784 faq link how to get tested

whiskey:
## Description
This PR add a link to faqs regarding the 'red risk card'.
Due to design reasons the link appears as a link in desktop HTML but the whole cell is tappable.

## Link to Jira
[EXPOSUREAPP-4784](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4784)

## Screenshots
Simulator Screen Shot - iPhone 12 mini - 2021-02-05 at 12 29 15
Simulator Screen Shot - iPhone 12 mini - 2021-02-05 at 12 29 19


src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
```diff
@@ -433,6 +434,18 @@ class ExposureDetectionViewModel: CountdownTimerDelegate {
 		])
 	}
 
+	private func faqLinkText(tintColor: UIColor = .enaColor(for: .textTint)) -> NSAttributedString {
```
marcussc, line 437: The link seems to have the wrong color its a dark Blue instead of our tint color.

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
```diff
@@ -127,6 +127,16 @@ final class ExposureDetectionViewController: DynamicTableViewController, Require
 		return cell
 	}
 
+	override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
```
ArturFriesen, line 130: All the other cells are using the .action property of DynamicCell. Cant it be used for the link cell as well?
If not, I think you need to call super here, otherwise maybe the other actions won't work.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
whiskey: I never applied the tint color. Fixed and screenshots are updated.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewController.swift
whiskey: This 'link in text' stuff is according to the design not a normal action. Normally Links are opened right from a UITextView and without any extra code. However, UILabels don't support links in attributed texts so I created this one for this one-off abomination.

I'll check actions but if that turns out to require tons of extra hacks I'd rather go with this and add that missing `super` call.

ArturFriesen: I'm not talking about the 'link in text' stuff.
The DynamicTableViewController is already implementing "didSelectRowAt". It calls the corresponding .action of the dynamic cell at the index path.

whiskey: It's using actions now, please review.

SwissMountainFrog (approved):

ArturFriesen (approved):

PR #1931, opened 2021-02-05
reset OTP token if the risk level changes from high to low.

30mar:
## Description
We should reset the OTP token every time the risk level changes from high to low

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4990

## Screenshots
No UI involved

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
@@ -11,8 +11,31 @@ import OpenCombine
 // swiftlint:disable:next type_body_length
 class ExposureDetectionViewModelTests: XCTestCase {
 
+	private var otpService: OTPServiceProviding!
+	private var ppacToken: PPACToken!
+	private var store: MockTestStore!
+	private var client: Client!
+	private var riskExpectation: XCTestExpectation?
+	
+	override func setUp() {
+		super.setUp()
+		
+		self.store = MockTestStore()
+		self.client = ClientMock()
+		self.otpService = OTPService(store: store, client: client)
+		self.ppacToken = PPACToken(apiToken: "apiTokenFake", deviceToken: "deviceTokenFake")
+	}
+	
+	override func tearDown() {
+		otpService = nil
+		ppacToken = nil
+		client = nil
+		store = nil
+		
+		super.tearDown()
+	}
+	
 	func testInitialLowRiskStateWithoutEncounters() {
-		let store = MockTestStore()
 
 		let homeState = HomeState(
 			store: store,
@@ -30,6 +53,7 @@ class ExposureDetectionViewModelTests: XCTestCase {
 		let viewModel = ExposureDetectionViewModel(
 			homeState: homeState,
 			appConfigurationProvider: CachedAppConfigurationMock(),
+			otpService: otpService,
 			onSurveyTap: { _ in },
 			onInactiveButtonTap: { _ in }
 		)
@@ -67,7 +91,6 @@ class ExposureDetectionViewModelTests: XCTestCase {
 	}
 
 	func testLowRiskStateWithEncounters() {
-		let store = MockTestStore()
 
 		let mostRecentDateWithLowRisk = Date()
 		let calculationDate = Date()
@@ -99,6 +122,7 @@ class ExposureDetectionViewModelTests: XCTestCase {
 		let viewModel = ExposureDetectionViewModel(
 			homeState: homeState,
 			appConfigurationProvider: CachedAppConfigurationMock(),
+			otpService: otpService,
 			onSurveyTap: { _ in },
 			onInactiveButtonTap: { _ in }
 		)
@@ -193,9 +217,7 @@ class ExposureDetectionViewModelTests: XCTestCase {
 
 	func testEventSurveyDisabled_cellShouldBeHidden() {
 		var subscriptions = Set<AnyCancellable>()
-		
-		let store = MockTestStore()
-		
+				
 		let mostRecentDateWithHighRisk = Date()
 		let calculationDate = Date()
 		store.riskCalculationResult = RiskCalculationResult(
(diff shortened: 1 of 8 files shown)


src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -343,6 +343,7 @@ final class RiskProvider: RiskProviding {
 			switch risk.level {
 			case .low:
 				store.shouldShowRiskStatusLoweredAlert = true
+				NotificationCenter.default.post(name: .riskStatusLowerd, object: nil)
```
marcussc, line 346: :/ Posting Notifications, is there really no other way to handle this?

30mar: of course there is, but this was the most direct way for me, another way is to inject the risk provider in the exposure detection to detect this. but what is the problem with notifications?

marcussc: Multiple Reasons:

- Notifications are more like a shotgun solution, and should only be used if really everything in the App needs to be notified about something.
- It is very hard to see the Information flow, who did send out this Notification?
- Would be a new way in which we handle Informationflow inside the App, as you mentioned we most of the time use Injection or Combine. 
- One Notification might not seem harmful but image the the whole app using notifications, it just becomes a mess and very hard to debug (race conditions, timing issues, clarity of the Information flow, etc)
- In my personal opinion Notifications are a anti-pattern

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
```diff
@@ -11,8 +11,31 @@ import OpenCombine
...
+	}
+	
+	override func tearDown() {
+		otpService = nil
+		ppacToken = nil
+		client = nil
+		store = nil
+		
+		super.tearDown()
+	}
```
whiskey, line 36: I doubt we need this because we initialize new instances of these properties in `setUp()`.
```suggestion
```

Btw, did you know we can tear down inline within an `addTeardown()` block? I just found out :) https://developer.apple.com/documentation/xctest/xctestcase/understanding_setup_and_teardown_for_test_methods

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
```diff
@@ -637,9 +656,54 @@ class ExposureDetectionViewModelTests: XCTestCase {
 			isLoading: false
 		)
 	}
+	
+	func testIfRishChangeFromHighToLow_thenOTPisReseted() throws {
```
whiskey, line 660: typos
```suggestion
	func testIfRiscChangesFromHighToLow_thenOTPisReseted() throws {
```

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
```diff
@@ -637,9 +656,54 @@ class ExposureDetectionViewModelTests: XCTestCase {
...
+		}
+		// we need the "let" or the viewModel instance will be deallocated when the expectations closure in line 701 returns
+		let viewModel = ExposureDetectionViewModel(
+			homeState: homeState,
+			appConfigurationProvider: CachedAppConfigurationMock(),
+			otpService: otpService,
+			onSurveyTap: { _ in },
+			onInactiveButtonTap: { _ in }
+		)
+		XCTAssertNotNil(viewModel, "View model should not be deallocated")
```
whiskey, line 692: Couldn't we just write… ? Or is the model emitting the `.riskStatusLowerd` notification?
```suggestion
		XCTAssertNotNil(ExposureDetectionViewModel(
		        homeState: homeState,
			appConfigurationProvider: CachedAppConfigurationMock(),
			otpService: otpService,
			onSurveyTap: { _ in },
			onInactiveButtonTap: { _ in }
		)
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
whiskey: They make sense for 'global' stuff like event tracking. This could be a closure passed by the 'listener'.

ArturFriesen: Replace notification with a risk publisher in the homeState of the ExposureDetectionViewModel.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
ArturFriesen: Removed it and nice find!

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
ArturFriesen: Fixed

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/ExposureDetectionViewModelTests.swift
ArturFriesen: The viewModel is system under test and resets the OTP (which is tested here)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: Need to refactor even this. Because the ExposureDetectionViewModel is not allocated all the time and we could miss a risk change.

ArturFriesen:
I close this PR, because we need a completely other approach and a new review.

PR #1935, opened 2021-02-07
Check if the OTP is authorized then navigate accordingly.

30mar:
## Description
Check if the OTP is authorized then navigate directly to the browser, if not then navigate to the consent screen.

_Small note_: the  **surveyOnHighRiskURL** was changed from an optional to a non-optional because the protobuf parameter **surveyOnHighRiskURL** is not an optional.
`	private var surveyOnHighRiskURL: String = ""
`
So we either have an empty string or a string.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4989

## Screenshots
No UI involved

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/Survey/SurveyConsentViewModelTests.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/Survey/SurveyConsentViewModelTests.swift
@@ -8,7 +8,7 @@ import XCTest
 class SurveyConsentViewModelTests: XCTestCase {
 
 	func testDynamicTableViewModel() {
-		if let url = URL(string: "https://www.test.de") {
+		if let url = URL(string: "https://www.example.com") {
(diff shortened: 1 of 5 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/__tests__/Survey/SurveyConsentViewModelTests.swift
```diff
@@ -8,14 +8,17 @@ import XCTest
 class SurveyConsentViewModelTests: XCTestCase {
 
 	func testDynamicTableViewModel() {
-		let placeHolderString = "https://www.test.de"
-		let viewModel = SurveyConsentViewModel(urlString: placeHolderString)
+		if let url = URL(string: "https://www.test.de") {
```
whiskey, line 11: No need to nag the Stiftung Warentest servers!
```suggestion
		if let url = URL(string: "https://www.example.com") {
```

ArturFriesen (approved):

whiskey (approved):

PR #1942, opened 2021-02-08
Feature/4919 ppa collect data

naveeddotio:
## Description
Mechanism for the collection of ExposureRiskMetadata and UserMetaData

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4919

## Screenshots
N/A


pascal-brause:
Is that only the PR for storing the data? Because the real collection of the data is missing, isn't it?

src/xcode/ENA/ENA/Source/Models/User/UserMetadata.swift
```diff
@@ -0,0 +1,35 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+struct UserMetadata: Codable {
+	var federalState: String
```
pascal-brause, line 8: Wrong data type. I need for submission the correct `federalState`-case from the `PPAFederalState`-enum.

src/xcode/ENA/ENA/Source/Models/User/UserMetadata.swift
```diff
@@ -0,0 +1,35 @@
...
+	var administrativeUnit: String
```
pascal-brause, line 9: Wrong data type. I need for submission the correct `Int32` identifier of the json where all units are listed. This is the `districtId`.

pascal-brause (changes requested):

src/xcode/ENA/ENA/Source/Models/User/UserMetadata.swift
pascal-brause: Perhaps use this?
https://github.com/corona-warn-app/cwa-app-ios/blob/b3e289caad1d38a556f46efd92c17618e32a43ba/src/xcode/ENA/ENA/Source/Scenes/Datadonation/District.swift

src/xcode/ENA/ENA/Source/Models/User/UserMetadata.swift
pascal-brause: Perhaps use this?
https://github.com/corona-warn-app/cwa-app-ios/blob/b3e289caad1d38a556f46efd92c17618e32a43ba/src/xcode/ENA/ENA/Source/Scenes/Datadonation/District.swift

naveeddotio: Changed it to Int as of districtId, thanks for the hint.

src/xcode/ENA/ENA/Source/Models/User/UserMetadata.swift
naveeddotio: Changed it to FederalStateName as discussed

naveeddotio:
> Is that only the PR for storing the data? Because the real collection of the data is missing, isn't it?

We have two type of data collection:

1) For User Data, the data will be taken from UI fields which is part of the other story. 
2) For RiskExposure, we already have the data in RiskCalculationResult, we just need to store that information before submitting it.

marcussc:
> > Is that only the PR for storing the data? Because the real collection of the data is missing, isn't it?
> 
> We have two type of data collection:
> 
> 1. For User Data, the data will be taken from UI fields which is part of the other story.
> 2. For RiskExposure, we already have the data in RiskCalculationResult, we just need to store that information before submitting it.

@naveeddotio  Which other Story?  My understanding was that this story should collect all the MVP data.

naveeddotio:
> > > Is that only the PR for storing the data? Because the real collection of the data is missing, isn't it?
> > 
> > 
> > We have two type of data collection:
> > 
> > 1. For User Data, the data will be taken from UI fields which is part of the other story.
> > 2. For RiskExposure, we already have the data in RiskCalculationResult, we just need to store that information before submitting it.
> 
> @naveeddotio Which other Story? My understanding was that this story should collect all the MVP data.

4788-Privacy-preserving Analytics: Onboarding & Settings, task 4996 datadonation screen

naveeddotio:
> > > Is that only the PR for storing the data? Because the real collection of the data is missing, isn't it?
> > 
> > 
> > We have two type of data collection:
> > 
> > 1. For User Data, the data will be taken from UI fields which is part of the other story.
> > 2. For RiskExposure, we already have the data in RiskCalculationResult, we just need to store that information before submitting it.
> 
> @naveeddotio Which other Story? My understanding was that this story should collect all the MVP data.

4788-Privacy-preserving Analytics: Onboarding & Settings, task 4996 datadonation screen

PR #1956, opened 2021-02-10
Feature/5013 analytics submission

pascal-brause:
## Description
The analytics are submitted by this component regarding several checks before the real submission is executed.
I had the clue how to test this, because normally there is no evaluation of the error or the success. So I implemented a solution to get the errors.
The submission is triggered everything the app comes to foreground and every time some data we **could** submit is written into our store.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4789

TechSpec:
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/60c24bea4cb6c8599d6be3befb66ccfb77da4968/docs/spec/data-donation-client-ppa.md#submission-of-usage-data


src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -0,0 +1,321 @@
...
+	) {
+		self.store = store
+		self.client = client
+		self.configurationProvider = appConfig
+	}
+
+	// MARK: - Protocol PPAnalyticsSubmitting
+
+	func triggerSubmitData() {
+		triggerSubmitData(ppacToken: nil, completion: nil)
```
marcussc, line 33: does that make sense to submit without a token? Or will this changed with an other PR?

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -0,0 +1,321 @@
...
+			return .ageGroup0To29
+		case .ageBetween30And59:
+			return .ageGroup30To59
+		case .age60OrAbove:
+			return .ageGroupFrom60
+		}
+	}
+
+	// swiftlint:disable cyclomatic_complexity
+	private func convertToProto(_ federalState: FederalStateName) -> SAP_Internal_Ppdd_PPAFederalState {
```
marcussc, line 285: Can we do this as an extension on `FederalStateName `? Same for the other `convertToProto` could be something like `func protobuf() ->SAP_Internal_Ppdd_PPAFederalState`

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
pascal-brause: This is just for testing purposes. So when the token is nil, a token is requested to create and authorize. For unit tests, this will fail (Simulator).
So for me it is okay as it is.

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
pascal-brause: yay, I will do

PR #1965, opened 2021-02-11
12.5 Exposure Logging ON/OFF jumps to text-switch-area

freshking:
## Description

Removed all stack views from the view controllers table view cells. They not only cause scroll lag but also layout issues. Also fixed a bug where view controller would not update UI after changing EN state.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5002


src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
```diff
@@ -292,20 +278,45 @@ final class ExposureNotificationSettingViewController: UITableViewController, Ac
 		alert.addAction(UIAlertAction(title: AppStrings.ExposureNotificationSetting.privacyConsentDismissAction, style: .cancel, handler: { action in completionHandler(action) }))
 		self.present(alert, animated: true, completion: nil)
 	}
+	
+	private func bannerCell(for indexPath: IndexPath, in tableView: UITableView) -> UITableViewCell {
+		guard let cell = tableView.dequeueReusableCell(withIdentifier: ReusableCellIdentifier.banner.rawValue, for: indexPath) as? ImageTableViewCell else {
+			return UITableViewCell()
```
ArturFriesen, line 284: Is displaying an empty cell expected behaviour? If not, I would prefer to fatalError here. Because not registering a cell, resulting in a crash here, is a programming error.

src/xcode/ENA/ENA/Source/View Helpers/SeperatorLineLayer.swift
```diff
@@ -0,0 +1,18 @@
...
+
+class SeperatorLineLayer: CAShapeLayer {
+	
+	override init() {
+		super.init()
+		lineWidth = 1
+		strokeColor = UIColor.enaColor(for: .hairline).cgColor
+	}
+	
+	required init?(coder: NSCoder) {
```
ArturFriesen, line 15: Putting "@available(*, unavailable)" would be a nice thing for the compiler, our beloved friend.

ArturFriesen (dismissed):
Highly appreciate that you removed the xibs 🤩
Wondering if and why the stack views are causing performance issues.
Would be nice to use stack views in code to have a cleaner layout code.

freshking:
> Highly appreciate that you removed the xibs 🤩
> Wondering if and why the stack views are causing performance issues.
> Would be nice to use stack views in code to have a cleaner layout code.

@ArturFriesen Stack views have a big performance impact because calculating the layout is very complex. That why it is not recommended to use them in reusable views. Besides that on iOS 12 they are very buggy and regular cause layout bugs. I had to remove quite a few stack views in past bug fixes for iOS 12.

marcussc (approved):

whiskey (approved):

PR #2025, opened 2021-02-18
Feature/task 5080 5090 adapt export

ArturFriesen:
## Description
This PR is about enhancing the contact diary export feature with the new data fields.
There is also a fix for a flaky database test.

## Link to Jira
[5090](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5090)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Model/ContactPersonEncounter.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Model/ContactPersonEncounter.swift
@@ -12,18 +12,51 @@ struct ContactPersonEncounter: Equatable {
 		case none
 		case lessThan15Minutes
 		case moreThan15Minutes
+
+		var description: String {
+			switch self {
+			case .none:
+				return ""
+			case .lessThan15Minutes:
+				return "< 15 Minuten"
+			case .moreThan15Minutes:
+				return "> 15 Minuten"
+			}
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Model/ContactPersonEncounter.swift
```diff
@@ -12,18 +12,51 @@ struct ContactPersonEncounter: Equatable {
...
 		case moreThan15Minutes
+
+		var germanDescription: String {
+			switch self {
+			case .none:
+				return ""
+			case .lessThan15Minutes:
+				return "< 15 Minuten"
+			case .moreThan15Minutes:
+				return "> 15 Minuten"
```
KaiTeuber, line 23: Wouldn't it be better to outsource all German strings to a kind of NoLocalizedAppStrings.swift?
Something like: ```static let lessThen15Minues = "< 15 Minuten"```

KaiTeuber (approved):

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Model/ContactPersonEncounter.swift
ArturFriesen: Was thinking about that, but I hope, that there will be a translated export in the future and we move this strings to the current strings file.

PR #2027, opened 2021-02-18
Fix/5192 close button missing on screens

KaiTeuber:
## Description
X / close button was sometimes hidden
The responsibility for the Close button now lies with the controller.
The close button is now no longer part of the NavigationController.
DismissHandling Protokol erweitert:
 * simple fulfilment is sufficient to deactivate pull to dismiss
 * special dismissHandlingCloseButton added

## Link to Jira
This PR will resolve 2 Jira Issues
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5205
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5192


tcfos (approved):

src/xcode/ENA/ENA/Source/Scenes/DataDonation/DataDonationViewController.swift
```diff
@@ -45,12 +45,6 @@ class DataDonationViewController: DynamicTableViewController, DeltaOnboardingVie
 		finished?()
 	}
 
-	// MARK: - Protocol DismissHandling
```
marcussc, line 48: Question, if we don't implement the Protocol do we still need to conform to it?

KaiTeuber: yes, if we conform to it the default implementation (do not dismiss) will get called

marcussc (approved):

PR #2031, opened 2021-02-18
Feature/5081 app launcher shortcut

whiskey:
## Description
This PR adds an Home Screen Quick Action (a.k.a. shortcut) to the app menu. Users can directly enter the 'add contact diary entry' view via this shortcut.

Under two conditions no shortcut is shown:
1. user onboarding is not done yet
2. user is in submission flow while having a positive test result – _exposure submission > contact diary_

## Link to Jira
[EXPOSUREAPP-5081](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5081)

## Screenshots
(translations are wip)
Simulator Screen Shot - iPhone 12 mini - 2021-02-19 at 00 14 57


src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
```diff
@@ -0,0 +1,147 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import XCTest
+import ExposureNotification
+
+class ENAUITestsQuickActions: XCTestCase {
+
```
ArturFriesen, line 9: New line not needed.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
```diff
@@ -0,0 +1,147 @@
...
+
+	private let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
+	private var cwaBundleDisplayName = "Corona-Warn" // dynamic app name!
```
ArturFriesen, line 12: I don't understand the comment "// dynamic app name!"

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
```diff
@@ -0,0 +1,96 @@
...
+			coordinator.diaryCoordinator?.showCurrentDayScreen()
+		}
+	}
+}
+
+extension RootCoordinator {
+
+	/// Direct access to the tabbar controller
+	fileprivate var tabBarController: UITabBarController? {
+		viewController.children.compactMap({ $0 as? UITabBarController }).first
```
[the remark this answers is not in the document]
whiskey: We have properties all over the place, I didn't want to add one more for currently one use case. IMHO, we should keep functional solutions until we need them in two different scenarios which can't be generalized. Then I'd think about adding more properties.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
```diff
@@ -0,0 +1,96 @@
...
+
+	func application(_ application: UIApplication, performActionFor shortcutItem: UIApplicationShortcutItem, completionHandler: @escaping (Bool) -> Void) {
+		handleShortcutItem(shortcutItem)
+	}
+
+	/// Checks wether the application is launched from a shortcut or not.
+	///
+	/// - Parameter launchOptions: the launch options passed in the app launch (will/did) functions
+	/// - Returns: `false` if the application was launched(!) from a shortcut to prevent further calls to `application(_:performActionFor:completionHandler:)`
+	func handleQuickActions(with launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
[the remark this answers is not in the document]
whiskey: About the timing: it's the ENFramework which needs to be started first, before the rest of the UI bootstraps (which is the real flaw that that might be reviewed). I could listen for changes in the `ENAExposureManagerObserver` to get rid of the timing but basically I don't need to care about the ENF state, so I just wait 'long enough' to have some UI I can go on with.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
whiskey: There was a "TODO:" prefixed previously. It's just a reminder to use the `BundleDisplayName` instead but honestly, we don't change that in this Project so I used the simple solution rather than digging more holes the build & test process might fall into.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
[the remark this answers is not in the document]
ArturFriesen: I think this timing is really fragile and 0.2 is a magic number.
It can fail for many reasons.
Maybe older devices, which needs longer.
Maybe apple changing something internally that takes longer.
Maybe we will add something to showHome, which takes longer.
In my opinion this is fragile.
Would appreciate if the coordinator handles it, after the view was really bootstrapped.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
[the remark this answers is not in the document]
ArturFriesen: Still I would appreciate a clean API without accessing the internals of the coordinator directly.
The logic is taking assumptions which will break silently when we change the coordinator.

If we have a property inside of the coordinator "tabBarController" and a function inside of the coordinator .showCurrentDayScreen(). There will be a clear dependency. And if someone changes the UI from tabBarController to something else, she will clearly see, that there is .showCurrentDayScreen() which is depending on it and will fix it.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
ArturFriesen: It's fine for me, but please either change the comment, so its clear and understandable, or remove it.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
```diff
@@ -0,0 +1,147 @@
...
+
+		let actionButton = springboard.buttons[newDiaryEntryLabel]
+		XCTAssertFalse(actionButton.exists, "Shortcuts should not be available once the user is in submission flow")
+	}
+
+	// MARK: - Install/Uninstall our app
+
+	/// Uninstalling the app manually, if present.
+	///
+	/// No hacks and `.resolve()` involved!
```
ArturFriesen, line 115: Dont understand this comment.

whiskey: I'm just not copying the SO stuff you find if you look up for this use case. Just leave it as it is.

ArturFriesen: Could you please change it.

Because, I still don't get it. What is the context of this comment? What do you want to communicate to other developers? I think this comment confuses me more than it helps me. This is going on in my head if I read it: "No hacks" – ok, good code without a hack. As I would expect; "`.resolve()` involved" – what is .resolve() and why is it mentioned here? Do I need to care about this here? And if yes, why do I need to care about this?

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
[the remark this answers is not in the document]
ArturFriesen: Please feel free to close or resolve this how you like. Its not that important, because the UI-Tests will come to the rescue.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/ENAUITestsQuickActions.swift
whiskey: I found a solution that should be fine for everybody

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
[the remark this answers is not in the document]
whiskey: I'm waiting until a UI is available which the app can apply the shortcut action. The whole bootstrap process is too complex to 'just refactor it' on the go.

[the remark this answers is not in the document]
ArturFriesen: Nice! Thanks! 🚀

ArturFriesen (approved):

KaiTeuber (approved):

naveeddotio (approved):

tcfos (approved):

PR #2053, opened 2021-02-22
Adding new localised string for high risk caused by multiple low risks

naveeddotio:
## Description
Adding localised string to cater high risk cause by multiple low risks

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5082

## Screenshots
NA


ArturFriesen (approved):

marcussc (approved):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -1465,6 +1465,8 @@ sein.";
 
 "ContactDiary_Overview_Risk_Text_2" = "Diese müssen nicht in Zusammenhang mit den von Ihnen erfassten Personen und Orten stehen.";
 
+"ContactDiary_Overview_Risk_Text_3" = "aufgrund von mehreren Begegnungen mit niedrigem Risiko";
```
nickguendling, line 1468: Shouldn't there also be a "." at the end? And maybe we could make it clearer, that this is an alternative to text 1? Maybe we could rename them to something like "ContactDiary_Overview_Risk_Text_StandardCause" (Text1), "ContactDiary_Overview_Risk_Text_LowRiskEncountersCause" (Text3) and "ContactDiary_Overview_Risk_Text_Disclaimer" (Text2)?

naveeddotio: Good catch. Let me change it.

nickguendling (approved):

tcfos (approved):

PR #2064, opened 2021-02-23
Bugfix/1876 voice over jumping in long html strings

30mar:
## Description

- The correct way to implement VoiceOver is to have each accessibility element with an identifier the OS then can manages scrolling between those elements based on the view hierarchy.

- In our case We have a problem of incorrect scrolling in the pages that uses long attributed strings usually from HTML files, since the Voice Over tries to parse the string but since there is no accessibility identifiers for each section , some scrolling problems appear when you move from one element to the other.

- As it is not feasible to split this big string "Legal docs or terms of use" into sections and labels, we can't assign an identifier to each section.

- The OS tries to parse the string and handle the scrolling, it manages to parse and scroll to the correct elements in stackviews but it fails in our dynamicTableView. so the appropriate solution here was to create a new ViewController that handles the big HTML files with a stackView, this way scrolling behavior is correct.

- Please see the attached video

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-1876

## Screenshots
**VoiceOver using dynamicTableView:**
https://user-images.githubusercontent.com/15270737/108854496-3ced9e00-75e8-11eb-9eb4-29864fb636d1.MP4

**VoiceOver using stackViews:**

https://user-images.githubusercontent.com/15270737/108855117-e7fe5780-75e8-11eb-8b02-c57d029ba17c.MP4


src/xcode/ENA/ENA/Source/Scenes/HTMLViewController/HTMLViewController.xib
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ArturFriesen, line 1: Cant read this code. Now I need to open Interface Builder. Already afraid, that it will take 1 bazillion days to load. 🥺🤪 Don't worry, I just don't like Interface Builder anymore.

ArturFriesen (approved):

SwissMountainFrog (approved):

tcfos (approved):
great step forward for accessibility!

PR #2075, opened 2021-02-24
Refactoring analytics data - reading / writing / accessing

pascal-brause:
## Description
The main idea of this refactoring was:
1. to centralize every call to access or write analytics data, which are stored in our big secure store. So far, in many occurrences of the app, we access it by calling `store.someAnalyticsData.someProperty`,
2. have a better overview of our analytics data, so that they are centralized in one file/component,
3. easily removing the data when they need to be deleted (user revokes his consent),
4. Prevent forbidden logging when the user did not give his consent to the data collection.

So I created a static component, the `PPAnalyticsCollector`, who capsulate all access to the analytics data. 
To make sure that nobody else then the collector access the database properties, it implements a separate `Store`-based-protocol, that the `Store` does not fulfill. So if you call `Store.` you will not see the analytics data. For this, I separated the existing protocol into two parts and removed the related properties and moved them into the new protocol.

The usage is straight forward: 
When you want to log for example that the test result was submitted via TeleTAN, you just call `Analytics.log(.keySubmissionMetadata(.submittedWithTeletan(true)))` 

Also i refactored, that the submission of the analytics data is only triggered by the collector. 

Before, some of the analytics data had separate services, these are now deleted and their code is moved into the collector, as they were doing some calculations and in them, they read and write some of the analytics data.

So the collector is the administrator of the analytics data. If you want to add new data, you just have to add it there. If you want to log something, give it to him. If you want to delete some data, just ask the collector. If you want to know something about the analytics data, ask the collector. No longer worries about the legals that if we access some properties directly, has the user already given his consent.

The analytics data we gathered so far had already unit tests, they are scattered in the unit tests of the parts, were these data are collected.

This refactoring also bugfixes a problem, that the ClientMetaData are always empty.
(https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5408)

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5129

## Screenshots
Nothing to see.


tcfos (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
```diff
@@ -120,16 +125,16 @@ class TaskExecutionHandler: ENATaskExecutionDelegate {
 		service.submitExposure { error in
 			switch error {
 			case .noSubmissionConsent:
-				self.updateStoreWithKeySubmissionInBackground(value: false)
+				Analytics.log(.keySubmissionMetadata(.submittedInBackground(false)))
```
naveeddotio, line 128: I would rename log to collect or store as the word "log" is misleading in this context.

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
```diff
@@ -285,20 +277,36 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
...
+			submittedAfterCancel: false,
+			submittedAfterSymptomFlow: false,
+			lastSubmissionFlowScreen: .submissionFlowScreenUnknown,
+			advancedConsentGiven: self.store.isSubmissionConsentGiven,
+			hoursSinceTestResult: 0,
+			hoursSinceTestRegistration: 0,
+			daysSinceMostRecentDateAtRiskLevelAtTestRegistration: -1,
+			hoursSinceHighRiskWarningAtTestRegistration: -1,
+			submittedWithTeleTAN: true)
+		Analytics.log(.keySubmissionMetadata(.complete(keySubmissionMetadata)))
```
naveeddotio, line 293: complete is more prone to completion, could we rename it to init or fill ?

naveeddotio (changes requested):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
```diff
@@ -0,0 +1,471 @@
...
+			Log.error("I will never submit any analytics data. Could not cast to correct store protocol", log: .ppa)
+			fatalError("I will never submit any analytics data. Could not cast to correct store protocol")
+		}
+		PPAnalyticsCollector.store = store
+		PPAnalyticsCollector.submitter = submitter
+	}
+
+	/// Setup Analytics for testing. The store or the submitter can be nil for testing purposes.
+	static func setupMock(
+		store: (Store & PPAnalyticsData)? = nil,
```
30mar, line 30: #if !RELEASE
....
#endif

30mar (changes requested):

30mar (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
```diff
@@ -220,13 +217,13 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 					self._fakeVerificationAndSubmissionServerRequest()
 				case .success(let token):
 					// because this block is only called in QR submission
-					self.updateStoreWithQRSubmissionSelected()
+					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), token)))
+					Analytics.collect(.keySubmissionMetadata(.submittedWithTeletan(true)))
```
naveeddotio, line 221: This value will be false. This block is only called in QR submission.

naveeddotio (changes requested):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -280,22 +280,24 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
...
+			let environment = store.selectedServerEnvironment
 
 			let newKey = try KeychainHelper().generateDatabaseKey()
 			store.clearAll(key: newKey)
 
 			/// write excluded value back to the 'new' store
 			store.ppacApiToken = ppacAPIToken
-			store.lastAppReset = Date()
+			store.selectedServerEnvironment = environment
+            Analytics.collect(.submissionMetadata(.lastAppReset(Date())))
```
whiskey, line 296: Maybe it's just Github, but:
```suggestion
			store.selectedServerEnvironment = environment
                      Analytics.collect(.submissionMetadata(.lastAppReset(Date())))
```

src/xcode/ENA/ENA/Source/Models/Metadata/Test Result Metadata/__tests__/TestResultMetadataTests.swift
```diff
@@ -47,69 +45,61 @@ class TestResultMetadataServiceTests: XCTestCase {
...
 		XCTAssertEqual(secureStore.testResultMetadata?.testResult, TestResult.positive, "incorrect testResult")
-		
-		/* The date shouldn't be updated if the test result is the same as the old one
-			- hoursSinceTestRegistration if updated should be (24 * 4)
-			- we explicitly set it into 0 in line 81, so we can see the change
-		*/
-		XCTAssertNotEqual(secureStore.testResultMetadata?.hoursSinceTestRegistration, (24 * 4), "incorrect hoursSinceTestRegistration")
+
+		// The date shouldn't be updated if the test result is the same as the old one
+		XCTAssertEqual(secureStore.testResultMetadata?.hoursSinceTestRegistration, (24 * 4), "incorrect hoursSinceTestRegistration")
```
whiskey, line 85: This was previously `XCTAssertNotEqual` is this new check for equality intentional?

src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/__tests__/DataDonationModelTests.swift
```diff
@@ -10,10 +10,11 @@ class DataDonationModelTests: XCTestCase {
 	func testGIVEN_ModelWithJson_WHEN_GetValues_THEN_ValuesAreEqualToDefault() throws {
 		// GIVEN
 		let mockStore = MockTestStore()
-
+		mockStore.isPrivacyPreservingAnalyticsConsentGiven = false
```
whiskey, line 13: It would be eve nicer if this is the default setting. We could then check on new (Mock-)Stores if this is is the case:
```
let mockStore = MockTestStore()
XCTAssertFalse(mockStore. isPrivacyPreservingAnalyticsConsentGiven)
```

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
```diff
@@ -0,0 +1,475 @@
...
+		store?.previousRiskExposureMetadata = nil
+		store?.userMetadata = nil
+		store?.lastSubmittedPPAData = nil
+		store?.lastAppReset = nil
+		store?.lastSubmissionAnalytics = nil
+		store?.clientMetadata = nil
+		store?.testResultMetadata = nil
+		store?.keySubmissionMetadata = nil
+		store?.exposureWindowsMetadata = nil
+		Log.info("Deleted all analytics data in the store", log: .ppa)
```
whiskey, line 85: Having a dedicated analytics table that could be dropped completely would be nicer. I guess there is a reason why that solution didn't made it into the code?

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
```diff
@@ -0,0 +1,475 @@
...
+	static func triggerAnalyticsSubmission(completion: ((Result<Void, PPASError>) -> Void)? = nil) {
+		guard let submitter = submitter else {
+			Log.warning("I cannot submit analytics data. Perhaps i am a mock or setup was not called correctly?", log: .ppa)
+			return
+		}
+		submitter.triggerSubmitData(completion: completion)
+	}
+
+
+	#if !RELEASE
```
whiskey, line 98: I would consider to move all this !RELEASE code out into a separate extension and maybe even another file (`PPAnalyticsCollector+Testing.swift`). This reduces file size and separates production from dev code.

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -79,19 +86,14 @@ final class MockTestStore: Store, AppConfigCaching, PrivacyPreservingProviding {
...
-		didSet {
-			currentRiskExposureMetadata = nil
-			previousRiskExposureMetadata = nil
-			userMetadata = nil
-			lastSubmittedPPAData = nil
-			lastAppReset = nil
-			lastSubmissionAnalytics = nil
-		}
-	}
+	var isPrivacyPreservingAnalyticsConsentGiven: Bool = true
```
whiskey, line 89: Shouldn't we align with the real store and keep the tinfoil attached, i.e. have `false` as the initial value?

src/xcode/ENA/ENA/Source/Workers/Store/Store.swift
```diff
@@ -145,32 +143,14 @@ protocol StatisticsCaching: AnyObject {
 protocol PrivacyPreservingProviding: AnyObject {
 	/// A boolean storing if the user has already confirmed to collect and submit the data for PPA. By setting it, the existing anlytics data will be reset.
 	var isPrivacyPreservingAnalyticsConsentGiven: Bool { get set }
+	// Do not mix up this property with the real UserMetadata in the PPAnalyticsData protocol
+	var userdata: UserMetadata? { get set }
```
whiskey, line 147: Personal preference would be `userData` but that's just me… :)

src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/__tests__/DataDonationModelTests.swift
whiskey: Update: I've seen the default value for the real store. Having two different default values might lead to confusion and invalid tests due to the different behaviour of the stores.

whiskey (changes requested):
I like the changes. Just having 1-2 questions/remarks.

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
pascal-brause: Good point. I will set the value to `false`

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
pascal-brause: jep, only in Github :/

src/xcode/ENA/ENA/Source/Models/Metadata/Test Result Metadata/__tests__/TestResultMetadataTests.swift
pascal-brause: Oh yes there you found a pretty one. I changed it not correctly, because there were missing a line. So the test did something wrong after my changes!

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
pascal-brause: Yes, we dropped that idea because the effort got bigger and bigger and somewhere we wanted to do a cut.
The only functional reason now for complete new table is that noone forgets to add here "his" new property. For this, I wrote a test that secures that.
Benefit of not having a separated table: No need to setup and then migrate the database values from one to the other.

naveeddotio: Yes. It has a reason behind it. It based on difference criteria to empty the specific store as per tech specs.

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
pascal-brause: Good idea. I will have a try.

pascal-brause: Move it into one extension to separate it from production code goes easy and is good idea.
Move it into a separate file need to make private properties internal and this I don't like. So i will only move it into an extension.

src/xcode/ENA/ENA/Source/Workers/Store/Store.swift
pascal-brause: Can we do :)

src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/__tests__/DataDonationModelTests.swift
pascal-brause: Yeah, you are right. I changed the default value to false. Unfortunately, many tests must declare it now explicitly.

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
pascal-brause: See answer above: I changed so the defaults match together

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
whiskey: If the specs are over-specific we should discuss it in future. Imho, specs should provide the constraints in which we decide how to implement it…

whiskey (approved):

PR #2086, opened 2021-02-25
Tab Bar not scalable

freshking:
## Description

Observe content size category changing and adjust tab bar title attributes

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5198

## Screenshots

IMG_2613


nickguendling:
Tab bars are not supposed to be scaled like that, just like navigation bars. They have the long press gesture for accessibility text sizes, I don't think we should work against the system here and just use the tab bar as is and educate the bug reporter 😁

freshking:
> Tab bars are not supposed to be scaled like that, just like navigation bars. They have the long press gesture for accessibility text sizes, I don't think we should work against the system here and just use the tab bar as is and educate the bug reporter 😁

But if it works why not use it? Isn't it better for users with visual impairments?

whiskey (commented):
I agree to Nick's objections, but in the current context our Tabbar is so simple to bend the rules. One note thought, if we adjust the TabBar, people WILL also want scaled UI components like segmented controls, etc. 

So, if we accept this PR we might face further refactorings later!

whiskey (approved):

src/xcode/ENA/ENA/Source/Views/AccessibleTabbarController.swift
```diff
@@ -0,0 +1,46 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import UIKit
+
+final class AccessibleTabbarController: UITabBarController {
```
nickguendling, line 7: I think we should discuss this further. Maybe in the DevCamp. I think we should follow the platform conventions here not scale the content of bars.

nickguendling (changes requested):

ArturFriesen:
>  Isn't it better for users with visual impairments?

Interesting discussion!

I think it isn't even better for users with visual impairments to keep the tabbar SMALL. Because the tabbar does not have dynamic content, it can be learned and remembered very fast by the user. Keeping it small, keeps more space on the screen for other elements, with dynamic content, which cannot be learned and remembered.

freshking:
> > Isn't it better for users with visual impairments?
> 
> Interesting discussion!
> 
> I think it isn't even better for users with visual impairments to keep the tabbar SMALL. Because the tabbar does not have dynamic content, it can be learned and remembered very fast by the user. Keeping it small, keeps more space on the screen for other elements, with dynamic content, which cannot be learned and remembered.

The tab bar itself does not change size. Only the text size adapts to the content size category selected by the user.

ArturFriesen:
> > > Isn't it better for users with visual impairments?
> > 
> > 
> > Interesting discussion!
> > I think it isn't even better for users with visual impairments to keep the tabbar SMALL. Because the tabbar does not have dynamic content, it can be learned and remembered very fast by the user. Keeping it small, keeps more space on the screen for other elements, with dynamic content, which cannot be learned and remembered.
> 
> The tab bar itself does not change size. Only the text size adapts to the content size category selected by the user.

Good point!

src/xcode/ENA/ENA/Source/Views/AccessibleTabbarController.swift
freshking: I think it is the platform convention since it's officially documented by Apple: 

> In iOS v5.0 and later, you can customize the appearance of tab bars by setting item label text attributes using appearance selectors declared by UIBarItem. You can also use the methods listed in Customizing the Item's Appearance. You can customize the appearance of all tab bar items using the appearance proxy (for example, [UITabBarItem appearance]), or just of a single tab bar item.
By default, unselected and selected images are automatically created from the alpha values in the source images. To prevent system coloring, provide images with UIImage.RenderingMode.alwaysOriginal.

nickguendling: You are talking about API availability, I'm talking about platform conventions. Show me one Apple app with a scaling tab bar 😉

freshking: As I noted below the tab bar itself does not scale. Only the title of the tabs.

nickguendling: I understand that. But the platform convention is not to scale it, because there already is an existing accessibility feature built in that covers that. It was even made available for custom UI components in form of the large content viewer starting from iOS 13, It's definitely the recommended way by Apple to do this: https://developer.apple.com/videos/play/wwdc2019/261/

freshking: And that is still available.

nickguendling: BTW the API are linked to was introduced before dynamic type, so it is technically possible to use it with scaling fonts, but was not intended when it was introduced.

nickguendling: I have put it on the agenda for Tuesday :)

PR #2093, opened 2021-02-25
Feature/task 5098 unit tests

nickguendling:
## Description
Adds unit tests for the `DiaryDayEntryCellModel` and the `DiaryDayViewModel`.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5098


tcfos (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Day/__test__/DiaryDayEntryCellModelTest.swift
```diff
@@ -6,92 +6,611 @@ import Foundation
...
+	func testDurationValues() {
+		let cellModel = contactPersonCellModelWithoutEncounter()
+
+		let expectedDurationValues: [DiaryDayEntryCellModel.SegmentedControlValue<ContactPersonEncounter.Duration>] = [
+			DiaryDayEntryCellModel.SegmentedControlValue(
+				title: AppStrings.ContactDiary.Day.Encounter.lessThan15Minutes,
+				value: .lessThan15Minutes
+			),
+			DiaryDayEntryCellModel.SegmentedControlValue(
+				title: AppStrings.ContactDiary.Day.Encounter.moreThan15Minutes,
```
whiskey, line 74: What about exactly 15 Minutes?

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Day/__test__/DiaryDayEntryCellModelTest.swift
nickguendling: I guess you are expected to either meet for 14.9999999 minutes or less and select `Less than 15 Minutes` or for at least 15.0000001 minutes and select `More than 15 minutes` 🤷🏼‍♂️

whiskey: > (Not sure if the viruses watch is that precise 😅)

True 😅
but seriously the states given do not reflect all possibilities. Can we change it to `moreThanorEqual15Minutes`?

nickguendling: The UI only gives the options "less than 15 min" of "more than 15 min", so the options given to the user are accurately represented. I guess that is something you would have to discuss with UX first, why we leave out the "equal 15 minutes" case and don't include it in either of the two brackets. But I hope that shouldn't hinder us from merging this PR for now 😁

whiskey: I would like to see a test case then for this one. Maybe it kills the app and we disclosure all the super-secret information we collect to the public. The tinfoils would hunt us down in such a case 😱

whiskey (approved):

PR #2098, opened 2021-02-26
Using App Launcher if Contact Journal is used for the first time does not open the actual day

freshking:
## Description

Open current day screen after info screen if requested through shortcuts/deep links

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5404


naveeddotio (approved):

nickguendling:
I remember us talking about this in the refinement. We decided that we don't want to jump to the day screen after the info screen, because the user might be confused as they have never seen the overview screen before.

freshking:
> I remember us talking about this in the refinement. We decided that we don't want to jump to the day screen after the info screen, because the user might be confused as they have never seen the overview screen before.

Martin said this is how it's supposed to be. As described in the ticket.

30mar (approved):

nickguendling (approved):

PR #2108, opened 2021-03-01
Fix/infinite loop

pascal-brause:
## Description
This fixes three problems after the analytics refactoring:
1. An infinite loop can occur when submitting the PPA, which triggers a new fetch of the app config, which triggers again a submission and so on. For this, the sink on the app config call has a proof to be called only once.
2. In the submission we call Analytics.collect, which triggers again a submission. To avoid this, we write directly into the store (what the Submitter is allowed to do)
3. UserDataMetadata was Splitter to differ between the UI-Data and the PPA-Data. Now we let the originally stored data and copy the data to the PPA-Data instead to recreate new ones (that would lead that the user looses his entered user data on the data donation screen). 

Jira Ticket (with error log):
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5458


freshking (approved):
Fixes my loop bug

marcussc (approved):

PR #2148, opened 2021-03-05
Chore/hint to plain logging

whiskey:
## Description
For data privacy reasons we need to prevent 'leakage' of information through log statements that are not monitored by us. This PR prevents the usage of `print`, `debugPrint` and `os_log` (with one exception) in production builds. Please use out `Log` handler for all kinds of logging.

The usage of these keywords is allowed in tests and during development. However, for non-test code it will produce a warning which the CI frowns upon.

## Link to Jira
n/a

## Screenshots
n/a

### DIFF
--- a/src/xcode/ENA/ENAUITests/Helper/AccessibilityLabels.swift
+++ b/src/xcode/ENA/ENAUITests/Helper/AccessibilityLabels.swift
@@ -13,27 +13,6 @@ enum AccessibilityLabels {
 		let uiTestBundle = Bundle(for: ENAUITests_01_Home.self)
 		return NSLocalizedString(key, bundle: uiTestBundle, comment: "")
 	}
-	
-	// print labels to console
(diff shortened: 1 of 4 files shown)


naveeddotio (approved):

SwissMountainFrog (approved):

src/xcode/ENA/ENAUITests/Helper/AccessibilityLabels.swift
```diff
@@ -13,27 +13,6 @@ enum AccessibilityLabels {
 		let uiTestBundle = Bundle(for: ENAUITests_01_Home.self)
 		return NSLocalizedString(key, bundle: uiTestBundle, comment: "")
 	}
-	
-	// print labels to console
-	static func printLabels(_ query: XCUIElementQuery) {
```
pascal-brause, line 18: Are these three methods not used in our code?

PR #2151, opened 2021-03-05
Feature/5297 5368 event checkin persistence mock

ArturFriesen:
## Description
Adds the API for Event and Checkin persistence.
Adds mock implementation of the API.


src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Event.swift
```diff
@@ -0,0 +1,23 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+enum EventType {
+	case type1
+	case type2
+	case type3
```
whiskey, line 8: Can you add some comments or make the cases more precise?

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Event.swift
```diff
@@ -0,0 +1,23 @@
...
+}
+
+struct Event {
+
+	// The ID of the event. Note that the ID is generated by the CWA server. It is stored as base64-encoded string of the guid attribute of Protocol Buffer message Event.
```
whiskey, line 13: SwiftDoc
```suggestion
	/// The ID of the event. Note that the ID is generated by the CWA server. It is stored as base64-encoded string of the guid attribute of Protocol Buffer message Event.
```

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Event.swift
```diff
@@ -0,0 +1,23 @@
...
+	let id: String
+	let type: EventType
+	let description: String
+	let address: String
+	let start: Date
+	let end: Date
+	let defaultCheckInLengthInMinutes: Int
+	// The signature of the event (provided by the CWA server). It is stored as a base64-encoded string of the signature attribute of Protocol Buffer message SignedEvent.
```
whiskey, line 21:
```suggestion
	/// The signature of the event (provided by the CWA server). It is stored as a base64-encoded string of the signature attribute of Protocol Buffer message SignedEvent.

```

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Event.swift
ArturFriesen: At the moment not, maybe in the near future.

ArturFriesen:
Spec changed. New implementation needed.

PR #2165, opened 2021-03-08
Feature/5279 5368 event checkin persistence mock

ArturFriesen:
## Description
Add persistence API and its mock implementation.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5368


src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Checkin.swift
```diff
@@ -0,0 +1,29 @@
...
+	var targetCheckinEndDate: Date
+	let createJournalEntry: Bool
+
+	mutating func update(checkinEndDate: Date) {
+		self.checkinEndDate = checkinEndDate
+	}
+
+	mutating func update(targetCheckinEndDate: Date) {
+		self.targetCheckinEndDate = targetCheckinEndDate
+	}
```
nickguendling, line 28: I understand that this makes it easier to update this struct in the mock store. But I think it might be clearer to the developer that you always have to go through the store, if this is not available here. I understand that you would then have to create a new instance in the mock store, as we do for the diary mock store, but I think that would be preferable.

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,91 @@
...
+		var checkin = checkins[id]
+		checkin.update(targetCheckinEndDate: targetCheckinEndDate)
+		checkins.remove(at: id)
+		checkins.insert(checkin, at: id)
+		checkinsPublisher.send(checkins)
+		return .success(())
+	}
+
+	func deleteCheckin(id: Int) -> EventStoring.VoidResult {
+		checkinsPublisher.value.remove(at: id)
```
nickguendling, line 52: Shouldn't this search for the `id` like you did in `deleteTraceLocation(id: String)`?

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,91 @@
...
+		return .success(())
+	}
+
+	func createTraceTimeIntervalMatch(_ match: TraceTimeIntervalMatch) -> EventStoring.IdResult {
+		traceTimeIntervalMatchesPublisher.value.append(match)
+		return .success((traceTimeIntervalMatchesPublisher.value.count - 1))
+	}
+
+	func deleteTraceTimeIntervalMatch(id: Int) -> EventStoring.VoidResult {
+		traceTimeIntervalMatchesPublisher.value.remove(at: id)
```
nickguendling, line 67: Same as above

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,91 @@
...
+		return .success(())
+	}
+
+	func createTraceWarningPackageMetadata(_ match: TraceWarningPackageMetadata) -> EventStoring.IdResult {
+		traceWarningPackageMetadatasPublisher.value.append(match)
+		return .success((traceWarningPackageMetadatasPublisher.value.count - 1))
+	}
+
+	func deleteTraceWarningPackageMetadata(id: Int) -> EventStoring.VoidResult {
+		traceWarningPackageMetadatasPublisher.value.remove(at: id)
```
nickguendling, line 77: Same as above

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	func updateCheckin(id: Int, endDate: Date) -> EventStoring.VoidResult {
+		var checkins = checkinsPublisher.value
+		var checkin = checkins[id]
+		checkin.update(checkinEndDate: endDate)
+		checkins.remove(at: id)
+		checkins.insert(checkin, at: id)
+		checkinsPublisher.send(checkins)
+		return .success(())
```
nickguendling, line 38: I think you can't just use the id as an index here, same for the other `updateCheckIn` method

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,91 @@
...
+
+	func deleteAllTraceLocations() -> EventStoring.VoidResult {
+		traceLocationsPublisher.value = ([])
+		return .success(())
+	}
+
+	func createCheckin(_ checkin: Checkin) -> EventStoring.IdResult {
+		checkinsPublisher.value.append(checkin)
+		return .success((checkinsPublisher.value.count - 1))
+	}
```
nickguendling, line 29: Don't we need a CheckIn struct without an id and some kind of auto increment mechanism?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
nickguendling: I thinks that's true for all other structs except the `TraceLocation`, as there the id comes from the server and is passed in.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/Checkin.swift
ArturFriesen: Yes it results in a lot of boilerplate code. But you point is valid. Changed it.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
ArturFriesen: Its not a string but an integer. The idea is, that the id is the index of the array.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
ArturFriesen: Why not?

ArturFriesen: Yes, didn't thought about deleting. Which would brake the id references. Will change that

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
ArturFriesen: Didn't though about the deletion, which will break the index references. Will change that.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
ArturFriesen: The TraceLocation struct has only a guid as the primary key. And the guid is delivered by the server.
But the Checkin struct needs an id which is the auto incremented primary key from the sql table.

ArturFriesen: Ok, now I got it. So basically. If you want to create a Checkin you can just pass 0 as ID. It will not be used in case of creation.

ArturFriesen: Another solution would be, to have a struct only for creation, without an id.

nickguendling: That's what I had in mind. And we already have the `struct New` in e.g. the `DiaryLocation`, but did we forget to use it there? Looks like it 😅

ArturFriesen: It feels also like a little to much overhead. I would just go with the id and not use it in case of creation.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,179 @@
...
+
+	var traceTimeIntervalMatchesPublisher = OpenCombine.CurrentValueSubject<[TraceTimeIntervalMatch], Never>([])
+
+	var traceWarningPackageMetadatasPublisher = OpenCombine.CurrentValueSubject<[TraceWarningPackageMetadata], Never>([])
+
+	// MARK: - Private
+
+	private func makeUpdatedTraceTimeIntervalMatch(with match: TraceTimeIntervalMatch, id: Int) -> TraceTimeIntervalMatch {
+		TraceTimeIntervalMatch(
+			id: match.id,
```
nickguendling, line 105:
```suggestion
			id: id,
```

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
```diff
@@ -0,0 +1,179 @@
...
+			checkinId: match.checkinId,
+			traceWarningPackageId: match.traceWarningPackageId,
+			traceLocationGUID: match.traceLocationGUID,
+			transmissionRiskLevel: match.transmissionRiskLevel,
+			startIntervalNumber: match.startIntervalNumber,
+			endIntervalNumber: match.endIntervalNumber
+		)
+	}
+
+	private func makeUpdatedTraceWarningPackageMetadata(with metadata: TraceWarningPackageMetadata, id: Int) -> TraceWarningPackageMetadata {
```
nickguendling, line 115: Just a nice-to-have idea: These could be extensions, so one could write `traceWarningPackageMetadata.updatedWith(id: 17)` or something similar, returning a new instance

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
nickguendling: Could be a private extension just here in the mock store

ArturFriesen: 👍Done

src/xcode/ENA/ENA/Source/Scenes/Events/Store/MockEventStore.swift
ArturFriesen: done

nickguendling (approved):

PR #2175, opened 2021-03-09
Fix/5621 submission qr issue

naveeddotio:
## Description
1. Fixing submission via QR issue. 
2. Fixing advance consent issue.
3. Refactoring where needed.
4. Editing tests. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5621

## Screenshots
NA

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
@@ -217,7 +217,7 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 				case .success(let token):
 					// because this block is only called in QR submission
 					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), token)))
-					Analytics.collect(.keySubmissionMetadata(.submittedWithTeletan(false)))
+					self.store.submittedWithQR = true
 					self.store.testRegistrationDate = Date()
 					self._getTestResult(token) { testResult in
 						completion(testResult)
@@ -282,7 +282,7 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 			hoursSinceTestRegistration: 0,
 			daysSinceMostRecentDateAtRiskLevelAtTestRegistration: -1,
 			hoursSinceHighRiskWarningAtTestRegistration: -1,
-			submittedWithTeleTAN: true)
+			submittedWithTeleTAN: !self.store.submittedWithQR)
 		Analytics.collect(.keySubmissionMetadata(.create(keySubmissionMetadata)))
 	}
 
--- a/src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
@@ -217,7 +217,7 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 				case .success(let token):
 					// because this block is only called in QR submission
 					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), token)))
-					self.store.submittedWithQR = true
+					Analytics.collect(.keySubmissionMetadata(.submittedWithQR(true)))
 					self.store.testRegistrationDate = Date()
 					self._getTestResult(token) { testResult in
 						completion(testResult)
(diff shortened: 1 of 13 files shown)


pascal-brause (commented):
1. Is it really necessary to store the property if it was submitted with qr or tan in our secureStore? The new property is only used in these two classes, so wouldn't it be sufficient to hold an local property in the `ExposureSubmissionCoordinator` and the `ExposureSubmissionService` or pass it by or something like that?
2. If you really need that property, we need unit tests.

whiskey (approved):

KaiTeuber (dismissed):

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
```diff
@@ -217,7 +217,7 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 				case .success(let token):
 					// because this block is only called in QR submission
 					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), token)))
-					Analytics.collect(.keySubmissionMetadata(.submittedWithTeletan(false)))
+					self.store.submittedWithQR = true
```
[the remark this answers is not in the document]
naveeddotio: Refactored.

pascal-brause (commented):
1. Hm I don't like to open the Collector to outside. My proposal to solve that would be:
Remove the property `submittedWithTeletan` from the initializer from the metadata object. Then, when creating, we pass it to the collector. There, just right before saving it to the store, we read the property `submittedWithQR` and set it on the property of the created object. 
When you are at the moment, when you know what kind we have (qr or tan), you just call Analytics.collect(<propertyToCollect>).
This has the advantage, that the collector stays "closed" and gives not information outside.
2. We have now 2 bool properties for the submissionType (QR and TAN). At least, that makes in combination 4 stats, what does not reflect reality. Perhaps we should fusion these properties and make an enum `SubmissionType` ( or `TransmissionType` to be more precise) with two cases: TAN and QR.

KaiTeuber:
Now we have two properties 'submittedWithTeleTAN' and 'submittedWithQR' - it would be much better to use an enum
```
enum SubmittedWith
   case qr
   case teleTan
```

Also I don't think we should open up Analytics for reading.

naveeddotio:
> Now we have two properties 'submittedWithTeleTAN' and 'submittedWithQR' - it would be much better to use an enum
> 
> ```
> enum SubmittedWith
>    case qr
>    case teleTan
> ```
> 
> Also I don't think we should open up Analytics for reading.

Removed the second case.

freshking (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
```diff
@@ -284,11 +285,16 @@ enum PPAnalyticsCollector {
 		case let .submittedAfterSymptomFlow(afterSymptomFlow):
 			store?.keySubmissionMetadata?.submittedAfterSymptomFlow = afterSymptomFlow
 		case let .submittedWithTeletan(withTeletan):
```
pascal-brause, line 287: i do not like negations. Why not following:

```suggestion
		case let .submittedWithQr(withQr):
		store?.submittedWithQR = withQr
```

pascal-brause (changes requested):
You removed now the `submittedWithTeleTAN` from the KeySubmissionMetadata. These data are stored in the KV Store. When you delete them and we update the app and a user has given properties, what happens then? Please test this case!

naveeddotio:
> > You removed now the `submittedWithTeleTAN` from the KeySubmissionMetadata. These data are stored in the KV Store. When you delete them and we update the app and a user has given properties, what happens then? Please test this case!
> 
> No, it is not removed from KeySubmissionMetadata, it is only removed from initialisers.



> > You removed now the `submittedWithTeleTAN` from the KeySubmissionMetadata. These data are stored in the KV Store. When you delete them and we update the app and a user has given properties, what happens then? Please test this case!
> 
> No, it is not removed from KeySubmissionMetadata, it is only removed from initialisers. Please check again.

pascal-brause:
> You removed now the `submittedWithTeleTAN` from the KeySubmissionMetadata. These data are stored in the KV Store. When you delete them and we update the app and a user has given properties, what happens then? Please test this case!

oh i see it's my fault. You only removed it from the initializer.

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
naveeddotio: I would rather prefer submittedWithTeletan to align with what we need to send to the api and what we have in the store i.e., submittedWithTan.

pascal-brause (approved):

PR #2180, opened 2021-03-10
Fix the missing windows metadata bug + refactoring

30mar:
## Description
Exposure windows metadata are missing from dev menu and are not sent to the server

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5609

## Screenshots
IMG_0035


marcussc (approved):

freshking (approved):

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -226,7 +229,7 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 		return false
 	}
 
-	private func generatePPACAndSubmitData(completion: ((Result<Void, PPASError>) -> Void)? = nil) {
+	private func generatePPACAndSubmitData(disableExposureWindowsProbability: Bool = false, completion: ((Result<Void, PPASError>) -> Void)? = nil) {
```
pascal-brause, line 232: Do we really want to disable this from a forced submission call from the dev menu?

pascal-brause (commented):
Sorry, I but I don't get what really fixed the bug, that the exposureWindows were not collected?

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
30mar: yes unfortunately, because when you force submit the probability flag is still 0 so the windows will never be added to the submission, only if it is a normal submission from outside the dev menu will everything work fine

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
30mar: how ?

pascal-brause: I though the testers could. But this is not blocking, more a question. Everything is fine.

PR #2188, opened 2021-03-11
loggers for event registration

pascal-brause:
## Description
Bring in two loggers for all how are working on the event registration.
Feel free for better naming proposals

Just for your information: in the tech spec, when talking about the event registration, it is often called "evreg"


src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -29,8 +29,10 @@ extension OSLog {
 	static let survey = OSLog(subsystem: subsystem, category: "survey")
 	/// PP Analytics
 	static let ppa = OSLog(subsystem: subsystem, category: "ppa")
-	/// Events
-	static let checkin = OSLog(subsystem: subsystem, category: "checkin")
+	/// Event registration - checkin / attendees
```
ArturFriesen, line 32: Cant we just call it "Event". Is the "registration" needed?

ArturFriesen: So then we have "EventCheckin" instead of "evregCheckin"

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -29,8 +29,10 @@ extension OSLog {
...
+	static let evregCheckin = OSLog(subsystem: subsystem, category: "evregCheckin")
+	/// Event registration - organizer / client
+	static let evregOrganizer = OSLog(subsystem: subsystem, category: "evregOrganizer")
```
nickguendling, line 35:
```suggestion
	/// Event / Location checkin
	static let checkin = OSLog(subsystem: subsystem, category: "checkin")
	/// Event / Location organizer
	static let traceLocation = OSLog(subsystem: subsystem, category: "traceLocation")
```

My suggestion 😁

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
pascal-brause: Event as name is so potentially misleading, because it also could be an system event or something similar.
So I think Nicks suggestion avoids this problem in a smart way :)

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
pascal-brause: why not :) 👍

ArturFriesen (approved):

nickguendling (approved):

PR #2190, opened 2021-03-11
Daily Contact Journal scrolls up during data maintenance

freshking:
## Description

Strip out all stack views and XIB files and replace with Autolayout code.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5551


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Day/DiaryDayViewController.swift
```diff
@@ -137,21 +137,20 @@ class DiaryDayViewController: UIViewController, UITableViewDataSource, UITableVi
...
 			forCellReuseIdentifier: String(describing: DiaryDayEntryTableViewCell.self)
 		)
 
 		tableView.delegate = self
 		tableView.dataSource = self
 
 		tableView.separatorStyle = .none
-		tableView.rowHeight = UITableView.automaticDimension
-		tableView.estimatedRowHeight = 60
+		tableView.estimatedRowHeight = 1000 // this is need to prevent jumping of scroll position when tableview datasource is reloaded
```
nickguendling, line 153: Interesting find! I just tried it, this line alone fixes the bug. Seems not to be related to the StackViews. If you can confirm that, I would like us to keep the stack views, because they are really convenient to create and update from a development perspective. Or did you observe any other issues with them?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Day/DiaryDayViewController.swift
freshking: We had UI issues from the stack views as we'll

nickguendling: Which other UI issues did we have here?

freshking: Cell wouldn't layout properly. All the views inside the stack view sometimes overlap each other. Especially when stack views are placed inside stack views.

nickguendling: Sounds like that could also be caused by underestimating the row height, maybe that is fixed with this change as well? Can you reproduce it if only this one-liner is implemented?

freshking: Yes, can be reproduced then as well. This one liner only sometimes fixes the jumping issue. It depends on how tall the cell being rendered is. The value 1000 works here, but stops working if changed to a different value. Then sometimes other issues appear such as all the cells are scrolled completely off screen.

nickguendling: That does not really answer my question. All you are saying is that the issue occurs once the estimation is lower than the actual row height.

I checked out your branch, changed the estimated row height back to 60. The bug occurs even with all your other changes. Therefore I conclude that the bug has nothing to do with the stack views and is only caused by the underestimated row height and suggest we keep the old implementation and just overestimate the row height as you did.

(Please don't resolve this conversation every time, I think it should be up to the opener of a discussion to mark it as resolved.)

freshking: The work is done. If you think it's not good enough and don't believe that there were other UI issues as well please delete this PR and take over the linked JIRA ticket. I will move on now.

PR #2213, opened 2021-03-15
Bugfix/5785 no submission for 48h

pascal-brause:
## Description
It seems to me that this `sink` on the appConfiguration in the `Submitter` made trouble when it was in the submission-call itself concatenated with the `cancelable`. I could reproduce that in some cases a submission was not triggered because the `cancelable` was nil.
So I moved the `sink` to the initializer and now the logs look fine. I also added a bunch of more logs and made some more precise.

If this really solved the bug, that there is no submission > 48h, i cannot say but has to be determined by the testers.

I also fixed the collection of the client and user metadata, because according to the tech spec, they should be graped right before the submission.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5785

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
@@ -252,8 +252,13 @@ class TaskExecutionHandler: ENATaskExecutionDelegate {
 	}
 
 	private func executeAnalyticsSubmission(completion: @escaping () -> Void) {
-		Analytics.triggerAnalyticsSubmission(completion: { _ in
-			// Ignore the result of the call, so we just complete after the call is finished.
+		Analytics.triggerAnalyticsSubmission(completion: { result in
+			switch result {
+			case .success:
+				Log.info("[ENATaskExecutionDelegate] Analytics submission was triggered succesfully from background", log: .ppa)
+			case let .failure(error):
+				Log.error("[ENATaskExecutionDelegate] Analytics submission was triggered not succesfully from background", log: .ppa, error: error)
+			}
 			completion()
 		})
 	}
--- a/src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/DataDonationModel.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/DataDonationModel.swift
@@ -16,7 +16,6 @@ struct DataDonationModel {
 		self.isConsentGiven = store.isPrivacyPreservingAnalyticsConsentGiven
 
 		let userMetadata = store.userData
-		Analytics.collect(.userData(.create(userMetadata)))
 		self.federalStateName = userMetadata?.federalState?.rawValue
 		self.age = userMetadata?.ageGroup?.text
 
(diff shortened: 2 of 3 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/DataDonation/Model/DataDonationModel.swift
```diff
@@ -57,7 +56,9 @@ struct DataDonationModel {
 		store.isPrivacyPreservingAnalyticsConsentGiven = isConsentGiven
 
 		// If user has not given or revoked his consent, delete all analytics data. If he gives not the consent, delete all analytics data to have a clean state.
-		Analytics.deleteAnalyticsData()
+		if isConsentGiven == false {
+			Analytics.deleteAnalyticsData()
+		}
 
 		guard isConsentGiven else {
 			region = nil
```
30mar, line 64: we can put the `Analytics.deleteAnalyticsData()`
inside the else of the `guard isConsentGiven` statement

src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
```diff
@@ -252,8 +252,13 @@ class TaskExecutionHandler: ENATaskExecutionDelegate {
...
 
 	private func executeAnalyticsSubmission(completion: @escaping () -> Void) {
-		Analytics.triggerAnalyticsSubmission(completion: { _ in
-			// Ignore the result of the call, so we just complete after the call is finished.
+		Analytics.triggerAnalyticsSubmission(completion: { result in
+			switch result {
+			case .success:
+				Log.info("[ENATaskExecutionDelegate] Analytics submission was triggered succesfully from background", log: .ppa)
+			case let .failure(error):
+				Log.error("[ENATaskExecutionDelegate] Analytics submission was triggered not succesfully from background", log: .ppa, error: error)
```
30mar, line 260: Maybe print the error inside the log statement as well, it will help debugging in the future

30mar (changes requested):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -37,6 +37,14 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 		self.store = store
 		self.client = client
 		self.configurationProvider = appConfig
+		
+		self.configurationProvider.appConfiguration().sink { [ weak self] configuration in
```
ArturFriesen, line 41: I'm concerned, that this sink is only executed once.
The config data could get stale.
Why not executing it every time triggerSubmitData is called like before? (nil check ("if cancellable == nil") should be removed nevertheless)

ArturFriesen (changes requested):

freshking (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
freshking: I believe this sink will always be called because there is a strong reference to the configurationProvider.

pascal-brause: i will have a look on your considerations and will reopen my PR because now I think the original problem was another :(

PR #2216, opened 2021-03-15
Bugfix/5785 no submission for 48h

pascal-brause:
## Description
Removed that cancellable flag and put the sink on the app config in a dispatch group. And first if we leave the completion handler from the sink, we continue with the submission. That all is done in the background to not block something.
Also I added several logs.
And least, I added a submission state to avoid triggering more then one submission at the same time. The submission state is reset at every completion call.

If this really solved the bug, that there is no submission > 48h, i cannot say but has to be determined by the testers.

I also fixed the collection of the client and user metadata, because according to the tech spec they should be graped right before the submission. And also to solve multiple collect calls at the same time.

i removed the collection of the client metadata in the app config completion handler to avoid infinite loops.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5785

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/Metadata/Client Metadata/__tests__/ClientMetadataTests.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Metadata/Client Metadata/__tests__/ClientMetadataTests.swift
@@ -1,0 +1,43 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+@testable import ENA
+import XCTest
+
+class ClientMetadataTests: XCTestCase {
+
+	func testGIVEN_ClientMetadat_WHEN_AnalyticsCollectIsCalled_THEN_ClientMetadataAreLoggedCorrect() {
+		// GIVEN
+		let mockStore = MockTestStore()
+		Analytics.setupMock(store: mockStore)
+		mockStore.isPrivacyPreservingAnalyticsConsentGiven = true
+
+		XCTAssertNil(mockStore.clientMetadata, "Client metadata should be initially nil")
+		let appVersionParts = Bundle.main.appVersion.split(separator: ".")
+		guard appVersionParts.count == 3,
+			  let majorAppVersion = Int(appVersionParts[0]),
+			  let minorAppVersion = Int(appVersionParts[1]),
+			  let patchAppVersion = Int((appVersionParts[2])) else {
+			return
+		}
+		let expectedAppVersion = Version(major: majorAppVersion, minor: minorAppVersion, patch: patchAppVersion)
+		// iOSVersion
+		let expectediosVersion = Version(
+			major: ProcessInfo().operatingSystemVersion.majorVersion,
+			minor: ProcessInfo().operatingSystemVersion.minorVersion,
+			patch: ProcessInfo().operatingSystemVersion.patchVersion
+		)
+		// eTag
+		let expectedETag = "fake"
+		
+		// WHEN
+		Analytics.collect(.clientMetadata(.create(ClientMetadata(etag: expectedETag))))
+	
+		// THEN
+		XCTAssertNotNil(mockStore.clientMetadata, "Client metadata should be not nil")
+		XCTAssertEqual(expectedAppVersion, mockStore.clientMetadata?.cwaVersion, "appVersion not equal clientMetaData appVersion")
+		XCTAssertEqual(expectediosVersion, mockStore.clientMetadata?.iosVersion, "iosVersion not equal clientMetaData iosVersion")
+		XCTAssertEqual(expectedETag, mockStore.clientMetadata?.eTag, "eTag not equal clientMetaData eTag")
+	}
+}
--- a/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
+++ b/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
@@ -249,9 +249,7 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 	}
 
 	private func obtainUsageData(disableExposureWindowsProbability: Bool = false) -> SAP_Internal_Ppdd_PPADataIOS {
-
-		Log.info("Obtaining now all usage data for analytics submission.", log: .ppa)
-
+		Log.info("Obtaining now all usage data for analytics submission...", log: .ppa)
 		let exposureRiskMetadata = gatherExposureRiskMetadata()
 		let userMetadata = gatherUserMetadata()
 		let clientMetadata = gatherClientMetadata()
--- a/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
+++ b/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
@@ -21,7 +21,6 @@ protocol PPAnalyticsSubmitting {
 	#endif
 }
 
-// swiftlint:disable:next type_body_length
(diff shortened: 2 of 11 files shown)


src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -93,7 +93,6 @@ final class CachedAppConfiguration {
 							lastAppConfigFetch: Date(),
 							appConfig: response.config
 						)
-						Analytics.collect(.clientMetadata(.create(ClientMetadata(etag: response.eTag))))
```
ArturFriesen, line 96: 👍

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -36,120 +37,145 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
...
-				self.hoursSinceTestRegistrationToSubmitTestResultMetadata = configuration.privacyPreservingAnalyticsParameters.common.hoursSinceTestRegistrationToSubmitTestResultMetadata
-				self.probabilityToSubmitExposureWindows = configuration.privacyPreservingAnalyticsParameters.common.probabilityToSubmitExposureWindows
-
-				if let token = ppacToken {
-					// Submit analytics data with injected ppac token
-					self.submitData(with: token, completion: completion)
-				} else {
-					self.generatePPACAndSubmitData(completion: completion)
-				}
+		// Sink on the app configuration if something has changed. But do this in background and wait for the result before continue.
```
ArturFriesen, line 70: I'm wondering because the request to submit should be async, why is it needed to operate on the background here?
Can you write in the comment, why it's needed to call this on the background?

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -36,120 +37,145 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
...
+			let ppaConfigData = configuration.privacyPreservingAnalyticsParameters.common
+			self?.probabilityToSubmitPPAUsageData = ppaConfigData.probabilityToSubmit
+			self?.hoursSinceTestResultToSubmitKeySubmissionMetadata = ppaConfigData.hoursSinceTestResultToSubmitKeySubmissionMetadata
+			self?.hoursSinceTestRegistrationToSubmitTestResultMetadata = ppaConfigData.hoursSinceTestRegistrationToSubmitTestResultMetadata
+			self?.probabilityToSubmitExposureWindows = ppaConfigData.probabilityToSubmitExposureWindows
+			self?.dispatchGroupSubmission.leave()
+		}.store(in: &subscriptions)
+		
+		// If we have the app config, we continue...
+		dispatchGroupSubmission.notify(queue: .global(qos: .background)) { [weak self] in
```
ArturFriesen, line 82: Why is the dispatch group needed? Can't the logic not use be called in the completion of the appconfig sink?

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
pascal-brause: We had already an issue that the main thread was blocked by the analytics submission. So I thought it would be better to do this in the background that the main is never blocked because of submitting the analytics data, what the user is not needed for a working app.

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
pascal-brause: Honestly I though it would give the code a better structure to not do everything in the completion of the appconfig sink.

pascal-brause: i removed the dispatchgroup

src/xcode/ENA/ENA/Source/Models/Metadata/Client Metadata/__tests__/ClientMetadataTests.swift
```diff
@@ -0,0 +1,43 @@
...
+			  let patchAppVersion = Int((appVersionParts[2])) else {
+			return
+		}
+		let expectedAppVersion = Version(major: majorAppVersion, minor: minorAppVersion, patch: patchAppVersion)
+		// iOSVersion
+		let expectediosVersion = Version(
+			major: ProcessInfo().operatingSystemVersion.majorVersion,
+			minor: ProcessInfo().operatingSystemVersion.minorVersion,
+			patch: ProcessInfo().operatingSystemVersion.patchVersion
+		)
```
whiskey, line 30: I think I've seen similar code somewhere else 🤷‍♂️ If that's the case we should create s simple struct to wrap `ProcessInfo().operatingSystemVersion` and all comparisons.

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -361,7 +361,9 @@ extension SecureStore: PrivacyPreservingProviding {
 	var isPrivacyPreservingAnalyticsConsentGiven: Bool {
 		get { kvStore["isPrivacyPreservingAnalyticsConsentGiven"] as Bool? ?? false }
 		set { kvStore["isPrivacyPreservingAnalyticsConsentGiven"] = newValue
-			userData = nil
+			if newValue == false {
```
whiskey, line 364: Swiftlint is not complaining that `kvStore["isPrivacyPreservingAnalyticsConsentGiven"] = newValue` is not positioned in the line below? Interesting.

pascal-brause: no, swiftline is fine here :)

src/xcode/ENA/ENA/Source/Models/Metadata/Client Metadata/__tests__/ClientMetadataTests.swift
pascal-brause: Because this is a high bug ticket, I'd liked not to refactor that because I just moved that code around. But I can create a chore ticket that this will not be forgotten.

whiskey: Fine by me, I did not expect this to happen right now. Sorry for being ambiguous.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/__tests__/PPAnalyticsSubmitterTests.swift
```diff
@@ -56,7 +56,7 @@ class PPAnalyticsSubmitterTests: XCTestCase {
 		XCTAssertNil(store.exposureWindowsMetadata?.newExposureWindowsQueue)
 		
 		/// Since the Date is super precise we have to be fuzzy here, and since we know our CI lets me a lot fuzzy here.
-		let tenSecondsAgo = Calendar.current.date(byAdding: .second, value: -10, to: Date())
+		let tenSecondsAgo = Calendar.current.date(byAdding: .second, value: -20, to: Date())
```
KaiTeuber, line 59: tenSecondsAgo are actually 20

whiskey (approved):

Issue #2203, opened 2021-03-12
OnboardingInfoViewController pushed twice on second next tap 


mozead1996:
## Avoid duplicates

* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: Iphone 8 
- iOS Version: iOS 14.4
- App Version: release/2.0.x

## Describe the bug

A screen pushed twice in OnboardingInfoViewController's navigaitonController stack when click next button  

## Steps to reproduce the issue

1. Open app for first time 
2. Click on 'Let's get started'
3. Click on  'next'
4. Click on 'next' again 

## Expected behaviour 
 Expected behavior is that screen that contains "If you are diagnosed with corona virus " Header pushed only once  
## Possible Fix
I see the issue that completion handler called twice in switch statement in onBoardingViewController 
I will submit a PR 
## Additional context

---
Internal Tracking ID: [EXPOSUREAPP-5771](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5771)

Ein-Tim:
Confirm this bug. Also visible on the 1.15.x branch AND 1.14.x branch (but it's not reproducible on a real iPhone 6s with 1.14.4 installed.

mozead1996:
I submitted a PR with GIF representing the fix  @Ein-Tim

Ein-Tim:
*Disclaimer:
I'm not a community manager or developer.*

@mozead1996 
I can confirm that the fix is working. 👍

dsarkar:
---
Internal Tracking ID: [EXPOSUREAPP-5771](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5771)

Ein-Tim:
#2225 was merged into the `Release 1.15.x` branch and I can confirm that the fix works. @dsarkar Please re-label to `Fix 1.15`.
@mozead1996 I think this issue can be closed now, or?

PR #2217, opened 2021-03-15
removing handleBlutooth calls

mhmdzaid:
## Description
#2203  fix refactoring code and removing unused handleBluttoth
## Link to Jira
https://github.com/corona-warn-app/cwa-app-ios/projects/1#card-56769025


mozead1996:
@whiskey  sorry i deleted the branch by mistake  so I opened a new PR

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -393,16 +393,8 @@ final class OnboardingInfoViewController: UIViewController {
 		case .privacyPage:
 			persistTimestamp(completion: completion)
 		case .enableLoggingOfContactsPage:
-			func handleBluetooth(completion: @escaping () -> Void) {
```
marcussc, line 396: Sorry but I don't get it. As I understood the Issue is that in the Dev and Community Builds, but not in the Production Build(!) a Screen is shown twice during onboarding. How is this fixing this Issue? Feels like the root cause lays somewhere else. Also it seems dangerous to just remove a presentation of an Alert during the onboarding.

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
mozead1996: ok , I got it as a review from @whiskey  , the problem is that there is two calls for completion on in alertForBluetoothOff(completion:  and a second one in handleBluetooth(completion:  itself  so when u push a ViewControlelr  it's pushed twice  @marcussc

Ein-Tim:
I can't say anything to the code but I can confirm that this fixes the problem.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
KaiTeuber: The issue only happens in Debug Mode - so it's not happening in Release build. I'm not sure if removing `handleBluetooth` might be be correct solution here.

mozead1996: I added it again in https://github.com/corona-warn-app/cwa-app-ios/pull/2217/commits/1d6663efe470ccab75a0ad6e005682230d425392 @KaiTeuber

mozead1996: I just wonder why build and test checks not working for me @KaiTeuber

freshking (changes requested):
This is the wrong approach. There was a faulty call of the completion handler in the `MockExposureManager`. And since this is used in the simulator environment it resulted in the bug. Check #2225 for the fix.

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
freshking: @mozead1996 This is the wrong approach. There was a faulty call of the completion handler in the `MockExposureManager`. And since this is used in the simulator environment it resulted in the bug. Check #2225 for the fix.

Ein-Tim:
This PR is now obsolete with the fix in #2225.

marcussc:
Closing this PR since it has been solved with: https://github.com/corona-warn-app/cwa-app-ios/pull/2225

PR #2227, opened 2021-03-16
Feature/5582 contact journal event extension

ArturFriesen:
## Description
This PR extends the contact journal database and store-api with the capability of Event and Checkout.
Also it refactors the contact journal store to use the SecureSQLStore protocol.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5682


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/AddAndEditEntry/__test__/DiaryAddAndEditEntryViewModelTest.swift
```diff
@@ -335,13 +335,13 @@ class DiaryAddAndEditEntryViewModelTest: XCTestCase {
 		let name = "Office"
 		let phone = "+1"
 		let email = "office@test.ui"
-		let result = mockStore.addLocation(name: name, phoneNumber: phone, emailAddress: email)
+		let result = mockStore.addLocation(name: name, phoneNumber: phone, emailAddress: email, traceLocationGUID: nil)
```
marcussc, line 338: Do we have to refactor this to something like:
```suggestion
		let result = mockStore.addLocation(name: name, phoneNumber: phone, emailAddress: email, traceLocationGUIDHash: nil)
```

With the latest proto changes? :/

ArturFriesen: Its not defined in the spec. Will it be changed? I would propose to rename it to "traceLocationId". So the database is not  in charge about what type of id is stored. Just that its some String.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStoreCreation.swift
```diff
@@ -0,0 +1,126 @@
...
+	// MARK: - Internal
+
+	static func make(url: URL? = nil) -> ContactDiaryStore {
+		let storeURL: URL
+
+		if let url = url {
+			storeURL = url
+		} else {
+			storeURL = ContactDiaryStore.storeURL
+		}
```
whiskey, line 51:
```suggestion
			let storeURL = url ?? ContactDiaryStore.storeURL
```
(supposed to be aligned correctly)

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/AddAndEditEntry/__test__/DiaryAddAndEditEntryViewModelTest.swift
ArturFriesen: Renamed to "traceLocationId" but in case of this PR only for the contact journal stuff.
I will do further renaming, for the event stuff, on another PR.

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/AddAndEditEntry/__test__/DiaryAddAndEditEntryViewModelTest.swift
ArturFriesen: @marcussc I think I need to rename it back. We have both now. GUID and hashed GUID in the schema. So it makes sense to distinguish it in the naming.

ArturFriesen: @marcussc The Location table stores the GUID, not the hash. So the guid naming at this point is fine.

ArturFriesen: Since on the TraceLocations table we store both GUID and hashed GUID. It is better to have a precise naming here, because at some point we compare these fields. I have reverted the renaming. So its "traceLocationGUID" again.

marcussc: Okay fine with me, just wanted to make sure the, when I wrote this comment, renaming in the proto was taken care off :)

marcussc (approved):

freshking (approved):

PR #2232, opened 2021-03-17
Dont block Main-Thread when running PPA with many ExposureWindows

30mar:
## Description

**UPDATE**: Confirmed cause of the problem:

- When we submit windows after the first time we try to compare the windows we have with the ones already sent "which we always keep a reference to inside the store"

-  in some extreme cases we already sent 1000 Windows and now we are trying to submit again, every time we compare the hash for a window we fetch the entire 1000 array to append the new item, and since we are still inside the same function  we keep multiplying this number.

so if we have 500 new windows and 1000 old windows the memory will have 1000 * 500 = 500,000 windows inside the function due to the continuous encoding and decoding which leads to the UI block and then to the crash.

**The solution** was just to compare the hashes once the then append the new windows in one step this way we need to call encode twice only and we just 2000 windows "which do not cause a spike in memory or a UI block"

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5852


marcussc (approved):

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -67,14 +67,16 @@ enum Log {
...
 		// Console logging
-		let meta: String = "[\(file):\(line)] [\(function)]"
-
-		// obviously we have to disable swiftline here:
-		// swiftlint:disable:next no_direct_oslog
-		os_log("%{private}@ %{private}@", log: log, type: type, meta, message)
-
-		// Save logs to File. This is used for viewing and exporting logs from debug menu.
-		fileLogger.log(message, logType: type, file: file, line: line, function: function)
+		DispatchQueue.global(qos: .background).async {
```
pascal-brause, line 70: Perhaps doing it on `.utility` and not `.background` to avoid being killed by the os for useless work too early?

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/PPAnalyticsCollector.swift
```diff
@@ -35,7 +35,7 @@ enum PPAnalyticsCollector {
 			return
 		}
 
-		Log.debug("Logging analytics data: \(dataType)", log: .ppa)
+		Log.debug("Logging analytics data: \(dataType.description)", log: .ppa)
```
pascal-brause, line 38: 👌

tcfos (approved):

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -67,14 +67,16 @@ enum Log {
...
-		// obviously we have to disable swiftline here:
-		// swiftlint:disable:next no_direct_oslog
-		os_log("%{private}@ %{private}@", log: log, type: type, meta, message)
-
-		// Save logs to File. This is used for viewing and exporting logs from debug menu.
-		fileLogger.log(message, logType: type, file: file, line: line, function: function)
+		DispatchQueue.global(qos: .utility).async {
+			let meta: String = "[\(file):\(line)] [\(function)]"
+			
+			// obviously we have to disable swiftline here:
```
KaiTeuber, line 73:
```suggestion
			// obviously we have to disable swiftlint here:
```

Ein-Tim:
Will this PR be included in version 2.0? If not the milestone should be modified accordingly. 

Thanks!

marcussc:
> Will this PR be included in version 2.0? If not the milestone should be modified accordingly.
> 
> Thanks!
@Ein-Tim 
We hoped to get it done in 2.0 but it looks like this will come in 2.1

Ein-Tim:
Okay thanks @marcussc.

I would then suggest to change the milestone to v2.1.0 👍

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/Model/PPADataType.swift
```diff
@@ -13,6 +13,25 @@ enum PPADataType {
...
 	case submissionMetadata(PPASubmissionMetadata)
+	
+	var description: String {
+		switch self {
+		case .userData:
+			return "userData"
+		case .riskExposureMetadata:
+			return "riskExposureMetadata"
+		case .clientMetadata:
+			return "clientMetadata"
```
KaiTeuber, line 24: I don't see this case exists in the enum

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsCollector/Model/PPADataType.swift
30mar: @KaiTeuber  I updated the code to 2.1 and it is now removed

KaiTeuber (approved):

PR #2237, opened 2021-03-18
Inconsistent & wrong behaviour of the button to reactivate exposure logging on the risk card in different scenarios

freshking:
## Description

Remove usage of CBCentral to prevent App Store rejection

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5579


src/xcode/ENA/ENA/Source/Scenes/ENSetting/Cells/ActionDetailTableViewCell.swift
```diff
@@ -90,9 +90,17 @@ class ActionDetailTableViewCell: UITableViewCell {
 		if let state = self.state, state == .unknown {
 			delegate?.performAction(action: .askConsent)
 		} else {
-			if let settingsUrl = URL(string: UIApplication.openSettingsURLString),
-				UIApplication.shared.canOpenURL(settingsUrl) {
-				UIApplication.shared.open(settingsUrl, completionHandler: nil)
+			let url: URL?
+			if state == .bluetoothOff {
+				// this will open the os settings
+				url = URL(string: "App-Prefs:root=General")
```
ArturFriesen, line 96: I think this could be rejected.

ArturFriesen (changes requested):
See comment

src/xcode/ENA/ENA/Source/Scenes/ENSetting/Cells/ActionDetailTableViewCell.swift
freshking: If that is the case then we don't have any possibility to open the OS settings and the corresponding ticket should be marked as invalid.

ArturFriesen: We had a similar story before, where it was requested to open the general settings instead of the app settings. We rejected the story because of the risk being rejected.

Ein-Tim:
If Bluetooth is disabled, the button on the home screen behaves like this: 


https://user-images.githubusercontent.com/67682506/111597509-85a30c00-87ce-11eb-8fc3-bf155af15846.MP4

Can‘t this also be done on the detail screen? Because currently, the button there just does nothing… 


https://user-images.githubusercontent.com/67682506/110162566-31bc1e80-7def-11eb-9ddc-758216f27e33.mp4

src/xcode/ENA/ENA/Source/Scenes/ENSetting/Cells/ActionDetailTableViewCell.swift
Ein-Tim: Could we just open the apps settings instead? See https://github.com/corona-warn-app/cwa-app-ios/pull/2237#issuecomment-801742523

PR #2249, opened 2021-03-19
Feature/5680 event details screen UI

30mar:
## Description
Add checking details screen
UPDATE:
1- still missing the formatter.
2-there is a bug with the UIDatepicker while using the **.countDownTimer mode**.
When you try to set a date while the picker is hidden then you show it and try to select a value, the .valueChanged selector is not called for the first time, but for every next time after it.
the solution was to set the date only on the first time we show the picker  "i.e" tap the picker button"
 and not on viewDidLoad or viewDidAppear.
 The same bug appeared to other developers as well:
 https://stackoverflow.com/questions/28295013/uidate-picker-valuechanged-does-not-update-the-first-spin-but-does-every-spin-a

 
 
 ## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5680
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5681

## Screenshots
Simulator Screen Shot - iPhone 12 - 2021-03-22 at 13 24 05


src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
+			secondComponentValues: [0, 15, 30, 45]
+		)
 	}
+	
+	// MARK: - Internal
+	
+	let numberOfComponents = 4
+	
+	@OpenCombine.Published var descriptionLabelTitle: String?
+	@OpenCombine.Published var addressLabelTitle: String?
```
30mar, line 35: just a comment: I'm using this to update the view as i don't want to make the check-in object public. but im not sure which is better in this case.
Because we already have these strings in the object and there is nothing asynchronous actually happening

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewController.xib
```diff
@@ -0,0 +1,375 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ArturFriesen, line 1: Is this view needed?
If yes, could you give it a name?

Bildschirmfoto 2021-03-19 um 11 15 30

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
 
-	// MARK: - Private
+	func pickerView(viewForRow row: Int, forComponent component: Int) -> UIView {
+		let pickerLabel = ENALabel()
+		let weight: UIFont.Weight = (component == 0 || component == 2) ? .regular : .semibold
+		pickerLabel.font = .enaFont(for: .subheadline, weight: weight, italic: false)
+		pickerLabel.adjustsFontForContentSizeCategory = false
+		
+		switch component {
+		case 0:
```
KaiTeuber, line 43: Can we add an enum for the components to make the cases more readable?

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
+			intValue = pickerValues.secondComponentSelectedValue ?? 0
+		default:
+			return "00"
+		}
+		
+		if intValue < 10 {
+			return "0" + String(describing: intValue)
+		} else {
+			return String(describing: intValue)
+		}
```
KaiTeuber, line 119: Better use a number formatter here?

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
+		default:
+			pickerLabel.text = ""
+		}
+		return pickerLabel
+	}
+	
+	func pickerView(_ pickerView: UIPickerView, widthForComponent component: Int) -> CGFloat {
+		switch component {
+		case 0, 2:
+			return pickerView.bounds.width * 0.15
```
ArturFriesen, line 64: Is 0.15 also working for different screensizes?

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TwoComponentsIntegerPicker.swift
```diff
@@ -0,0 +1,22 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+struct TwoComponentsIntegerPicker {
+	var firstComponentSelectedValue: Int?
+	var secondComponentSelectedValue: Int?
+	let firstComponentValues: [Int]
+	let secondComponentValues: [Int]
+
```
KaiTeuber, line 10: MARK: - Init

ArturFriesen (approved):

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
+			components,
+			from: checkin.traceLocationStartDate ?? Date(),
+			to: checkin.traceLocationEndDate ?? Date()
+		)
+		pickerValues = TwoComponentsIntegerPicker(
+			firstComponentSelectedValue: eventDuration.hour,
+			secondComponentSelectedValue: eventDuration.minute,
+			firstComponentValues: Array(0...23),
+			secondComponentValues: [0, 15, 30, 45]
+		)
```
KaiTeuber, line 26: Did you tried to use a DatePicker with mode countDownTimer?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewController.xib
30mar: since this view lies on top of two different color views, the color ENAHairlineContrast will show the background view like this image, so i hade to add an extra view to fix this 
Screenshot 2021-03-19 at 11 24 37

30mar: how it looks without the view
Screenshot 2021-03-19 at 11 29 41

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
30mar: only tried iphone 8 and 12 to be honest, but if it didn't work for small size what should i do? because the picker in the design has less width than the  screen

ArturFriesen: I would just test it on the smallest device. If it works, its fine.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewController.swift
```diff
@@ -13,104 +14,123 @@ class CheckinDetailViewController: UIViewController {
...
+	
+	private func addBorderAndColorToView(_ view: UIView, color: UIColor) {
+		view.layer.borderColor = color.cgColor
+		view.layer.borderWidth = 1
+	}
+	
+	private func updatePickerButtonTitle() {
+		let hours = viewModel.formattedValueFor(component: 0)
+		let minuets = viewModel.formattedValueFor(component: 2)
+		let title = "\(hours):\(minuets)" + " " + AppStrings.Checkin.Details.hoursShortVersion
```
naveeddotio, line 120: minutes

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -11,10 +12,115 @@ final class CheckinDetailViewModel {
...
+			pickerValues.firstComponentSelectedValue = pickerValues.firstComponentValues[row]
+		case 2:
+			pickerValues.secondComponentSelectedValue = pickerValues.secondComponentValues[row]
+		default:
+			break
+		}
+	}
+	
+	func pickerView(numberOfRowsInComponent component: Int) -> Int {
+		switch component {
```
naveeddotio, line 84: Please add enums

naveeddotio (changes requested):

KaiTeuber (approved):

naveeddotio (approved):

janetback (approved):
UA reviewed

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
```diff
@@ -2,19 +2,75 @@
...
+			selectedDurationInMinutes = defaultDuration
+		} else {
+			let eventDuration = Calendar.current.dateComponents(
+				[.minute],
+				from: traceLocation.startDate ?? Date(),
+				to: traceLocation.endDate ?? Date()
+			)
+			// the 0 should not be possible since we expect either the defaultCheckInLengthInMinutes or the start and end dates to be available always
+			selectedDurationInMinutes = eventDuration.minute ?? 0
+		}
```
KaiTeuber, line 27: why it it with traceLocation? Checkin should have all values require.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
30mar: This is what i understood from our call, parsing a QR code produces a TraceLocation Object which we yse for the checking details, then we use this + the user choices for checkout time and the switch value to create the check-in model which we save to the store

KaiTeuber: What is the screen for? Editing a CheckIn or a TranceLocation? If first the model, should be a CheckIn if second, the ViewController name is confusing. You are right the QR scan will resolve to a TraceLocation.
If the screen shows an trace location you should rename it if it shows the checkin, think about creating the checkin before this screen is created and pass it in.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
```diff
@@ -0,0 +1,183 @@
...
+import OpenCombine
+
+class TraceLocationDetailViewController: UIViewController {
+
+	// MARK: - Init
+
+	init(
+		_ traceLocation: TraceLocation,
+		dismiss: @escaping () -> Void,
+		presentCheckins: @escaping () -> Void
```
nickguendling, line 15: As we discussed, I think you can remove this.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/CheckinDetailViewModel.swift
30mar: I renamed the screen to TraceLocationDetailsViewController

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
```diff
@@ -0,0 +1,183 @@
...
+			}
+			.store(in: &subscriptions)
+		
+		viewModel.$initialDuration
+			.sink { [weak self] duration in
+				self?.selectedDuration = duration
+				self?.setupPicker(with: duration ?? 0)
+
+			}
+			.store(in: &subscriptions)
```
nickguendling, line 98: Are these ever updated? If not, you could just assign them once without using combine.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
```diff
@@ -0,0 +1,183 @@
...
+	private func didSelectDuration(datePicker: UIDatePicker) {
+		viewModel.pickerView(didSelectRow: datePicker.date.todaysMinutes)
+	}
+
+	private func addBorderAndColorToView(_ view: UIView, color: UIColor) {
+		view.layer.borderColor = color.cgColor
+		view.layer.borderWidth = 1
+	}
+
+
```
nickguendling, line 146: One newline would be enough 😁

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
```diff
@@ -0,0 +1,183 @@
...
+	@IBAction private func pickerViewTapped(_ sender: Any) {
+		showPickerButton(sender)
+	}
+	
+	@IBAction private func showPickerButton(_ sender: Any) {
+		self.pickerContainerView.isHidden = !self.pickerContainerView.isHidden
+		self.pickerSeparator.isHidden = self.pickerContainerView.isHidden
+		
+		let color: UIColor = pickerContainerView.isHidden ? .enaColor(for: .textPrimary1) : .enaColor(for: .buttonPrimary)
```
nickguendling, line 173:
```suggestion
		let color: UIColor = pickerContainerView.isHidden ? .enaColor(for: .textPrimary1) : .enaColor(for: .textTint)
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+		
+	func setupView() {
+		descriptionLabelTitle = traceLocation.description
+		addressLabelTitle = traceLocation.address
+		initialDuration = selectedDurationInMinutes
+	}
+
+	func getTraceLocationStatus() -> TraceLocationDateStatus? {
```
nickguendling, line 52:
```suggestion
	var traceLocationStatus: TraceLocationDateStatus? {
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,125 @@
...
+		if startDate > Date() {
+			return .notStarted
+		} else if endDate < Date() {
+			return .ended
+		} else {
+			return .inProgress
+		}
+	}
+	
+	func getFormattedString(for component: Calendar.Component) -> String? {
```
nickguendling, line 66:
```suggestion
	func formattedString(for component: Calendar.Component) -> String? {
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,125 @@
...
+			dateFormatter.locale = Locale.current
+			return dateFormatter.string(from: traceLocation.startDate ?? Date())
+		case .hour:
+			return getFormattedHourString(traceLocation.startDate)
+		default:
+			return nil
+		}
+	}
+	
+	func getFormattedHourString(_ date: Date?) -> String? {
```
nickguendling, line 80:
```suggestion
	func formattedHourString(_ date: Date?) -> String? {
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TwoComponentsIntegerPicker.swift
```diff
@@ -0,0 +1,27 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+struct TwoComponentsIntegerPicker {
```
nickguendling, line 5: This is not used anymore I think, can we throw it away?

nickguendling (changes requested):
I tried it on a 1. Gen iPhone SE, the smaller screen we have to support, there are some problems with Dynamic Type. Some Labels are not set to numberOfLines = 0. The Text next to the duration is cut off. And I think the duration picker should go all the way to the sides because there is overlapping text.

Please also try VoiceOver, the first label scrolls the whole screen to the side, and the duration picker showing/hiding might be a bit weird, it's best you try it yourself.

The newly introduced information for past and future events seems to have a different font size than the rest.

And the button has no margin at the bottom on devices with a home button.

And I still see the "3:00 Std." default bug.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
30mar: yeah i left it as we need to reload the checkin tableViewController to fetch the new checkin as well

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
30mar: I wanted to keep the TraceLocation object Private and only accessible from the viewModel, but ok i can add some properties in the class from the init method

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
nickguendling: The reload is triggered from the eventStore as soon as you add the check-in, using a publisher. No need to tell it to update from your view controller :)

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,122 @@
...
+import UIKit
+import OpenCombine
+
+final class TraceLocationDetailViewModel {
+
+	// MARK: - Init
+	
+	init(_ traceLocation: TraceLocation) {
+		self.traceLocation = traceLocation
+	}
```
nickguendling, line 14:
```suggestion
	init(_ traceLocation: TraceLocation) {
		self.traceLocation = traceLocation
		self.locationDescription = traceLocation.description
 		self.locationAddress = traceLocation.address
		
		selectedDurationInMinutes = initialDuration
	}
```

If you move these here, you can keep them in the viewModel and keep the traceLocation private. And you can get rid of `setupViewModel()`. And instead of the `locationInitialDuration` you can just directly use selectedDurationInMinutes.

30mar: so we keep the locationDescription as an internal property in the viewModel ?because we still need to access it from the VC

30mar: it is done now

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,118 @@
...
+		let components = Calendar.current.dateComponents([.hour, .minute], from: date ?? Date())
+		return dateComponentsFormatter.string(from: components)
+	}
+}
+
+enum TraceLocationDateStatus {
+	case notStarted
+	case inProgress
+	case ended
+}
```
KaiTeuber, line 118: This can be part of the ViewModel class 'TraceLocationDetailViewModel'

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,118 @@
...
+//			traceLocationEndDate: traceLocation.endDate,
+//			traceLocationDefaultCheckInLengthInMinutes: traceLocation.defaultCheckInLengthInMinutes,
+//			traceLocationSignature: traceLocation.signature,
+//			checkinStartDate: Date(),
+//			checkinEndDate: endDate ?? Date(),
+//			checkinCompleted: false,
+//			createJournalEntry: shouldSaveToContactJournal
+//		)
+
+		// eventStore.createCheckin(checkin)
```
KaiTeuber, line 97: What about this?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,118 @@
...
+			return .inProgress
+		}
+	}
+	
+	func pickerView(didSelectRow numberOfMinutes: Int) {
+		selectedDurationInMinutes = numberOfMinutes
+		let components = numberOfMinutes.quotientAndRemainder(dividingBy: 60)
+		let date = Calendar.current.date(bySettingHour: components.quotient, minute: components.remainder, second: 0, of: Date())
+		if let hour = formattedHourString(date) {
+			pickerButtonTitle = hour + " " + AppStrings.Checkins.Details.hoursShortVersion
```
nickguendling, line 57: Please use a localizable String like "%@ Std." because we cannot assume that in every language the unit comes after the digits with one space.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
```diff
@@ -0,0 +1,164 @@
...
+	private func setupPicker() {
+		datePickerView.locale = Locale(identifier: "de_DE")
+		datePickerView.datePickerMode = .countDownTimer
+		datePickerView.minuteInterval = 15
+		datePickerView.addTarget(self, action: #selector(didSelectDuration(datePicker:)), for: .valueChanged)
+	}
+	
+	@objc
+	private func didSelectDuration(datePicker: UIDatePicker) {
+		viewModel.pickerView(didSelectRow: datePicker.date.todaysMinutes)
```
nickguendling, line 120: I think the clean way would be to use the `.countDownDuration` of the DatePicker

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -0,0 +1,118 @@
...
+	
+	func formattedString(for component: Calendar.Component) -> String? {
+		switch component {
+		case .day:
+			let dateFormatter = DateFormatter()
+			dateFormatter.dateStyle = .short
+			dateFormatter.locale = Locale.current
+			return dateFormatter.string(from: traceLocation.startDate ?? Date())
+		case .hour:
+			return formattedHourString(traceLocation.startDate)
```
nickguendling, line 69: I think this should also be a date formatter and not a date components formatter to get an adequately formatted localized date string.

nickguendling (changes requested):
I found some more things when I looked at your code to copy some stuff that are similar on the screen that I'm currently working on 😅

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
30mar: this will be done in another ticket, where we actually parse the QR correctly so we save correct values, or should i just save the mocked value ?

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewController.swift
30mar: we will get the values in seconds then we need to divide by 60, i did it anyway

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
30mar: but then we cant get the padded zeros, the property **zeroFormattingBehavior** is only available in DateComponentsFormatter

KaiTeuber (approved):

nickguendling (approved):

PR #2263, opened 2021-03-23
5784 - risk metadata missing after first submission

naveeddotio:
## Description
Risk metadata was missing after first submission and also in the case of force submission because risk calculation was not done in the meanwhile.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5784

## Screenshots
NA


pascal-brause (commented):
Accordingly to the tech spec: "A data structure to represent the values for PPAExposureRiskMetadata is created ad-hoc before submitting usage data. The client does not submit more than one of these data structures."
We should do your two additions just once right before the submission, like we do it for the client and user metadata?

naveeddotio:
> Accordingly to the tech spec: "A data structure to represent the values for PPAExposureRiskMetadata is created ad-hoc before submitting usage data. The client does not submit more than one of these data structures."
> We should do your two additions just once right before the submission, like we do it for the client and user metadata?

If I call 

Analytics.collect(.riskExposureMetadata(.updateRiskExposureMetadata(riskCalculationResult)))

in the PPAAnalyticsSubmitter before the submission, it will create an infinite loop. let's talk about it.

pascal-brause:
> > Accordingly to the tech spec: "A data structure to represent the values for PPAExposureRiskMetadata is created ad-hoc before submitting usage data. The client does not submit more than one of these data structures."
> > We should do your two additions just once right before the submission, like we do it for the client and user metadata?
> 
> If I call
> 
> Analytics.collect(.riskExposureMetadata(.updateRiskExposureMetadata(riskCalculationResult)))
> 
> in the PPAAnalyticsSubmitter before the submission, it will create an infinite loop. let's talk about it.

Ah I see: because you need the hole functionally implemented in the collector. Then it is fine for me as it is.

pascal-brause (approved):

ArturFriesen (approved):

PR #2266, opened 2021-03-23
Fix/5899 ppa hours since test registration are mainly 0

30mar:
## Description
PPA hoursSinceTestRegistration are not recieving the correct values

1- the bug was fixed by sending the correct parameter in the protobuf mapper
2- I also added more in depth tests that cover the **creation, persistence and mapping** of each metadata metric in once cycle so we can:

- compare what we saved compared to what we expected
- compare what was mapped to protobuf compared to what was saved.

NOTE: I completely removed the the **Log(ClientMetadata) parts** because it is only used in one unit test, while we create the actual data in the submitter not in the collector since all the parameters are from the system (example: iOS version, app version, etc
)
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5899


pascal-brause (approved):

pascal-brause (commented):
👌

marcussc (approved):
Great job! Thank you!

PR #2269, opened 2021-03-24
Fix terminology in legal strings (contact journal) - English only

janetback:
## Description
Fixed incorrect terminology in legal strings (use of 'contact diary' instead of 'contact journal').


tcfos (approved):

pascal-brause (approved):

src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.legal.strings
```diff
@@ -48,19 +48,19 @@
 "AutomaticSharingConsent_DescriptionPart5" = "Your test result will then be shown to you.";
 
 
-/* Contact diary information */
+/* Contact jorunal information */
```
marcussc, line 51:
```suggestion
/* Contact journal information */
```

PR #2285, opened 2021-03-26
Feature/5582 event check in UI scan qr code/5850 save checkin to store

30mar:
## Description
-Save Checkin to the event store
-save the new default value for "add checkin to contact diary" switch "be default first time is true"

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5850

## Screenshots
https://user-images.githubusercontent.com/15270737/112648097-7153b480-8e49-11eb-9efd-5dea81cd854c.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
+++ b/src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
@@ -282,21 +282,6 @@ final class SecureStore: Store {
 		get { kvStore["journalWithExposureHistoryInfoScreenShown"] as Bool? ?? false }
 		set { kvStore["journalWithExposureHistoryInfoScreenShown"] = newValue }
 	}
-
-	var checkinInfoScreenShown: Bool {
-		get { kvStore["checkinInfoScreenShown"] as Bool? ?? false }
-		set { kvStore["checkinInfoScreenShown"] = newValue }
-	}
-	
-	var traceLocationsInfoScreenShown: Bool {
-		get { kvStore["traceLocationsInfoScreenShown"] as Bool? ?? false }
-		set { kvStore["traceLocationsInfoScreenShown"] = newValue }
-	}
-	
-	var shouldAddCheckinToContactDiarybyDefault: Bool {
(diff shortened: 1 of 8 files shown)


src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -292,7 +292,12 @@ final class SecureStore: Store {
 		get { kvStore["traceLocationsInfoScreenShown"] as Bool? ?? false }
 		set { kvStore["traceLocationsInfoScreenShown"] = newValue }
 	}
-
+
```
pascal-brause, line 295: perhaps we should introduce here a separate extension with a new protocol for all the event stuff and move this and the property above in it?

30mar: I noticed we have the same for 
```
	var checkinInfoScreenShown: Bool {
		get { kvStore["checkinInfoScreenShown"] as Bool? ?? false }
		set { kvStore["checkinInfoScreenShown"] = newValue }
	}
	
	var traceLocationsInfoScreenShown: Bool {
		get { kvStore["traceLocationsInfoScreenShown"] as Bool? ?? false }
		set { kvStore["traceLocationsInfoScreenShown"] = newValue }
```
	so i just followed it, also artur told me to add it to the general store so i don't know, what do you think ?

KaiTeuber (approved):

ChrisLoes:
Hey @30mar ,

thanks for implementing this! Is there any public documentation/information about the concept of this feature? I've only found this https://github.com/corona-warn-app/cwa-documentation/blob/master/event_registration.md#interoperability-with-other-contact-tracing-apps-draft - but it's only a draft...

thank you!

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
pascal-brause: I like your approach, but not the name. Because if we gather all event registration stuff in that extension, "caching" would be for me misleading.
So perhaps only "EventRegistration"?

pascal-brause (approved):

PR #2294, opened 2021-03-29
Feature/5582 event check in UI scan qr code/6018 implement rounding for automatic checkout time

30mar:
## Description

- We need to round the Event Duration to a number divisible by 15 so we can have 15 min steps in the date picker**, i.e 0, 15, 30 and 45**.
- We need to do this because the in the creation of the event the time is can any value typed by the user
- The rounding is done in the Init() because it will only be executed once before assigning the variables of the ViewModel.
- unit tests will be added in the story unit tests ticket

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6018


src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -112,4 +132,15 @@ final class TraceLocationDetailViewModel {
...
+	private func generateSHA256(_ guid: String) -> Data? {
+		let encoder = JSONEncoder()
+		do {
+			let guidData = try encoder.encode(guid)
+			return guidData.sha256()
+		} catch {
+			Log.error("traceLocationGUID Encoding error", log: .checkin, error: error)
+			return nil
+		}
+	}
```
whiskey, line 145: Why here ant not a utility function/property on `ExposureWindow` so others could use it as well?

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -104,4 +98,11 @@ final class MockTestStore: Store, AppConfigCaching, PrivacyPreservingProviding,
 	var keySubmissionMetadata: KeySubmissionMetadata?
 	var testResultMetadata: TestResultMetadata?
 	var exposureWindowsMetadata: ExposureWindowsMetadata?
+	
+	// MARK: - EventRegistrationCaching
+
+	var checkinInfoScreenShown: Bool = false
+	var traceLocationsInfoScreenShown: Bool = false
+	var shouldAddCheckinToContactDiaryByDefault = true
```
whiskey, line 106: Just a style-comment: either declare the type everywhere or only if no default value is given. I prefer the latter.
```suggestion
	var checkinInfoScreenShown = false
	var traceLocationsInfoScreenShown = false
	var shouldAddCheckinToContactDiaryByDefault = true
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -9,21 +9,33 @@ final class TraceLocationDetailViewModel {
...
+			duration = eventDuration ?? 0
+		}
+		// rounding up to 15
+		let durationStep = 15
+		let reminderMinutes = duration % durationStep
+		if reminderMinutes != 0 {
+			selectedDurationInMinutes = duration + (durationStep - reminderMinutes)
+		} else {
+			selectedDurationInMinutes = duration
 		}
```
whiskey, line 39: > We need to do this because the in the creation of the event the time is can any value typed by the user

So why is this not handled during the creation?

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
30mar: Good point, i will do so

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
30mar: You mean inside TraceLocations struct ?

whiskey: Yes. This reduces clutter in the view model and the location has a nice property 'hash'.

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
whiskey: Better solve the real issue than fixing the symptoms ;)

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
```diff
@@ -104,9 +129,19 @@ struct TraceLocation {
...
+	var guidHash: Data? {
+		let encoder = JSONEncoder()
+		do {
+			let guidData = try encoder.encode(guid)
+			return guidData.sha256()
+		} catch {
+			Log.error("traceLocationGUID Encoding error", log: .checkin, error: error)
+			return nil
+		}
+	}
```
whiskey, line 141: Sorry, missed that one yesterday, there is no need to use the JSON encoder:

```suggestion
	var guidHash: Data {
	    return guid.data(using: .utf8).sha256()
	}
```
Please review the indentation. Tabs suck!

whiskey (approved):
Approving to speed up things but with one suggestion.

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
30mar: done, thanks!

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
```diff
@@ -18,6 +21,29 @@ struct TraceLocation {
 	let cryptographicSeed: Data
 	let cnPublicKey: Data
 
+	var initialTimeForCheckout: Int {
```
nickguendling, line 24:
```suggestion
	var suggestedCheckinLength: Int {
```

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
```diff
@@ -26,10 +52,13 @@ struct TraceLocation {
 		return Date() < endDate
 	}
 	
+	var guidHash: Data? {
```
nickguendling, line 55:
```suggestion
	var idHash: Data? {
```

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -79,35 +68,36 @@ final class TraceLocationDetailViewModel {
...
+			cryptographicSeed: traceLocation.cryptographicSeed,
+			cnPublicKey: traceLocation.cnPublicKey,
+			checkinStartDate: checkinStartDate,
+			checkinEndDate: checkinEndDate,
+			checkinCompleted: false,
+			createJournalEntry: shouldSaveToContactJournal
+		)
+
+		store.shouldAddCheckinToContactDiaryByDefault = shouldSaveToContactJournal
+		 eventStore.createCheckin(checkin)
```
nickguendling, line 100:
```suggestion
		eventStore.createCheckin(checkin)
```

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
```diff
@@ -18,6 +21,29 @@ struct TraceLocation {
...
+				[.minute],
+				from: startDate ?? Date(),
+				to: endDate ?? Date()
+			).minute
+			// the 0 should not be possible since we expect either the defaultCheckInLengthInMinutes or the start and end dates to be available always
+			duration = eventDuration ?? 0
+		}
+		// rounding up to 15
+		let durationStep = 15
+		let reminderMinutes = duration % durationStep
```
nickguendling, line 39:
```suggestion
		let remainderMinutes = duration % durationStep
```

nickguendling (changes requested):
On remark I found in the tech spec: "If storing fails, an error message is presented to the user (technical error message, SQL error). The check-in is aborted." I think the error handling is missing so far. Could be solved in a separate PR :)

nickguendling (approved):

PR #2301, opened 2021-03-29
localized strings for adjust checkin times

KaiTeuber:
## Description
german localized texts for editing checkin times


janetback (commented):
UA reviewed

PR #2307, opened 2021-03-29
Fix Daylight timing tests in testResultMetadata

30mar:
## Description
We have wrong date calculations in some unit tests because of daylight Time Saving

- for **testResultMetadata**: we can just test the functionality by comparing difference in hours then we don't have to deal with this problem.

- for **DiaryOverviewViewModelTest**: the test is failing because we save the date inside DiaryDay struct as a string, and when we re-construct the date object from it we don't take into account the daylight Time Saving.
the solution is to save the date into the riskCalculationResult without the hour offset for the daylight saving, i.e: 23.03.2021 00:00 instead of 23.03.2021 01:00

This only happens when today is in daylight time but the other date is in winter time, if both dates are the same time saving then we don't need this

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6085


src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/__test__/DiaryOverviewViewModelTest.swift
```diff
@@ -194,3 +214,9 @@ class DiaryOverviewViewModelTest: XCTestCase {
 	}
 
 }
+
+extension Bool {
+	static func ^ (left: Bool, right: Bool) -> Bool {
```
ArturFriesen, line 219: What is this doing?

src/xcode/ENA/ENA/Source/Models/Metadata/Test Result Metadata/__tests__/TestResultMetadataTests.swift
```diff
@@ -92,15 +92,16 @@ class TestResultMetadataTests: XCTestCase {
...
 			XCTFail("registration date is nil")
+			return
 		}
-
+		Analytics.collect(.testResultMetadata(.registerNewTestMetadata(registrationDate, "")))
 		Analytics.collect(.testResultMetadata(.updateTestResult(.positive, "")))
 		XCTAssertEqual(secureStore.testResultMetadata?.testResult, TestResult.positive, "incorrect testResult")
-		XCTAssertEqual(secureStore.testResultMetadata?.hoursSinceTestRegistration, (24 * 4), "incorrect hoursSinceTestRegistration")
+		
+		let hours = Calendar.current.dateComponents([.hour], from: registrationDate, to: Date()).hour
```
ArturFriesen, line 103: A question in generel. Would it help to run all the tests on a UTC timebase? You could use a Calendar in UTC timezone instead of the current, which involves timezone and so on.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/__test__/DiaryOverviewViewModelTest.swift
30mar: Forgot to remove it, sorry,

src/xcode/ENA/ENA/Source/Models/Metadata/Test Result Metadata/__tests__/TestResultMetadataTests.swift
30mar: mmm, i think it will, but I'm not sure how big are the changes we need to do since most of the app deals with time

ArturFriesen: I only see a timezone problem in the scope of this test.
By calculating 4 days back with a calendar, this test brings the current timezone into play.

The whole problem is, that this test depends on a timezone which can change, but is comparing with a fixed number of hours.

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/__test__/DiaryOverviewViewModelTest.swift
```diff
@@ -120,7 +131,7 @@ class DiaryOverviewViewModelTest: XCTestCase {
 			numberOfDaysWithLowRisk: 1,
 			numberOfDaysWithHighRisk: 1,
 			calculationDate: today,
-			riskLevelPerDate: [todayMinus5Days: .low],
+			riskLevelPerDate: [todayMinus5DaysAfterCheckingDaylightSaving: .low],
```
ArturFriesen, line 134: The RiskCalculationResult is based on UTC.
The todayString is calculated based on a UTC formatter.
It would be fine to use a UTC calendar here and not to deal with timezone hiccups.

src/xcode/ENA/ENA/Source/Models/Metadata/Test Result Metadata/__tests__/TestResultMetadataTests.swift
```diff
@@ -135,18 +136,17 @@ class TestResultMetadataTests: XCTestCase {
...
-
+		Analytics.collect(.testResultMetadata(.registerNewTestMetadata(registrationDate, "")))
 		Analytics.collect(.testResultMetadata(.updateTestResult(.positive, "")))
 		XCTAssertEqual(secureStore.testResultMetadata?.testResult, TestResult.positive, "incorrect testResult")
-
-		// The the date is updated if the risk results changes e.g from pendong to positive
-		XCTAssertEqual(secureStore.testResultMetadata?.hoursSinceTestRegistration, (24 * 4), "incorrect hoursSinceTestRegistration")
+		
+		// The the date is updated if the risk results changes e.g from pending to positive
+		XCTAssertNotEqual(secureStore.testResultMetadata?.hoursSinceTestRegistration, (24 * 4), "incorrect hoursSinceTestRegistration")
```
ArturFriesen, line 149: Sure about changing evolution from XCTAssertEqual to XCTAssertNotEqual?

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Overview/__test__/DiaryOverviewViewModelTest.swift
```diff
@@ -106,10 +106,12 @@ class DiaryOverviewViewModelTest: XCTestCase {
 		// GIVEN
 		let dateFormatter = ISO8601DateFormatter.contactDiaryUTCFormatter
 
+		// Check if one of the dates is daylight saving time, then there difference will have an offset and we need to regenerate the a generic date without the offset hours
```
ArturFriesen, line 109: Comment still needed?

ArturFriesen (approved):

freshking (approved):

PR #2314, opened 2021-03-30
Feature/5592 native camera checkin

whiskey:
## Description
This PR allows direct scan of an event QR code via the native camera app (or any other QR scanner).

## Caveats
Because the URL domain changed during development, the latest url domain will not be recognized as a valid associated domain. As long as the old domain contains the corresponding `apple-app-site-association` file we can use the old QR codes to test and demo this feature. This hack is realized in the `Route` initializer (see below) and will removed once the server updated.

## Link to Jira
[EXPOSUREAPP-5592](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5592)

## Screenshots
https://user-images.githubusercontent.com/149976/113003053-a7a77180-9172-11eb-81e7-16a886ba8891.mov


marcussc (approved):

tcfos (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -376,6 +380,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
...
+
+		for key in activityDictionary.keys {
+			if let userActivity = activityDictionary[key] as? NSUserActivity,
+			   userActivity.activityType == NSUserActivityTypeBrowsingWeb,
+			   let url = userActivity.webpageURL {
+				return Route(url: url)
+			}
+		}
+
+		return nil
```
KaiTeuber, line 400: why not move it to the Router struct? Somthing like init?(with launchOptions: ...)

naveeddotio (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
whiskey: It would reduce clutter in the AppDelegate but I think it's not the right place. We already have two initializers in `Route` and would create a third one. Also we have to 'soften' up the scope of `Route` by importing UIKit and tailor this initializer to AppDelegates while ignoring the different handling in scenes (yes, I know, we'll never see them again in this project).

Long story short: as much as I would reduce the amount of code in our app delegate, I don't think this is a good part to move out (expect another extension)

KaiTeuber (approved):

PR #2326, opened 2021-03-31
Feature/6089 configure error correction level

naveeddotio:
## Description
According to tech specs:

**Error Correction Level: the error correction mode shall be set according to the value of Configuration Parameter .qrCodeErrorCorrectionLevel.**

Error correction level was missing in the current QR code generation, this PR is about fixing this issue.

## Screenshots
NA


src/xcode/ENA/ENA/Source/Extensions/TraceLocation+GenerateQRCode.swift
```diff
@@ -9,17 +9,17 @@ extension TraceLocation {
 
 	// MARK: - Internal
 	
-	func qrCode(size: CGSize = CGSize(width: 400, height: 400)) -> UIImage? {
+	func qrCode(size: CGSize = CGSize(width: 400, height: 400), qrCodeErrorCorrectionLevel: String = "M") -> UIImage? {
```
30mar, line 12: Can you Please use an enum for this? qrCodeErrorCorrectionLevel is already an enum inside the protobuf we should keep it like this, if you need the string maybe create a property inside the enum to return the corrosponding letter but i think an enum would be enough

src/xcode/ENA/ENA/Source/Services/Event Registration/QRCodePosterTemplate/QRCodeErrorCorrectionLevelProvider.swift
```diff
@@ -0,0 +1,41 @@
...
+	private var subscriptions: Set<AnyCancellable> = []
+
+	func errorCorrectionLevel(
+		appConfigurationProvider: AppConfigurationProviding,
+		onCompletion: @escaping((String) -> Void)
+	) {
+		appConfigurationProvider.appConfiguration().sink { appConfig in
+			let qrCodeErrorCorrectionLevel = appConfig.presenceTracingParameters.qrCodeErrorCorrectionLevel
+			var mappedErrorCorrectionLevel = "M"
+
```
30mar, line 25: this is the same regarding the enum maybe here is a good place to do so.
`var mappedErrorCorrectionLevel: QRCodeErrorCorrectionLevel = .medium`

30mar (changes requested):

src/xcode/ENA/ENA/Source/Extensions/TraceLocation+GenerateQRCode.swift
naveeddotio: Yes, we need string letters; already present enums are meaningless for QR code generation, adding a property with in the enum of pb files will already change them, have to find an alternative.

src/xcode/ENA/ENA/Source/Services/Event Registration/QRCodePosterTemplate/QRCodeErrorCorrectionLevelProvider.swift
naveeddotio: It doesn’t make sense, we need letters and not numbers for QR code generation.

src/xcode/ENA/ENA/Source/Extensions/TraceLocation+GenerateQRCode.swift
pascal-brause: you can declare enums with string property and use the cases as string or at least their raw values

src/xcode/ENA/ENA/Source/Services/Event Registration/QRCodePosterTemplate/QRCodeErrorCorrectionLevelProvider.swift
pascal-brause: you can declare enums with string property and use the cases as string or at least their raw values

pascal-brause: or you could pass it as associated value in the enum case

src/xcode/ENA/ENA/Source/Extensions/TraceLocation+GenerateQRCode.swift
pascal-brause: or you could pass it as associated value in the enum case

src/xcode/ENA/ENA/Source/Services/Event Registration/QRCodePosterTemplate/QRCodeErrorCorrectionLevelType.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+enum MappedErrorCorrectionType: Int {
+	case medium
+	case large
+	case quantile
+	case high
+	
+	var mappedValue: String {
+		switch self {
```
30mar, line 14: nice

30mar (approved):

pascal-brause (approved):
So much better :)

PR #2332, opened 2021-04-01
Fix flaky TraceLocation tests

freshking:
## Description

Fix failing tests by swiping up until button is visible before tapping


marcussc (approved):

pascal-brause (approved):
Good idea. Perhaps we should think about to add this count to the swipe up method itself to prevent other tests to be flaky again

PR #2333, opened 2021-04-01
Fix TraceWarningDiscoveryCall

pascal-brause:
## Description
We must change the handling of nil latest or oldest property in the repsonse of the discovery call.
If we have one of them nil, we create a faked TraceWarningDiscovery that has set its properties as the computed property availablePackagesOnCDN return an empty array so the download knows, that there are no packages to download.
I also added a test for that scenario and fixed one when there is only one package on the cdn.

## Link to Jira
Urgent fix no jira ticket.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
+++ b/src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
@@ -354,14 +354,16 @@ final class HTTPClient: Client {
 							TraceWarningDiscoveryResponse.self,
 							from: body
 						)
+						let eTag = response.httpResponse.value(forCaseInsensitiveHeaderField: "ETag")
+						
 						guard let oldest = decodedResponse.oldest,
 							  let latest = decodedResponse.latest else {
-							Log.error("Failed to get oldest or latest out of decoded response", log: .api)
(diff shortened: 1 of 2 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -354,14 +354,16 @@ final class HTTPClient: Client {
...
 							from: body
 						)
+						let eTag = response.httpResponse.value(forCaseInsensitiveHeaderField: "ETag")
+						
 						guard let oldest = decodedResponse.oldest,
 							  let latest = decodedResponse.latest else {
-							Log.error("Failed to get oldest or latest out of decoded response", log: .api)
-							completion(.failure(.decodingJsonError(response.statusCode)))
+							Log.info("Succesfully discovered that there are no availablePackagesOnCDN", log: .api)
+							// create an false package with latest < oldest, then computed property availablePackagesOnCDN will be empty for the downloading check later.
```
naveeddotio, line 362: create false package

naveeddotio (approved):

30mar (approved):

PR #2349, opened 2021-04-05
Feature/5686 add unit tests for qr code scan

30mar:
## Description
Dependencies needed to be merged to 2.0 first to make this branchs tests work: 

- New default app config file

- PR for fixing the rounding problem

the current target is the branch for the new app config file, as there are dependency in the tests on this file.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5686


src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeScanner/CheckinQRCodeScannerViewController.swift
```diff
@@ -42,17 +58,16 @@ class CheckinQRCodeScannerViewController: UIViewController {
 
 	override func viewWillDisappear(_ animated: Bool) {
 		super.viewWillDisappear(animated)
-		viewModel.deactivateScanning()
+		viewModel?.deactivateScanning()
```
ArturFriesen, line 61: Why is the viewmodel optional?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeScanner/CheckinQRCodeScannerViewController.swift
30mar: So we can access properties within the closure 
Screenshot 2021-04-06 at 10 27 33

Screenshot 2021-04-06 at 10 26 08

ArturFriesen: I understand. I would prefer to have the 2 callbacks as optional parameters of the view model. 
But this ways works too, just a little confusing.

ArturFriesen (approved):

marcussc (approved):

PR #2351, opened 2021-04-06
Fix suggested checkout time max duration value

30mar:
## Description
Fix rounding issue in the **traceLocationDetailsViewController**.
The issue was not related to rounding but to having a trace location with duration more than one day, the maximum checkout time inside the picker as per design is 23:45, so if the event exceeds this time we should use 23:45.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6103


SwissMountainFrog (approved):

src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckInDetail/TraceLocationDetailViewModel.swift
```diff
@@ -25,7 +25,13 @@ final class TraceLocationDetailViewModel {
 		self.locationAddress = traceLocation.address
 		self.locationDescription = traceLocation.description
 		self.shouldSaveToContactJournal = store.shouldAddCheckinToContactDiaryByDefault
-		self.selectedDurationInMinutes = traceLocation.suggestedCheckoutLength
+		// max duration in the picker is 23:45
+		let maxDurationInMinutes = (23 * 60) + 45
```
marcussc, line 29: I like that you displayed it like this, so it is not a magic number but easy to understand 👍

marcussc (approved):

tcfos (approved):

PR #2355, opened 2021-04-06
5585 check in download matching risk calculation/add some unit tests

ArturFriesen:
## Description
Added some happy path unit tests for the RiskProvider.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6105


30mar (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/__tests__/RiskProviderTests.swift
```diff
@@ -9,6 +9,228 @@ import ExposureNotification
...
+			didFailCalculateRiskExpectation.fulfill()
+		}
+		consumer.didChangeActivityState = { _ in
+			didChangeActivityStateExpectation.fulfill()
+		}
+	
+		riskProvider.observeRisk(consumer)
+		
+		// WHEN
+
```
naveeddotio, line 77: Small thing, could we remove the extra spacing after when and then?

ArturFriesen: Sure about that? If I remove it, if looks like a comment just for the next line.  But its a comment for several following lines.

PR #2356, opened 2021-04-06
Use Base64url instead of Base32 encoding for QR URL

30mar:
## Description
the String we get from the qr code should be 64 bit URL string which needs to be converted

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6183


janetback (approved):
UA reviewed

marcussc (approved):

PR #2385, opened 2021-04-07
External camera is not redirecting to the event details

30mar:
## Description
removed mock code and now uses the whole qrCode string

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6136


KaiTeuber (changes requested):
removing the path from the check makes thing more insecure, I don't think this is a good solution

30mar:
> removing the path from the check makes thing more insecure, I don't think this is a good solution

30mar:
> removing the path from the check makes thing more insecure, I don't think this is a good solution

@KaiTeuber  the verification steps changed in the tech spec and now we are using **QRCodeVerificationHelper** class to do all this verification, the only step we need to do is to check if the URL base is **e.coronawarn.app** to verify it should open the app.
Inside the QRCodeVerificationHelper we compare the full qrCodeURLString to some regex this is why im returning the full url instead of the last compnent.

marcussc (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/Route.swift
```diff
@@ -17,18 +17,11 @@ enum Route {
...
 		let components = URLComponents(url: url, resolvingAgainstBaseURL: true)
-		guard components?.host?.lowercased() == "e.coronawarn.app",
-			let lowercasedPath = components?.path.lowercased(),
-			  lowercasedPath.contains("/c1") else {
+		guard components?.host?.lowercased() == "e.coronawarn.app" else {
 			return nil
 		}
-		self = .checkin(url.lastPathComponent)
-		#endif
+		self = .checkin(url.absoluteString)
```
KaiTeuber, line 24: Only verifying the host isn't enough - if you accept everything appended to the URL you can easily start the app and inject some codes. You could at least check for the right query parameter here?

If the Tech Spec changed please let me know where the URL checks are described.

`QRCodeVerificationHelper`should only validate if the content or the QR Code is correct? It's not responsible for the URLs.
Please add the the give MARKS and move test to the designated folder `__tests__`

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/Route.swift
30mar: To be done in this subtask: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6273

KaiTeuber (approved):

PR #2388, opened 2021-04-08
Feature/6120 add error messages for qr codes

30mar:
## Description
This PR contains 2 parts:

1. Add new qrCode error messages to describe which stage did fail when scanning the QRCode (invalid url, invalid payload, etc)
2.  I Separated the verification for External scan and internal scan (both were calling the same helper inside the checkinCoordinator).
3. reasons for the separation:

- if an error occurred after scanning a code by the **Internal camera** the user should have an alert to dismiss or **reactivate** the scanning process (i.e remain inside the scan view controller to scan another qrcode)
- if an error occurred after scanning a code by the **External camera** the user can only have an alert to show the error but we can\t redirect back to the external camera to scan a new code.
- this means we have two different alerts, one has only one action and the other has two
- The checkin coordinator would have been too messy with handling both Alerts + passing a completion handler through different functions to re-activate the scanning in the QRCodeScannerViewController.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6120


tcfos (approved):

SwissMountainFrog (approved):

janetback (approved):
UA reviewed.

PR #2409, opened 2021-04-09
updated text on tab bar item

KaiTeuber:
## Description
Update text of tabbar item

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6279

## Screenshots
Simulator Screen Shot - iPhone 12 Pro - 2021-04-09 at 16 35 28


src/xcode/ENA/ENA/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -1793,7 +1793,7 @@
 
 "Tabbar_Diary_Title" = "Journal";
 
-"Tabbar_CheckIn_Title" = "Check In";
+"Tabbar_CheckIn_Title" = "Check-in";
```
marcussc, line 1796: Not sure if we are allowed to change translated texts.

KaiTeuber: @janetback Can you answer that?

janetback (approved):
UA approved

nickguendling (approved):

PR #2426, opened 2021-04-12
Feature/5138 international hotline

pascal-brause:
## Description
Adds to two screens an international phone number where an user can contact from foreign countries.
Note: To match the changed keys, i added the changed and new strings and removed the old keys also in the other localizabled files. The correct translation will be done later.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5138

## Link to Figma
https://www.figma.com/file/KuDEcMSREg3VhZRhC3FXTY/COVID19_Master_2.0?node-id=8856%3A6325

## Screenshots
Bildschirmfoto 2021-04-12 um 14 53 16
Bildschirmfoto 2021-04-12 um 14 54 05
Bildschirmfoto 2021-04-12 um 14 53 52
Bildschirmfoto 2021-04-12 um 14 54 09


src/xcode/ENA/ENA/Resources/Localization/bg.lproj/Localizable.strings
```diff
@@ -451,6 +451,9 @@
 /* phone number; no spacing allowed */
 "ExposureSubmission_Hotline_Number" = "08007540002";
 
+/* phone number; no spacing allowed */
```
marcussc, line 454: As far as I remember we don't have to touch the Localizable strings in other languages. Translation Team take care of that.

marcussc (approved):

ArturFriesen (approved):

SwissMountainFrog (approved):

src/xcode/ENA/ENA/Resources/Localization/bg.lproj/Localizable.strings
pascal-brause: I asked Andreas and he told me that if I change in the German file something (any key change or addition), it will be paired into the other files and then translated. 
But because I renamed the keys, I wanted to ensure that this will be not forgotten. 
So my idea was: The translation team should see the german terms and will translate them for the correct key.

tcfos: Whatever we change in the english/turkish/... files will be overwritten by translation. For a developer test it might help  to have the text keys already in the localizable.strings files.

tcfos (approved):

janetback (approved):

PR #2427, opened 2021-04-12
Feature/6144 rat preparations/6145 implement service

nickguendling:
## Description
Introduces the CoronaTestService so we can have a PCR test as well as a RAT registered. Goal was to keep the app working as it did previously for PCR Tests.

There will be more unit test added for the CoronaTestService later on.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6144


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -101,22 +102,8 @@ final class RiskProvider: RiskProviding {
...
-			appConfigurationProvider.appConfiguration().sink { [weak self] appConfiguration in
-				self?.downloadTraceWarningPackages(with: appConfiguration, completion: { _ in })
-			}.store(in: &subscriptions)
-
-			return
-		}
-
-		guard !WarnOthersReminder(store: store).positiveTestResultWasShown else {
-			Log.info("RiskProvider: Positive test result was already shown. Don't start new risk detection.", log: .riskDetection)
+		guard !coronaTestService.hasAtLeastOneShownPositiveOrSubmittedTest else {
```
ArturFriesen, line 105: The downloads should not run anymore if there is a positive or submitted test?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -512,7 +500,7 @@ final class RiskProvider: RiskProviding {
 	private func updateRiskProviderActivityState() {
 		if keyPackageDownloadStatus == .downloading || traceWarningDownloadStatus == .downloading {
 			self.updateActivityState(.downloading)
-		} else if (keyPackageDownloadStatus == .idle && (self.store.lastSuccessfulSubmitDiagnosisKeyTimestamp != nil || WarnOthersReminder(store: self.store).positiveTestResultWasShown)) &&
+		} else if keyPackageDownloadStatus == .idle && coronaTestService.hasAtLeastOneShownPositiveOrSubmittedTest &&
```
ArturFriesen, line 503: Why is this check needed? Cant the RiskProvider be stuck in .downloading state with this check?

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
nickguendling: Once at least one registered test has a positive result that was shown to the user or the keys were submitted from it, only the downloads should run. No risk calculation should happen, as we don't show the risk card on the Home Screen in that state. Basically the same logic as we had before, just applying to both tests.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
nickguendling: This is here to get it out of the downloading state. If the risk calculation is running as normal, we don't want to switch back to idle here, because it changes to `.detecting` in `executeExposureDetection()`. But if we only download, and that is the case if `coronaTestService.hasAtLeastOneShownPositiveOrSubmittedTest` is true, then we switch back to idle here when both downloads are finished. This logic was actually introduced because without it the RiskProvider got stuck in the `.downloading` state when we were only downloading and not calculating the risk anymore.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -0,0 +1,396 @@
...
+
+enum CoronaTestServiceError: Error, Equatable {
+	case responseFailure(URLSession.Response.Failure)
+	case unknownTestResult
+	case testExpired
+	case noRegistrationToken
+	case noCoronaTestOfRequestedType
+}
+
+class CoronaTestService {
```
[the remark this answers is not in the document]
nickguendling: Yes, I think we should, I added lots of them now 😁

pascal-brause (approved):

pascal-brause:
At all, nice huge refactoring 😉! 💎

PR #2428, opened 2021-04-12
Feature/6144 rat preparations/6148 update qr code scanner to support rat

30mar:
## Description

-Add Logic for scanning an Antigen test
-Add verification for the payload
-Connect the parsed payload with the test registration API
-Add some unit tests

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6148


nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -843,17 +843,18 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
...
-		model.registerPCRTestAndGetResult(
-			for: guid,
+		model.registerTestAndGetResult(
+			for: testQrCodeInformation,
 			isSubmissionConsentGiven: submissionConsentGiven,
 			isLoading: isLoading,
 			onSuccess: { [weak self] testResult in
-				self?.model.coronaTestType = .pcr
+				
+				self?.model.coronaTestType = CoronaTestType(rawValue: testQrCodeInformation.rawValue)
```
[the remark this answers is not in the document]
30mar:
```suggestion
				self?.model.coronaTestType =  testQrCodeInformation.testType
```

nickguendling (approved):

src/xcode/ENA/ENA/Source/Extensions/Foundation+Base64URL.swift
```diff
@@ -21,6 +21,11 @@ extension Data {
 }
 
 extension String {
+	
+	var isBase64Encoded: Bool {
+		return (self.contains("+") || self.contains("/") || self.contains("="))
```
marcussc, line 26: Hmm not sure if this valid. Should we just try to decode the string into Data and if this fails we can return false here?

30mar: this is the weird part, base64 can be decoded using the base64URL class initializer and it actually works, but im not sure if it is working because of the mock QRCodes or whats so i thought better to just check first and decode accordingly

marcussc: Yeah but `YW55IGNhcm5hbCBwbGVhcw` should be valid Base64 and this function would say its not :/

30mar: I copied the definition of a valid base64 from the techSpecs maybe it is wrong i will fix it then
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/rapid-test/docs/spec/corona-test-client.md#qr-code-of-a-rapid-antigen-test
Screenshot 2021-04-13 at 13 00 11

marcussc (approved):

PR #2435, opened 2021-04-13
Feature/5948 retrieve test result/6154 background update

ArturFriesen:
## Description
Adds background update refactoring and tests for updating test results.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6154


30mar (approved):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -393,6 +441,11 @@ class CoronaTestService {
 			return
 		}
 
+		guard coronaTest.finalTestResultReceivedDate == nil else {
+			completion(.success(coronaTest.testResult))
+			return
+		}
```
nickguendling, line 447: Sadly we cannot do this, because according to the tech spec [the result shall also be updated even if a result other than pending has been received for the same test previously (and regardless of whether the result is stored or not).](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/rapid-test/docs/spec/corona-test-client.md#poll-the-test-result-in-foreground) This is only stated for the foreground, but whether this is also valid for the background is [still open](https://github.com/corona-warn-app/cwa-app-tech-spec/pull/37#discussion_r606218513). Maybe we should leave this check out for now? 🤔

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -422,13 +475,11 @@ class CoronaTestService {
...
-						case .pcr:
-							self.pcrTest?.testResultReceivedDate = Date()
-						case .antigen:
-							self.antigenTest?.testResultReceivedDate = Date()
-						}
+					switch coronaTestType {
+					case .pcr:
+						self.pcrTest?.finalTestResultReceivedDate = Date()
+					case .antigen:
+						self.antigenTest?.finalTestResultReceivedDate = Date()
```
nickguendling, line 482: Just a note: if we take out the check above, it needs to be reinstated here

src/xcode/ENA/ENA/Source/Services/CoronaTestService/TestResult.swift
```diff
@@ -8,6 +8,18 @@ enum TestResult: Int, CaseIterable, Codable {
...
+	//	1	Negative (PCR test)
+	//	2	Positive (PCR test)
+	//	3	Invalid (PCR test)
+	//	4	Redeemed (PCR test; locally referred to as Expired)
+	//	5	Pending (rapid antigen test)
+	//	6	Negative (rapid antigen test)
+	//	7	Positive (rapid antigen test)
+	//	8	Invalid (rapid antigen test)
+	//	9	Redeemed (rapid antigen test; locally referred to as Expired))
+
```
nickguendling, line 22: ❤️

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
```diff
@@ -306,6 +306,102 @@ class CoronaTestServiceTests: XCTestCase {
...
+		testService.pcrTest = PCRTest.mock(registrationToken: "regToken")
+
+		let completionExpectation = expectation(description: "Completion should be called.")
+		testService.updateTestResults(presentNotification: true) { _ in
+			completionExpectation.fulfill()
+		}
+		waitForExpectations(timeout: .short)
+
+		XCTAssertEqual(mockNotificationCenter.notificationRequests.count, 2)
+	}
```
nickguendling, line 331: Would be cool to have a test for the opposite, `presentNotification: false`

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
ArturFriesen: Yes, I think the background task should call the update only if the test result does not exist. Because otherwise it would result in multiple notifications. Therefore I have added a "force" flag for the update.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
ArturFriesen: Done

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
ArturFriesen: @nickguendling could you have a look, if its OK now?

nickguendling (approved):

PR #2442, opened 2021-04-13
Fix/6400 reject manipulated qr codes

30mar:
## Description
Add verification to Reject manipulated QR codes

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6400

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeVerificationHelper.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeVerificationHelper.swift
@@ -95,7 +95,7 @@ class QRCodeVerificationHelper {
 		let startTimeStamp = traceLocation.startDate?.timeIntervalSince1970 ?? 0
 		let endTimeStamp = traceLocation.endDate?.timeIntervalSince1970 ?? 0
 		let bothTimeStampsAreZero = (startTimeStamp == 0 && endTimeStamp == 0)
-		let startTimeIsBeforeEndTime = startTimeStamp < endTimeStamp
+		let startTimeIsBeforeEndTime = startTimeStamp <= endTimeStamp
 		
 		guard bothTimeStampsAreZero || startTimeIsBeforeEndTime else {
 			Log.error("startTimeStamp must be less than endTimeStamp or both should be 0", log: .checkin)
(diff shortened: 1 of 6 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeVerificationHelper.swift
```diff
@@ -83,6 +91,17 @@ class QRCodeVerificationHelper {
 			onError(CheckinQRScannerError.invalidCryptoSeed)
 			return
 		}
+		
+		let startTimeStamp = traceLocation.startDate?.timeIntervalSince1970 ?? 0
+		let endTimeStamp = traceLocation.endDate?.timeIntervalSince1970 ?? 0
+		let bothTimeStampsAreZero = (startTimeStamp == 0 && endTimeStamp == 0)
+		let startTimeIsBeforeEndTime = startTimeStamp < endTimeStamp
```
nickguendling, line 98: In our current implementation it is possible to set startDate and endDate to be equal. 
```suggestion
		let startTimeIsBeforeEndTime = startTimeStamp <= endTimeStamp
```

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Checkin/QRCodeVerificationHelper.swift
30mar: it says here it must be less, should i ignore it ?
Screenshot 2021-04-13 at 15 23 09
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/f5ba0fdcaffc05e44df80067b8ec21d7ba721503/docs/spec/event-registration-client.md#attendee-check-in

nickguendling: I think it's best to ask Max about it, I am not aware of a restriction like that for the trace location generation. We would need to change the implementation there as well if this was the new requirement.

BLUM78691 (approved):
UA reviewed

nickguendling (approved):

PR #2446, opened 2021-04-13
Fix/6364 event store fix cleanup

ArturFriesen:
## Description
Fixes the cleanup of EventStore.
Before this fix, the cleanup was deleting all matches. Also the matches which were still valid time wise.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6364

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Events/Store/EventStore.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Events/Store/EventStore.swift
@@ -12,7 +12,8 @@ import FMDB
 // swiftlint:disable:next type_body_length
 class EventStore: SecureSQLStore, EventStoringProviding {
 
-	static let dataRetentionPeriodInDays = 15
+	static let tenMinutesIntervalLength: TimeInterval = 600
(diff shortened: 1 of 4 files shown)


freshking (approved):

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Events/Store/EventStore.swift
```diff
@@ -12,7 +12,8 @@ import FMDB
 // swiftlint:disable:next type_body_length
 class EventStore: SecureSQLStore, EventStoringProviding {
 
-	static let dataRetentionPeriodInDays = 15
+	static let tenMinutesIntervalLength: TimeInterval = 600
+	static let dataRetentionPeriodInDays: TimeInterval = 15
```
nickguendling, line 16: A bit unusual data type, since a `TimeInterval` is in seconds. Maybe you could use Double instead, but not blocking for this.

nickguendling (approved):

PR #2460, opened 2021-04-15
Event-Check-in Native Camera - wrong screenflow if app is only updated

freshking:
## Description

Show consent info screen before event

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6447


src/xcode/ENA/ENA/Source/Scenes/Checkin/CheckinCoordinator.swift
```diff
@@ -146,6 +158,7 @@ final class CheckinCoordinator {
 		get { store.checkinInfoScreenShown }
 		set { store.checkinInfoScreenShown = newValue }
 	}
+	private var showTraceLocationDetailsAfterInfoScreen: String?
```
marcussc, line 161:
```suggestion
	private var qrCodeAfterInfoScreen: String?
```

naveeddotio (approved):

marcussc (approved):

PR #2469, opened 2021-04-16
Gradient is horizontal and not tilted

freshking:
## Description

Make a proper header view to allow a tilted gradient that also looks good while scrolling

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6293

## Screenshots

ezgif-4-2af7c86a763c


KaiTeuber (changes requested):
I don't think the blur effect was requested here. All screens using the titled gradient shall look the same? Better check with UX first.

SwissMountainFrog (approved):

PR #2470, opened 2021-04-16
Feature/6063 censor private logs/storybranch

pascal-brause:
## Description
This PR enables logging in release builds and ensures that no private informations are logged like the names of persons in the contact diary. This also enables logging in os_log so the logs can be read out with the system console and later in a separate PR, of a file to export it.
The main idea is to flag some private informations by an extensions on the string interpolation.
The usage is then to prefix private informations with `private: ` in the string interpolation of a log entry.
Example:
`let somePrivateInformation = "Martina Musterfrau"`
`Log.info("Created new contact journal person: \(private: somePrivateInformation)")`

I have also added the possibility to add an public explanation text. Example:
		
`let someOtherPrivateInformation = "Max Mustermann"`
`Log.info("Created new contact journal person: \(private: someOtherPrivateInformation, public: "Name of the new created contact journal person")")`

The extension then censores the value only for release builds, so that the developer and testers can see the values in plain for testing purposes.
This works for any type passed into the string interpolation (Int, String, custom structs, enums,...).

##Update to first review:
- Added public parameter to the extension to add some explanation text
- Put in the code the data which we habe to censor
- Make explanation in the` Logging.swift` more visible
- Add some explanation text to the readme.md

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6063

## Screenshots
Examples logged in a release build:
Bildschirmfoto 2021-04-16 um 14 18 50
and the same logged in a debug build:
Bildschirmfoto 2021-04-16 um 14 21 19


ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -45,15 +45,18 @@ extension OSLog {
 /// Log.info("something broke", log: .api)
 /// Log.warning("validation failed", log: .crypto)
 /// Log.error("my hovercraft is full of eels", log: .ui)
+///
+/// ```
+///
+/// Usage for sensitive logging data:
```
marcussc, line 51: Can you add some 🚨🚨🚨🚨 to get more attention on this? :D

marcussc: Also maybe add an Entry to the Readme?

src/xcode/ENA/ENA/Source/Extensions/StringInterpolation+Private.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+
+import Foundation
+	
+extension String.StringInterpolation {
+	/// Use this method when logging sensitive data. This ensures, that the logged information is censored in release builds, but not in debug build, which is needed for us devs and the testers.
+	/// - Parameters:
+	///   - private: The sensitive data to be replaced in release builds with '🙈🙉🙊'.
+	///   - public: Additional explination text what should be logged as data but is censored.
+	mutating func appendInterpolation<T>(private value: T, public text: String = "") {
```
whiskey, line 12: Why not defaulting to the monkeys?
`mutating func appendInterpolation<T>(private value: T, public text: String = "🙈🙉🙊") {`

src/xcode/ENA/ENA/Source/Extensions/StringInterpolation+Private.swift
```diff
@@ -0,0 +1,19 @@
...
+		#if DEBUG
+			appendLiteral(String(describing: value))
+		#else
+			text.isEmpty ? appendLiteral("🙈🙉🙊") : appendLiteral("🙈🙉🙊. (Censoring cause: " + text + ")")
```
whiskey, line 16: I would keep it simple and just use the alternative `text` as is.

whiskey (approved):

src/xcode/ENA/ENA/Source/Extensions/StringInterpolation+Private.swift
pascal-brause: so I would see the monkey OR the public text....with my implementation I would see the monkey in all ways and perhaps the public text.
I think I prefer my approach because then it is more easily visible that we censored something.

src/xcode/ENA/ENA/Source/Extensions/StringInterpolation+Private.swift
pascal-brause: See comment above

marcussc (approved):

PR #2492, opened 2021-04-20
Rearrange Backgroundtasks priority 

marcussc:
## Description
Currently the background task triggers all our Processes that need to run in the Background at the same time. 
This could lead to a situation that the Watchdog is killing our Background process, which could then lead to a situation that the ExposureDetection is triggered too often, which then will lead to Error 13.

This PR changes this behaviour and runs all Backgroundtasks AFTER the ExposureDetection is done.
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5836


pascal-brause (approved):
Indentations seems to be broken in GitHub...

naveeddotio (approved):

PR #2498, opened 2021-04-21
Fix/6388 6389 6312 voice over for checkins and tracelocations

naveeddotio:
## Description
Adding missing voice over for QR code and checkins

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6388
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6389
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6312

## Screenshots
NA


janetback (approved):
UA reviewed

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
```diff
@@ -89,6 +89,8 @@ class EventTableViewCell: UITableViewCell {
 
 		titleLabel.text = cellModel.title
 		addressLabel.text = cellModel.address
+		addressLabel.accessibilityLabel = String(format: AppStrings.TraceLocations.Configuration.address, cellModel.address)
```
nickguendling, line 92: Please put this in the cell model, as this cell is used for checkins and for QR codes. With this implementation it would also read it for checkins.

Also another remark, maybe check with Oliver Keim again: I think it would make more sense to read "QR-Code für ...." before the title and not after the title and before the address.

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
```diff
@@ -89,6 +89,8 @@ class EventTableViewCell: UITableViewCell {
...
+		addressLabel.accessibilityIdentifier = String(format: AppStrings.TraceLocations.Configuration.address, cellModel.address)
```
nickguendling, line 93:
```suggestion
```

The identifier is just used for the UI test, I don't think it makes sense to put the same value as the accessibility label in there.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
naveeddotio: There are two different identifiers, I have created a new one, so not using the one for the UI Tests.

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
naveeddotio: He has another ticket for that. He wants us to add "Check in" <item details> for title. 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6389

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
```diff
@@ -88,6 +88,9 @@ class EventTableViewCell: UITableViewCell {
 		durationStackView.isHidden = cellModel.isDurationStackViewHidden
 
 		titleLabel.text = cellModel.title
+		titleLabel.accessibilityLabel = cellModel.titlePrefix
+		titleLabel.accessibilityIdentifier = cellModel.titlePrefix
```
nickguendling, line 92: You don't need to set the accessibilityIdentifier here at all, that's what I was trying to say with my previous suggestion. The label is all VoiceOver needs. The identifier is not exposed to the user.

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventCellModel.swift
```diff
@@ -18,7 +18,8 @@ protocol EventCellModel {
 
 	var title: String { get }
 	var address: String { get }
-
+	var titlePrefix: String { get }
```
nickguendling, line 21:
```suggestion
	var titleAccessiblityLabel: String { get }
```

I think this is not really a prefix as it also contains the title itself. Just a suggestion attempting to make it more clear what it is :)

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
nickguendling: I see that you grabbed that one as well and fix in with the other one, nice!

src/xcode/ENA/ENA/Source/Scenes/Events/Cells/EventTableViewCell.swift
naveeddotio: Yeah just noticed that.

nickguendling (approved):

marcussc (approved):

PR #2518, opened 2021-04-22
Fix/screenshots

marcussc:
## Description
This PR fixes the failing Screenshots.

The Tests failed because there where two navigation bars on screen. Two? You might ask yourself. Yes two. After we moved from the onboarding to the HomeScreen the Onboarding didn't got removed from the view hierarchy. This PR corrects this.

### DIFF
--- a/src/xcode/ENA/ENA/Source/RootCoordinator.swift
+++ b/src/xcode/ENA/ENA/Source/RootCoordinator.swift
@@ -117,7 +117,8 @@ class RootCoordinator: RequiresAppDependencies {
 		tabBarController.tabBar.tintColor = .enaColor(for: .tint)
 		tabBarController.tabBar.barTintColor = .enaColor(for: .background)
 		tabBarController.setViewControllers([homeCoordinator.rootViewController, checkInCoordinator.viewController, diaryCoordinator.viewController], animated: false)
-
+		
+		viewController.clearChildViewController()
 		viewController.embedViewController(childViewController: tabBarController)
 	}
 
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENAUITests/ENAUITests.swift
```diff
@@ -128,7 +128,7 @@ class ENAUITests: XCTestCase {
 		// ScreenShot_0006: Negative result
 		try navigateThroughOnboarding()
 		XCTAssertTrue(app.buttons[AccessibilityIdentifiers.Home.submitCardButton].waitForExistence(timeout: .medium))
-		app.buttons[AccessibilityIdentifiers.Home.submitCardButton].tap()
```
whiskey, line 131: I see stuff like this all over the place. Why do we have to explicitly select the first one here? 🤯

src/xcode/ENA/ENA/Source/RootCoordinator.swift
```diff
@@ -117,7 +117,8 @@ class RootCoordinator: RequiresAppDependencies {
 		tabBarController.tabBar.tintColor = .enaColor(for: .tint)
 		tabBarController.tabBar.barTintColor = .enaColor(for: .background)
 		tabBarController.setViewControllers([homeCoordinator.rootViewController, checkInCoordinator.viewController, diaryCoordinator.viewController], animated: false)
-
+
```
whiskey, line 120: Forget the old white space. This is NEW white space!
Are we even supposed to call it 'white' space anymore? How about empty space?

whiskey (approved):

janetback (approved):

naveeddotio (approved):

src/xcode/ENA/ENAUITests/ENAUITests.swift
marcussc: TBH I just wanted to fix the Screenshot on this part, I will talk with @nickguendling that we improve here, since he is working on the new UITests for the RAT home. My guess would be that we then have better identifiers :)

PR #2527, opened 2021-04-22
"Data is being downloaded.." is shown every time you navigate back to the homescreen

freshking:
## Description

Prevent package downloading if trace warning metadata is empty. No need to try discovering anything.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6608


src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/TraceWarningPackageDownload.swift
```diff
@@ -226,6 +226,12 @@ class TraceWarningPackageDownload: TraceWarningPackageDownloading {
 		// 3. Clean up revoked Packages.
 		let revokedPackages = appConfig.keyDownloadParameters.revokedTraceWarningPackages
 		removeRevokedTraceWarningMetadataPackages(revokedPackages)
+				
+		guard !eventStore.traceWarningPackageMetadatasPublisher.value.isEmpty else {
+			Log.info("TraceWarningPackageDownload: Aborted due to checkin metadata database is empty.", log: .checkin)
+			completion(.success(.success))
```
whiskey, line 232: I heard you like success, …

whiskey (approved):

30mar (approved):

src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/TraceWarningPackageDownload.swift
```diff
@@ -275,6 +273,14 @@ class TraceWarningPackageDownload: TraceWarningPackageDownloading {
 				
 		// 7. Determine packagesToDownload
 		let packagesToDownload = determinePackagesToDownload(availables: availablePackagesOnCDN, earliest: earliestRelevantPackage)
+		
+		guard !packagesToDownload.isEmpty else {
```
pascal-brause, line 277: Is this really necessary for the bugfix or does this handle another bug?
Isn't that handled by the forEach in line 294 ff.? Because when the packages are empty, the success should be triggered anyway?
If not, I recommend to add a new case for `TraceWarningSuccess`, because this is not equal to the regular success.

src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/__tests__/TraceWarningPackageDownloadTests.swift
```diff
@@ -93,7 +93,7 @@ class TraceWarningPackageDownloadTests: XCTestCase {
 		
 		// THEN
 		waitForExpectations(timeout: .medium)
-		XCTAssertEqual(responseCode, .noCheckins)
+		XCTAssertEqual(responseCode, .success)
```
pascal-brause, line 96: Because we have these special success cases just for testing and nothing else, I recommend to use this case. 
And because we have no other test to proof for a .noCheckin success, I would think that the code does not what the test expect...Could you check that again?

pascal-brause (changes requested):

src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/TraceWarningPackageDownload.swift
```diff
@@ -275,6 +273,14 @@ class TraceWarningPackageDownload: TraceWarningPackageDownloading {
...
+			Log.info("TraceWarningPackageDownload: Aborted due to no packages to download.", log: .checkin)
+			completion(.success(.success))
+			return
+		}
+		
+		status = .downloading
```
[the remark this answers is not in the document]
freshking: Yes

src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/__tests__/TraceWarningPackageDownloadTests.swift
freshking: I added a new case for this

src/xcode/ENA/ENA/Source/Services/TraceWarningDownload/TraceWarningPackageDownload.swift
freshking: That is correct but the reason this guard is here is to prevent changing the status to "downloading". Also, there is no need to try to attempt to download an empty package array.

pascal-brause: When its preventing the status to change, it' fine. Otherwise i agree with you with the "no need", but actually its exactly implemented like the tech spec defines it. So I would be careful with any logical change. But it is okay for me here.

pascal-brause (approved):

PR #2538, opened 2021-04-23
Fix/6692 migration

nickguendling:
## Description
The migration of PCR tests submitted in v2.0 and earlier was not correct. As the registration token was already nil and the test result could not be fetched from the server anymore, we need to set the test result to .positive for submitted tests. And we need to keep the positiveTestResultShown flag set to true even after the submission because the home table view model logic is relying on that in v2.1.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6692

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
@@ -405,7 +405,7 @@ class CoronaTestService {
 				registrationToken: store.registrationToken,
 				testResult: keysSubmitted || store.positiveTestResultWasShown ? .positive : .pending,
 				finalTestResultReceivedDate: store.testResultReceivedTimeStamp.map { Date(timeIntervalSince1970: TimeInterval($0)) },
-				positiveTestResultWasShown: store.positiveTestResultWasShown,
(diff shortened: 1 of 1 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -397,16 +397,17 @@ class CoronaTestService {
...
 
 	func migrate() {
-		if store.registrationToken != nil || store.lastSuccessfulSubmitDiagnosisKeyTimestamp != nil, let testRegistrationTimestamp = store.devicePairingConsentAcceptTimestamp {
+		let keysSubmitted = store.lastSuccessfulSubmitDiagnosisKeyTimestamp != nil
+		if store.registrationToken != nil || keysSubmitted, let testRegistrationTimestamp = store.devicePairingConsentAcceptTimestamp {
 			pcrTest = PCRTest(
 				registrationDate: Date(timeIntervalSince1970: TimeInterval(testRegistrationTimestamp)),
 				registrationToken: store.registrationToken,
-				testResult: .pending,
+				testResult: keysSubmitted || store.positiveTestResultWasShown ? .positive : .pending,
```
marcussc, line 405: Some Comments for future us would be nice :)

SwissMountainFrog (approved):

PR #2557, opened 2021-04-27
Fix/6766 reset appconfig on environment change

nickguendling:
## Description
Reset app config on environment switch so the country list is up to date when submitting keys immediately after switching the environment.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6766
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6781
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6783

### DIFF
--- a/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMServerEnvironmentViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Developer Menu/Features/DMServerEnvironmentViewController.swift
@@ -33,7 +33,7 @@ class DMServerEnvironmentViewController: UIViewController, UIPickerViewDelegate,
 
 		currentEnvironmentLabel = UILabel(frame: .zero)
 		currentEnvironmentLabel.translatesAutoresizingMaskIntoConstraints = false
-		updateCurrentEnviromentLabel()
+		updateCurrentEnvironmentLabel()
 
 		picker = UIPickerView(frame: .zero)
 		picker.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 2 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMServerEnvironmentViewController.swift
```diff
@@ -99,7 +99,8 @@ class DMServerEnvironmentViewController: UIViewController, UIPickerViewDelegate,
 
 			let selectedRow = self.picker.selectedRow(inComponent: 0)
 			self.store.selectedServerEnvironment = self.serverEnvironment.availableEnvironments()[selectedRow]
-			self.updateCurrentEnviromentLabel()
+			self.updateCurrentEnvironmentLabel()
+			self.store.appConfigMetadata = nil
```
ArturFriesen, line 103: Should we clear the whole store to avoid problems like this in the future?
Also, not in scope of your fix, but shouldn't we also reset the event store?

ArturFriesen (changes requested):
Please see comment.

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMServerEnvironmentViewController.swift
nickguendling: I'm not actually sure if we should reset everything, we should probably talk to testing about it if this is desirable for them to reset everything on every environment switch. Would definitely help to avoid these kinds of problems in the future

ArturFriesen (approved):

PR #2625, opened 2021-05-05
remove precondition failures to prevent crashes in the catch block

30mar:
## Description
Remove precondition failures to prevent crashes in the catch block, which blocks the testers

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6602

## Screenshots
Screenshot 2021-05-05 at 09 59 43


src/xcode/ENA/ENA/Source/Workers/Logging/StreamReader.swift
```diff
@@ -29,7 +29,7 @@ class StreamReader {
 				let delimData = delimiter.data(using: encoding),
 				let buffer = NSMutableData(capacity: chunkSize)
 			else {
-				preconditionFailure("Cannot initialize StreamReader for file at \(url)")
+				Log.error("Cannot initialize StreamReader for file at \(url)", log: .localData)
```
marcussc, line 32: This would lead to no Log being displayed right?
I wonder why the guard fails in the first place. Do we have a clue why?

30mar: actually the current crash is caused by the catch block but i though to remove the fatal error in both cases,
specifically this line:
`let fileHandle = try FileHandle(forReadingFrom: url)`
Hopefully now with the log we can see the actual reason

marcussc: Ah okay so this is not the real fix but changing the logs to see the real Issue?

marcussc (approved):

KaiTeuber (approved):

freshking (approved):

PR #2645, opened 2021-05-06
 PCR & Rapid Test - After deleting a test, the risk card is not updated

freshking:
## Description

Provide risk result to consumer to properly update UI + download trace packages after key packages to ensure no race condition occurs

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7076


src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -107,11 +108,28 @@ final class RiskProvider: RiskProviding {
...
+				}
+				self.appConfigurationProvider.appConfiguration().sink { [weak self] appConfiguration in
+					self?.downloadTraceWarningPackages(with: appConfiguration, completion: { [weak self] result in
+						guard let self = self else {
+							return
+						}
+						switch result {
+						case .success:
+							if let risk = self.previousRiskIfExistingAndNotExpired(userInitiated: userInitiated) {
+								self.successOnTargetQueue(risk: risk)
```
ArturFriesen, line 123: Do I understand in correctly?
After the download of key packages and warning packages, success is called with the previously calculated risk.
And then the calculation of the risk is triggered, which the also may return success with the new calculated risk.

To avoid race conditions, why is the logic not simply calling "_requestRiskLevel" inside of download completion?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -107,11 +108,28 @@ final class RiskProvider: RiskProviding {
...
+							} else {
+								assertionFailure("This should never happen. At this point there should be a risk to use available.")
+							}
+						case .failure(let error):
+							self.failOnTargetQueue(error: error)
```
ArturFriesen, line 128: Why does a failed trace warning package call fail, but not a failed key package download?
In general I'm not sure if its a good pattern, to have several completion calls (maybe even failure and success) for 1 risk request.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: I missed that its in a guard with return. Cleared.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
ArturFriesen: I missed that its in a guard with return. Cleared.

ArturFriesen (approved):

whiskey (approved):
Would be much cleaner if we get time to refactor the 'download...' stuff (basically the whole Client) to use Combine 😐

pascal-brause (approved):

Issue #2631, opened 2021-05-05
UI: Empty space on some screens


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone XR, iPhone 6s, iPhone SE 1, iOS 14.5.1. (So probably affects all screen sizes)
- iOS version: 14.5.1, 14.6 Beta 2
- App version: 2.1.3 (0)

## Describe the bug

There are some screens where there are unnecessary free spaces.

## Steps to reproduce the issue

Depending on screens, see screenshots below, should be clear. On the Home Screen the bug is only visible below the "Create QR Code" screen.

## Expected behaviour

No free/empty space

## Screenshots
<img src="https://user-images.githubusercontent.com/67682506/117199726-d60c1280-adea-11eb-925c-6227e8fdae10.PNG" width="300">
<img src="https://user-images.githubusercontent.com/67682506/117199742-da383000-adea-11eb-88e3-e1dda73b6cd3.PNG" width="300">
<img src="https://user-images.githubusercontent.com/67682506/117199750-dc9a8a00-adea-11eb-81a8-38c43958bcb3.PNG" width="300">
<img src="https://user-images.githubusercontent.com/67682506/117199752-ddcbb700-adea-11eb-8008-bf902ddc11c9.PNG" width="300">

---
Internal Tracking ID: [EXPOSUREAPP-7098](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7098)

dsarkar:
@Ein-Tim Good morning! Thanks.
- Affect only main screen apparently?
- I can reproduce.

jucktnich:
Can reproduce it with iPhone SE 1, iOS 14.5.1. (So probably affects all screen sizes)

dsarkar:
@jucktnich thanks, will add this to OP.

dsarkar:
@jucktnich I could find this bug only on the main screen. You? Thanks for the feedback.

jucktnich:
I too only on the homescreen, sorry forgot to add this to my comment

jucktnich:
[...]

PR #2650, opened 2021-05-07
Use grouped table view style to prevent headers and footers from floating above the cells

nickguendling:
## Description
With the introduction of the table view on the Home Screen we had the wrong background color for a while, so recently I fixed the color using header and footer views. Now the table view style needs to be set to grouped so they are not floating above the cells.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7098


dsarkar:
Fixes #2631

freshking (approved):

marcussc (approved):

ArturFriesen (approved):

PR #2661, opened 2021-05-10
Fix/6989 wrong number while calling

naveeddotio:
## Description
Wrong number was called due to extra string

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6989

## Screenshots
NA


freshking (approved):

naveeddotio:
@NataliaLemmerth @ChristinaTsigganos Can I have your tick here on priority basis?

NataliaLemmerth (approved):
UA approved

marcussc (changes requested):
We have to wait for the Translation delivery.
Then check and maybe replace this string in the translations as well.

naveeddotio:
> We have to wait for the Translation delivery.
> Then check and maybe replace this string in the translations as well.

That's the confusing part, it's a separate key.

marcussc:
> > We have to wait for the Translation delivery.
> > Then check and maybe replace this string in the translations as well.
> 
> That's the confusing part, it's a separate key. Please recheck.

@naveeddotio Yes, but it is contained in the other translation relevant files as well:
Screenshot 2021-05-10 at 11 18 58
So changing it now would screw the translation process, so we should wait for the incoming translation delivery and then change it in all languages.

naveeddotio:
> > > We have to wait for the Translation delivery.
> > > Then check and maybe replace this string in the translations as well.
> > 
> > 
> > That's the confusing part, it's a separate key. Please recheck.
> 
> @naveeddotio Yes, but it is contained in the other translation relevant files as well:
> Screenshot 2021-05-10 at 11 18 58
> So changing it now would screw the translation process, so we should wait for the incoming translation delivery and then change it in all languages.

Same goes here then!
Screenshot 2021-05-10 at 14 23 09

marcussc (approved):

PR #2662, opened 2021-05-10
Feature/6742 vaccination certificate/7132 proof zertificate base45

ArturFriesen:
## Description
Fetch the proof certificate as base45.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7132


src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/DigitalGreenCertificateAccess.swift
```diff
@@ -29,29 +31,23 @@ public struct DigitalGreenCertificateAccess {
...
 
-    public func extractDigitalGreenCertificate(from cborData: CBORData) -> Result<DigitalGreenCertificate, CertificateDecodingError> {
-        return extractCertificate(from: cborData)
-    }
-
     // MARK: - Internal
 
     func extractCBOR(from base45: Base45) -> Result<CBORData, CertificateDecodingError> {
-        guard let zipData = try? base45.fromBase45() else {
+        let base45WithoutPrefix = base45.dropPrefix(hcPrefix)
```
nickguendling, line 48: I'm wondering if we should throw the VC_PREFIX_INVALID error here

ArturFriesen: added error and handling.

KaiTeuber (approved):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/DigitalGreenCertificateAccess.swift
```diff
@@ -29,29 +31,26 @@ public struct DigitalGreenCertificateAccess {
...
-    public func extractDigitalGreenCertificate(from cborData: CBORData) -> Result<DigitalGreenCertificate, CertificateDecodingError> {
-        return extractCertificate(from: cborData)
-    }
-
     // MARK: - Internal
 
     func extractCBOR(from base45: Base45) -> Result<CBORData, CertificateDecodingError> {
-        guard let zipData = try? base45.fromBase45() else {
+        guard base45.hasPrefix(hcPrefix) else {
+            return .failure(.HC_COSE_TAG_INVALID)
```
nickguendling, line 49:
```suggestion
            return .failure(.HC_PREFIX_INVALID)
```

(I guess it needs to be "HC" instead of "VC" like it is in the tech spec)

nickguendling (changes requested):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/DigitalGreenCertificateAccess.swift
ArturFriesen: Yes I think it needs to bo HC. Because from the perspective of the package it can happen to both VC and PC. Added HC_PREFIX_INVALID. And also use it now 🤦‍♂️

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/HTTP/HTTPService.swift
```diff
@@ -0,0 +1,19 @@
+//
+// 🦠 Corona-Warn-App
+//
+
```
pascal-brause, line 4: Why do we encapsulate the http stuff now in a separate service and not in our common httpClient?

pascal-brause: ah forget you are working in a separate package. Forget about my comment.

pascal-brause (approved):

nickguendling (approved):

PR #2680, opened 2021-05-11
Feature/5474 flaky tests/6877 prevent new flaky tests

pascal-brause:
## Description
One common problem while analyzing our flaky UI tests was, that a lot of tests are too fast for the UI. So whenever we tap a button, we should wait a moment and then tap the element. 
For this we have now the method `tapAndWait` with the possibility to pass a timeout duration. 
I replaced ALL occurrences of `tap()` with this new method and removed, where it was adding before the tap, the `waitForExistence()`

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6877


src/xcode/ENA/ENAUITests/TraceLocations/ENAUITests_09_TraceLocations.swift
```diff
@@ -35,8 +34,7 @@ class ENAUITests_09_TraceLocations: XCTestCase {
 		app.launch()
 
 		let traceLocationsCardButton = app.buttons[AccessibilityIdentifiers.Home.traceLocationsCardButton]
-		XCTAssertTrue(traceLocationsCardButton.waitForExistence(timeout: .extraLong))
-		traceLocationsCardButton.tap()
```
ArturFriesen, line 39: maybe it makes sense to preserve the .extraLong timeout

ArturFriesen (approved):

marcussc (approved):

src/xcode/ENA/ENAUITests/ENAUITests-Extensions.swift
```diff
@@ -35,6 +35,13 @@ extension XCUIElement {
 		guard exists, !frame.isEmpty else { return false }
 		return XCUIApplication().windows.element(boundBy: 0).frame.contains(frame)
 	}
+	
+	func waitAndTap(_ timeout: Double = .short) {
+		// Need XCTAssertEqual() here because we can append a message, this is not possible for XCTAssertTrue()
+		// swiftlint:disable:next xct_specific_matcher
+		XCTAssertEqual(self.waitForExistence(timeout: timeout), true, "Could not find XCUIElement: \(self) before tapping it.")
```
whiskey, line 42: `isHittable` might prevent this:

```suggestion
		XCTAssertTrue(self.waitForExistence(timeout: timeout))
		XCTAssertTrue(self.isHittable)		
```

> isHittable returns true if the element exists and can be clicked, tapped, or pressed at its current location. It returns false if the element does not exist, is offscreen, or is covered by another element.
>
> **Note**
isHittable only returns true if the element is already visible and hittable onscreen. It returns false for an offscreen element in a scrollable view, even if the element would be scrolled into a hittable position by calling click(), tap(), or another hit-point-related interaction method.

([Apple](https://developer.apple.com/documentation/xctest/xcuielement/1500561-ishittable))

whiskey (approved):

src/xcode/ENA/ENAUITests/TraceLocations/ENAUITests_09_TraceLocations.swift
pascal-brause: i had the idea to start for all tests with a small timeout, and if it fails, we can increase the timeout afterwards. Until now, Actually it seems that it works with the small timeout (About 30+ runs with no error)

src/xcode/ENA/ENAUITests/ENAUITests-Extensions.swift
pascal-brause: Hmm yeah, but isHittable would only add here another check, but it does not replaces another check....?
But I'd love to let here the error message, because we use this generic approach for so many calls (< 340), that could make it more easy to get for wich element it does not work.

PR #2684, opened 2021-05-12
PCR scanned followed by TAN Input does not trigger the PCR exchange screen

freshking:
## Description

Present overwrite notice when adding a new test via TAN and another test is already registered.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7143


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -92,6 +92,32 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
...
+					didTapCloseButton: { [weak self] in
+						self?.navigationController?.dismiss(animated: true) {
+							feedbackHandler(false)
+						}
+					}
+				)
+
+				let footerViewController = FooterViewController(footerViewModel)
+				let topBottomViewController = TopBottomContainerViewController(topController: overwriteNoticeViewController, bottomController: footerViewController)
+				let nc = UINavigationController(rootViewController: topBottomViewController)
```
KaiTeuber, line 118: `TestOverwriteNoticeViewController` uses the protocol `dismissHandling` to handle that you should use `DismissHandlingNavigationController` instead.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
freshking: @KaiTeuber I need the `UINavigationController` to show a navigation bar. `TopBottomContainerViewController` is presented and not pushed.

KaiTeuber: @freshking Just replace `UINavigationController` with `DismissHandlingNavigationController` that's all needed to respect the DismissHandling

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/TANInputViewController/TanInputViewModel.swift
```diff
@@ -116,5 +114,22 @@ final class TanInputViewModel {
...
 	}
-
+	
+	private func registerPCRTest() {
+		isPrimaryButtonEnabled = false
+		isPrimaryBarButtonIsLoading = true
+		coronaTestService.registerPCRTest(teleTAN: text, isSubmissionConsentGiven: false) { [weak self] result in
+			switch result {
+			case let .failure(error):
+				// If teleTAN is incorrect, show Alert Controller
```
pascal-brause, line 124: is in one of the following methods (like the alert) some error logged? If not, perhaps add some logs here...

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -92,6 +92,32 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
...
+						self?.navigationController?.dismiss(animated: true) {
+							feedbackHandler(false)
+						}
+					}
+				)
+
+				let footerViewController = FooterViewController(footerViewModel)
+				let topBottomViewController = TopBottomContainerViewController(topController: overwriteNoticeViewController, bottomController: footerViewController)
+				let nc = DismissHandlingNavigationController(rootViewController: topBottomViewController)
+				self?.navigationController?.present(nc, animated: true)
```
nickguendling, line 119: Is there a reason why it is presented here and not pushed like it is after scanning a second test?

freshking: It looked very weird to me pushing that view here.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/TANInputViewController/TanInputViewModel.swift
freshking: Done

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: UX wanted it to be pushed and the whole submission flow to be dismissed when the close button is pushed.

I think it would be easiest and most consistent with the rest of the coordinator to call `showOverrideTestNotice` here and keep the logic that decides if it needs to be shown also in the coordinator. We would just need to extend the `CoronaTestQRCodeInformation` with the TAN case.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
freshking: Done

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -92,6 +92,31 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
...
+					didTapCloseButton: { [weak self] in
+						self?.navigationController?.dismiss(animated: true) {
+							feedbackHandler(false)
+						}
+					}
+				)
+
+				let footerViewController = FooterViewController(footerViewModel)
+				let topBottomViewController = TopBottomContainerViewController(topController: overwriteNoticeViewController, bottomController: footerViewController)
+				self?.push(topBottomViewController)
```
nickguendling, line 118: I checked it out: it doesn't work, closes the flow when tapping on continue on the override notice screen. Thats why I suggested to extend the `CoronaTestQRCodeInformation` with the TAN case and use the same pattern as we do after scanning a QR code. It also shows the appropriate loading state on the confirmation screen as well as the error messages.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
freshking: That is exactly the expected behavior!

nickguendling: The same thing that happened before: the positive test result is shown so the user can submit the keys

nickguendling (changes requested):
I like this approach, just have a couple of smaller suggestions :)

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -441,7 +463,7 @@ class CoronaTestService {
 			.store(in: &subscriptions)
 	}
 
-	private func getRegistrationToken(
+	func getRegistrationToken(
```
nickguendling, line 466: Is this change needed?

nickguendling (approved):

pascal-brause (approved):

src/xcode/ENA/ENAUITests/ENAUITests-Extensions.swift
```diff
@@ -37,6 +37,7 @@ extension XCUIElement {
 		// Need XCTAssertEqual() here because we can append a message, this is not possible for XCTAssertTrue()
 		// swiftlint:disable:next xct_specific_matcher
 		XCTAssertEqual(self.waitForExistence(timeout: timeout), true, "Could not find XCUIElement: \(self) before tapping it.")
+		XCTAssertTrue(self.isEnabled)
```
freshking, line 40: @pascal-brause FYI

PR #2705, opened 2021-05-17
information when CWA app was installed partially wrong

freshking:
## Description

Show an installation time of one day if installation date was yesterday. Otherwise it will still show today because 24 hours might not have passed yet.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7145


pascal-brause (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeState.swift
```diff
@@ -100,7 +100,13 @@ class HomeState: ENStateHandlerUpdating {
...
 
 	var daysSinceInstallation: Int {
-		store.appInstallationDate?.ageInDays ?? 0
+		guard let appInstallationDate = store.appInstallationDate else {
+			return 0
+		}
+		if Calendar.current.isDateInYesterday(appInstallationDate) {
+			return 1
+		}
+		return appInstallationDate.ageInDays ?? 0
```
nickguendling, line 109: I think the desired effect is to have it change at midnight for every amount of days since installation, not just for the change from "today" to "yesterday". For clarification see also the duplicate linked in the comments of the bug ticket.
```suggestion
		return Calendar.current.startOfDay(for: appInstallationDate).ageInDays ?? 0
```

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeState.swift
naveeddotio: Even better -> Calendar.autoupdatingCurrent

freshking: Done

nickguendling (approved):

naveeddotio (approved):

KaiTeuber (approved):

Issue #2695, opened 2021-05-12
Rapid Test Profile: Date of birth off one day


abbe79:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 12 Pro
- iOS version: 14.5.1
- App version: 2.2.1 (0)
- MESZ timezone
- German language and localization settings

## Describe the bug

After entering my data in the rapid test profile mask, the birthday included in the QR code (and below in the text summary) is one day before the entered day.

## Steps to reproduce the issue

See screenshots.
1.  Enter a new rapid test profile, choose a date. 
2. Click next. 
3. Compare the date.

## Expected behaviour

The data entered in the mask should be the same as encoded in the qr code (and displayed below)

## Additional context

![IMG_1346](https://user-images.githubusercontent.com/20374998/118053662-3bc94300-b385-11eb-9c45-41036cdddb21.PNG)
![IMG_1347](https://user-images.githubusercontent.com/20374998/118053670-3e2b9d00-b385-11eb-9c4c-c013334357bf.PNG)

----
Internal Tracking ID: [EXPOSUREAPP-7263](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7263)

abbe79:
Also reported in #2696

Ein-Tim:
Hey @abbe79 same goes here, I am not able to reproduce this (see https://github.com/corona-warn-app/cwa-app-ios/issues/2696#issuecomment-840377994). I even entered the birthday shown on your screenshot and am not able to reproduce this.

Do you use a special keyboard app downloaded from the App Store?

I tried on a real iPhone 6s with iOS 14.5.1 and CWA 2.2.1 and on an Emulator (iPhone 12 Pro) running iOS 14.5 and CWA 2.2.1.

I used MESZ timezone & German language and localization settings, just like you.

oleksiikutuzov:
I can also confirm the issue. No special keyboard, MESZ timezone. I tried to create it about at the same time as @abbe79 did. 20 March 1998 worked fine, 30 March 1998 was always saving as 29 March 1998. Currently I can't reproduce it as well.

[...]

Issue #2696, opened 2021-05-12
Date of birth is always one day in the past


Jens07:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 11
- iOS version: 14.4.2
- App version: 2.2.1

## Describe the bug

After the creation of a rapid test profile the date of birth is always one day in the past compared to the value in the previous input field.

## Steps to reproduce the issue

1. Go to 'register your test'
2. Click on 'next steps'
3. Click on 'create rapid test profile'
4. See that the date of birth is always one day before the selected date. I tried it with a plurality of dates. It seems always the same problem.

## Expected behaviour

The date of birth should be correct.

Ein-Tim:
Hey @Jens07 

I tried the following:
1. Click on "Next Steps" in the "Register your Test" box
2. "Create Rapid Test Profile"
3. Enter "01.01.2000" into the field via the picker
4. Click on "Save" 
5. The date of birth is shown correctly as "1. Jan 2000".

Could you maybe upload a screen recoding? Are you using a keyboard from the App Store? 

Thank you!

Ein-Tim:
@Jens07 Please also see #2695, where the same was reported before. I think it would be good if you could follow up there and close this issue here to be able to better keep track.

dsarkar:
@Jens07 Thanks for reporting. Internal Tracking ID: [EXPOSUREAPP-7263](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7263)


---
Corona-Warn-App Open Source Team

dsarkar:
FYI PR https://github.com/corona-warn-app/cwa-app-ios/pull/2707

dsarkar:
FYI new PR https://github.com/corona-warn-app/cwa-app-ios/pull/2741

dsarkar:
Fixed in 2.3

dsarkar:
@Jens07 Thanks for reporting. Best, DS


---
Corona-Warn-App Open Source Team

PR #2707, opened 2021-05-17
RAT Profile: DoB off one day

freshking:
## Description

Save local date instead of UTC date to antigen test profile because when encoding it will convert it to just date.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7263


naveeddotio (approved):

dsarkar:
Fixes: https://github.com/corona-warn-app/cwa-app-ios/issues/2695 https://github.com/corona-warn-app/cwa-app-ios/issues/2696

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/__tests__/CreateAntigenTestProfileViewModelTests.swift
```diff
@@ -100,4 +100,42 @@ class CreateAntigenTestProfileViewModelTests: XCTestCase {
...
+			let decodedAntigenTestProfile = try decoder.decode(AntigenTestProfile.self, from: encodedData)
+			let decodedDate = try XCTUnwrap(decodedAntigenTestProfile.dateOfBirth)
+			
+			XCTAssertTrue(Calendar.current.component(.day, from: decodedDate) == day, "\(Calendar.current.component(.day, from: decodedDate)) != \(day)")
+			XCTAssertTrue(Calendar.current.isDate(decodedDate, equalTo: date, toGranularity: .day))
+			
+		} catch {
+			XCTFail(error.localizedDescription)
+		}
+	}
```
pascal-brause, line 140: 💎

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/Cells/CreateAntigenTestProfileInputCell.swift
```diff
@@ -35,6 +35,7 @@ class CreateAntigenTestProfileInputCell: UITableViewCell, ReuseIdentifierProvidi
 
 		birthdayPicker = UIDatePicker()
 		birthdayPicker.locale = Locale(identifier: "de_DE")
+		birthdayPicker.timeZone = TimeZone.current
```
KaiTeuber, line 38: just wonder why the picker locale is set to 'de_DE' - don't we support other settings as well?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
```diff
@@ -189,7 +189,10 @@ class CreateAntigenTestProfileViewController: UITableViewController, FooterViewH
 	
 	@objc
 	private func dateOfBirthDidChange(datePicker: UIDatePicker) {
-		viewModel.update(datePicker.date, keyPath: \.dateOfBirth)
+		let offset = datePicker.timeZone?.secondsFromGMT() ?? 0
+		let utcDate = datePicker.date
+		let date = utcDate.addingTimeInterval(TimeInterval(offset))
+		viewModel.update(date, keyPath: \.dateOfBirth)
```
nickguendling, line 195: Wouldn't it be possible to let the user pick the time in UTC instead of converting it to the equivalent UTC date manually?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
nickguendling: And I just checked another implementation detail: we are currently using UTC to decode the date without time information, so it is decoded as the UTC midnight date. But then we are using a formatter with local time zone in the `AntigenTestProfileViewModel`. I think we should also use a etc date formatter here, because the profile breaks if it is viewed in time zones that have an offset < 0 to UTC. We're just lucky that most of our users are in a time zone where it works.

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/Cells/CreateAntigenTestProfileInputCell.swift
freshking: @KaiTeuber I was wondering that myself but I don't know.

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
freshking: @nickguendling The user is picking the that in UTC. That is exactly the issue here. That is why I converted the date to local time because when encoding it is converted to just date.

nickguendling: The default time zone for the date picker is the local time so it interprets the selected date in the local time zone. `ISO8601DateFormatter.justDate` uses UTC to interpret that date. It is an option to now calculate the UTC date that has the same date components as the selected local date has (although I would suggest a different naming since the `utcDate` is the selected date in local time, not UTC). Or you can just set the UIDatePicker to UTC to interpret it in UTC already (I tried it 😉). And since or en- and decoding uses UTC, you need to use a UTC date formatter to show it.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
nickguendling: ‎

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
freshking: I do not understand the discussion currently. This PR fixes the issue. Please open another feature ticket if you want to do a code refactoring.

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/Cells/CreateAntigenTestProfileInputCell.swift
```diff
@@ -37,6 +37,7 @@ class CreateAntigenTestProfileInputCell: UITableViewCell, ReuseIdentifierProvidi
...
```
marcussc, line 40:
```suggestion
		birthdayPicker.timeZone = TimeZone.utcTimeZone
```

Would make the most sense for me since we are supposed to handle all dates in UTC.

src/xcode/ENA/ENA/Source/Scenes/CreateAntigenTestProfile/CreateAntigenTestProfileViewController.swift
freshking: @nickguendling Please compare the following two screenshots. Check the time on my device, the date I selected in the app and the debug output from the date picker's date. How is that not UTC?

Bildschirmfoto 2021-05-19 um 15 44 01
IMG_2970

nickguendling: Since a `Date` object does not contain time zone information, it is shown in its UTC representation by the debugger. The date only contains the seconds from the reference date (2001-01-01 00:00 UTC) to the date selected by the user (2020-05-19 00:42:59 UTC+02:00). So it can only be interpreted correctly as the date of birth if you know the time zone it was selected in. That is the bug. We interpret it in UTC, while we have to take the user time zone into account to interpret it correctly. That's what your solution does: It calculates the point in time when UTC has the same time (00:42:59) as the user had locally when selecting the date. After that calculation, it can be correctly interpreted in the UTC time zone. So if you want to call something the `utcDate`, it would be the result of the calculation, not the input. Because date calculations and time zones are hard and we can make it simpler: start with UTC in the date picker, then the calculation is not needed.

freshking: I understand all that but you literally said - and I quote - "The naming of utcDate for a local date is misleading[...]". But this is a UTC date. So why is naming the UTC date "utcDate" wrong?

PR #2727, opened 2021-05-18
Fix/@Published timing issues

nickguendling:
## Description
This PR introduces the `@DidSetPublished` property wrapper to solve timing issues that originally lead to a bug (linked below) where the risk provider did not provide a risk even though the test was already supposed to be deleted. This was due to the publisher triggering on `willSet` and the `RiskProvider` still seeing the tests that were about to be deleted and returning early.

This is basically part 2 to this pull request: [#2721](https://github.com/corona-warn-app/cwa-app-ios/pull/2721) as it allows the risk provider to be triggered immediately after deleting a corona test.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7076

### DIFF
--- a/src/xcode/ENA/ENA/Source/Utils/__tests__/DidSetPublishedTests.swift
+++ b/src/xcode/ENA/ENA/Source/Utils/__tests__/DidSetPublishedTests.swift
@@ -12,8 +12,8 @@ class DidSetPublishedTests: XCTestCase {
 
 		let expectedValues = [false, false, true, true, false]
 
-		let expectation = expectation(description: "Received value")
-		expectation.expectedFulfillmentCount = expectedValues.count
+		let publishingExpectation = expectation(description: "Received value")
+		publishingExpectation.expectedFulfillmentCount = expectedValues.count
 
 		var receivedValues = [Bool]()
 		let subscription = publishingStruct.$publishingBool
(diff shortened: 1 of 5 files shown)


src/xcode/ENA/ENA/Source/Utils/__tests__/DidSetPublishedTests.swift
```diff
@@ -0,0 +1,48 @@
...
+		publishingStruct.publishingBool = false
+		publishingStruct.publishingBool = true
+		publishingStruct.publishingBool = true
+		publishingStruct.publishingBool = false
+
+		waitForExpectations(timeout: .short)
+
+		XCTAssertEqual(receivedValues, expectedValues)
+
+		subscription.cancel()
```
whiskey, line 37: Just out of interest, wouldn't this cancel automatically as the `subscription` is only scoped in this function?

whiskey (approved):

src/xcode/ENA/ENA/Source/Utils/__tests__/DidSetPublishedTests.swift
nickguendling: I think that's true, it's just an elegant way for me to silence the `Initialization of immutable value 'subscription' was never used; consider replacing with assignment to '_' or removing it` warning

marcussc (approved):

ArturFriesen (approved):

PR #2737, opened 2021-05-19
Crash on NSProgress addChild:withPendingUnitCount:

freshking:
## Description

Prevent multiple exposure detections simultaneously. Why this is called multiple times sometimes I do not know but this prevents it from executing simultaneously.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4368


src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -356,15 +357,21 @@ final class RiskProvider: RiskProviding {
 		appConfiguration: SAP_Internal_V2_ApplicationConfigurationIOS,
 		completion: @escaping (Result<[ExposureWindow], RiskProviderError>) -> Void
 	) {
+		guard exposureDetection == nil else {
+			// exposureDetection was already set and is therefor running so there is nothing to do here.
+			// in the future someone should debug why this funtion is called twice in the first place.
+			return
```
marcussc, line 363: Shouldn't we call the completion? Otherwise the call will be stuck.

marcussc:
```suggestion
completion(.failure(.riskProviderIsRunning))
			return
```
Something like this?

freshking: @marcussc I thought that too but I wasn't sure in this case. I will add it.

marcussc (approved):

KaiTeuber (approved):

ArturFriesen (approved):

PR #2738, opened 2021-05-19
Feature/7535 a11y

KaiTeuber:
## Description
Accessibility to the vaccination health certificate flow

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7235


src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -2504,6 +2506,8 @@ Die Angaben werden statistisch ausgewertet, sie werden nicht zu einem Profil ges
 
 "HealthCertificate_Details_certificate" = "Impfzertifikat";
 
+"HealthCertificate_Details_QRCode_Image_Description" = "QR-Code mit Ihrem Impfzertifikat %i von %i";
```
InkenTownsend, line 2509:
```suggestion
"HealthCertificate_Details_QRCode_Image_Description" = "Ein QR-Code mit Ihrem Impfzertifikat %i von %i";
```

InkenTownsend: @KaiTeuber to stay consistent with the other entries, can you please add the "Ein" as suggested. Thank you :)

src/xcode/ENA/ENA/Source/Scenes/HealthCertificate/Cells/HealthCertificateQRCodeCellViewModel.swift
```diff
@@ -42,5 +44,5 @@ struct HealthCertificateQRCodeCellViewModel {
 	let certificate: String
 	let validity: String
 	let qrCodeImage: UIImage
-
+	let accessibility: String?
```
nickguendling, line 47:
```suggestion
	let accessibilityLabel: String?
```

InkenTownsend (approved):
UA approved

nickguendling (changes requested):
Just suggesting a little renaming to make it clearer which accessibility property is targeted

src/xcode/ENA/ENA/Source/Scenes/HealthCertificate/Cells/HealthCertificateQRCodeCellViewModel.swift
naveeddotio, line 47: Perhaps AccessibilityText will be more appropriate here.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificate/Cells/HealthCertificateQRCodeCellViewModel.swift
KaiTeuber: It's not a label, its a model, accessibilityText may fit better.

nickguendling: I don't necessarily agree 🤓😄 "Label" has two meanings, one is how it is used in `UILabel` -> A thing that has something written on it. The other meaning is "label" as the name or description for something. That's how it's used for the `.accessibilityLabel` of the `UIAccessibility` protocol. So I think it's totally valid to have `accessibilityLabel` as a property name in a model. But I won't block if you still prefer `accessibiliyText` 😄

"But that won't match with the whole model." Not sure what you mean by that.

naveeddotio: I initially suggested accessibilityText too (feels more like a model attribute) because that is primarily a text which one can assign to a label. Again, it’s a matter of taste.

KaiTeuber: I don't think a model should reflect what might be the use of it. I'll change it to `accessibiliyText` - this sounds like the best solution to me.

SwissMountainFrog (approved):

nickguendling (approved):

PR #2741, opened 2021-05-20
RAT Profile: DoB off one day

freshking:
## Description

Dont strip time when encoding date but store full etc date.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7263


nickguendling (changes requested):
The problem I see with this approach is that the saved date is still time zone dependent, so if you change the time zone your date of birth is displayed incorrectly.

I would still suggest to use UTC from start to finish to get rid of all time zone related issues.

freshking:
> The problem I see with this approach is that the saved date is still time zone dependent, so if you change the time zone your date of birth is displayed incorrectly.
> 
> I would still suggest to use UTC from start to finish to get rid of all time zone related issues.

I don't understand how changing the timezone changes the outcome here at all, since this is a UTC date. Please see screenshots.

Bildschirmfoto 2021-05-20 um 14 15 20
Bildschirmfoto 2021-05-20 um 14 14 07

freshking:
> Try the following steps to reproduce the time zone issue:
> 
> 1. Set time and date to today, 00:00
> 2. Add antigen test profile with date of birth 2020-05-20
> 3. Date of birth is shown as 2020-05-20
> 4. Set time zone to New York
> 
> Expected result: Date of birth is shown as 2020-05-20
> Actual result: Date of birth is shown as 2020-05-19
> 
> My proposed solution:
> 
> In `CreateAntigenTestProfileInputCell`:
> `birthdayPicker.timeZone = .utcTimeZone`
> 
> In `AntigenTestProfileViewModel`:
> 
> ```
> private var dateOfBirth: String? {
> 	guard let dateOfBirth = antigenTestProfile.dateOfBirth else {
> 		return nil
> 	}
> 
> 	let dateFormatter = DateFormatter()
> 	dateFormatter.timeZone = .utcTimeZone
> 	dateFormatter.dateStyle = .medium
> 	dateFormatter.timeStyle = .none
> 
> 	return String(
> 		format: AppStrings.ExposureSubmission.AntigenTest.Profile.dateOfBirthFormatText,
> 		dateFormatter.string(from: dateOfBirth)
> 	)
> }
> ```
> 
> In `CreateAntigenTestProfileViewController`:
> 
> ```
> let dateFormatter = DateFormatter()
> dateFormatter.timeZone = .utcTimeZone
> dateFormatter.dateStyle = .medium
> dateFormatter.timeStyle = .none
> 
> cell.textFields[2].text = dateFormatter.string(from: datePicker.date)
> ```

Doesn't the app throw an error and stop working if you change your timezone like that? And if so, isn't it a huge waste of time being concerned with such edge cases?

nickguendling:
I think the upside of my proposal is that we have a clean data model that only contains year, month, and day for the date of birth, correctly interpretable without any time zone information. It also has no known issues yet. Let me know if you found any! Do you see any downsides over your proposed solution or upsides of yours over mine?

I think it goes without discussion that we want to use the cleanest and most bug-free solution available to us. If you think that your proposed solution wins in that regard, feel free to argue for it!

src/xcode/ENA/ENA/Source/Models/Profile/AntigenTestProfile.swift
```diff
@@ -35,7 +35,7 @@ struct AntigenTestProfile: Codable {
 		try container.encodeIfPresent(firstName, forKey: .firstName)
 		try container.encodeIfPresent(lastName, forKey: .lastName)
 		if let dateOfBirth = dateOfBirth {
-			try? container.encodeIfPresent(ISO8601DateFormatter.justDate.string(from: dateOfBirth), forKey: .dateOfBirth)
+			try? container.encodeIfPresent(ISO8601DateFormatter().string(from: dateOfBirth), forKey: .dateOfBirth)
```
nickguendling, line 38: I think we can keep the `.justDate` formatter as we don't need the time information, same for the decoding below. It makes the date model a bit more precise

freshking: Done

nickguendling (approved):

marcussc (approved):

pascal-brause (approved):

PR #2742, opened 2021-05-20
Feature/7068 refactor launch arguments/7314 refactoring

naveeddotio:
## Description
PR to refactor launch arguments and separate them in one place with small description on what they do.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7314
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7315

## Screenshots
NA


src/xcode/ENA/ENA/Source/Scenes/TraceLocations/TraceLocationsCoordinator.swift
```diff
@@ -30,9 +30,7 @@ class TraceLocationsCoordinator {
 		
 		#if DEBUG
 		if isUITesting {
-			if let TraceLocationsInfoScreenShown = UserDefaults.standard.string(forKey: "TraceLocationsInfoScreenShown") {
-				store.traceLocationsInfoScreenShown = (TraceLocationsInfoScreenShown != "NO")
-			}
+			store.traceLocationsInfoScreenShown = UserDefaults.standard.bool(forKey: UITestingLaunchArguments.infoScreen.traceLocationsInfoScreenShown.remove(prefix: "-"))
```
marcussc, line 33: I saw this removePrefix a lot, but don't seem to understand why.
Can you explain it to me?

naveeddotio: So, we have launchArguments as strings which have "-" prefix in the start like "-riskLevel" but when we fetch the values from userDefaults, we get them by "riskLevel". I didn't want to have two files stores one with "-" and one without "-" which would have killed the whole idea of "single source of truth" for the launch arguments, so I am removing the prefix while fetching it from UserDefaults.

marcussc: And having the enum/launcharguments without the - does not work?

naveeddotio: Then I have always append "-" in the beginning of launch arguments and semantically launch arguments are always with a "-" so I left it like that.

marcussc: Okay understood

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/TraceLocations/TraceLocationsCoordinator.swift
pascal-brause: and couldn't you append the "-" as a computed property when reading out the enum value?
And extending `UserDefaults.standard.bool(forKey:` and adding every time the "-" there?
So the usage would be U`serDefaults.standard.bool(forKey: UITestingLaunchArguments.infoScreen.traceLocationsInfoScreenShown)` and under the hood, we add/remove the "-"?

pascal-brause, line 33: And another thing: When now using real bool, can't we use true/ false and not longer YES/NO? That's so ObjC....

nickguendling: Just a quick idea:

```
enum LaunchArguments {
	enum common {
		// Coming from ENF and used to set exposure notification to active
		static let ENStatus = LaunchArgument(name: "ENStatus")
		// Number of days the app is installed, shown on home / detail screens
		static let appInstallationDays = LaunchArgument(name: "appInstallationDays")
	}
}
```

```
struct LaunchArgument {

	let name: String

	var boolValue: Bool {
		UserDefaults.standard.bool(forKey: name)
	}

	var stringValue: String? {
		UserDefaults.standard.string(forKey: name)
	}

	var intValue: Int {
		UserDefaults.standard.integer(forKey: name)
	}

}
```

```
extension XCUIApplication {

	func setLaunchArgument(_ launchArgument: LaunchArgument, to value: String) {
		launchArguments.append(contentsOf: ["-\(launchArgument.name)", value])
	}

}
```

Setting in the UI test like this: `app.setLaunchArgument(LaunchArguments.common.ENStatus, to: ENStatus.active.stringValue)`
Accessing would work like this: `LaunchArguments.common.ENStatus.intValue`

pascal-brause (approved):

nickguendling (approved):

KaiTeuber (approved):

PR #2802, opened 2021-05-28
fix/#6608-data-beeing-downloaded-again

pascal-brause:
## Description
This should fix the bug of data being download again and again and again and....
The problem was, that empty `MetadataPackages` were not stored, so the cleanup could not cleanup the old empty packages and so client thought he has to re-download these packages. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6608


marcussc:
Approvals have been given in: https://github.com/corona-warn-app/cwa-app-ios/pull/2801

KaiTeuber (approved):

PR #2810, opened 2021-05-31
Fix/7554 wrong screen after TAN input

30mar:
## Description

- Remove test result available screen from TAN test registration and show directly the +ve test result if it exists
- Update UITests 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7554

## Screenshots
https://user-images.githubusercontent.com/15270737/120309210-45373280-c2d5-11eb-9acf-5ab5f079425f.MP4


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -515,12 +515,19 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 	}
 	
 	private func handleCoronaTestRegistrationInformation(_ testRegistrationInformation: CoronaTestRegistrationInformation, isLoading: @escaping (Bool) -> Void) {
+		var isConsentGiven: Bool
+		switch testRegistrationInformation {
+		case .teleTAN:
+			isConsentGiven = false
+		case .pcr, .antigen:
+			isConsentGiven = true
+		}
```
nickguendling, line 524:
```suggestion
	private func handleCoronaTestRegistrationInformation(_ testRegistrationInformation: CoronaTestRegistrationInformation, submissionConsentGiven: Bool, isLoading: @escaping (Bool) -> Void) {
```

I think I would prefer to set it on the calling side, because it feels like it should depend on the flow that one is coming from instead of the registration type. That's at least what I had in mind when we were talking earlier. But this solution is not wrong. I checked it out and tried it, worked correctly for me. 👍

If you want to leave it like it is, I'll approve, if you want to change it the way I suggested, I can check it out again :)

30mar: i will change it

freshking (approved):

SwissMountainFrog (approved):

whiskey (approved):

nickguendling (approved):

PR #2846, opened 2021-06-04
Possible fix for UI crash

freshking:
## Description

In this cell the subscriptions would keep getting added every time it was reused. Now they are cleared beforehand. This may fix a UI crash occurring sometimes.


pascal-brause (approved):
Do we not have a ticket for that?

freshking:
> Do we not have a ticket for that?



> Do we not have a ticket for that?

No, it's just something I found.

marcussc (approved):

ArturFriesen (approved):

PR #2853, opened 2021-06-07
Deleted persons and locations reappear after reopening the app

freshking:
## Description

Delete the referenced event checkin before deleting a location in the contact diary so no connection is left in the event store. 

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7588

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
@@ -263,7 +263,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		#else
 		return RiskProvider(
 			configuration: .default,
-			diaryStore: diaryStore,
+			store: store,
 			appConfigurationProvider: appConfigurationProvider,
 			exposureManagerState: exposureManager.exposureManagerState,
 			checkinRiskCalculation: checkinRiskCalculation,
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
@@ -263,7 +263,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		#else
 		return RiskProvider(
 			configuration: .default,
-			store: store,
+			diaryStore: diaryStore,
 			appConfigurationProvider: appConfigurationProvider,
 			exposureManagerState: exposureManager.exposureManagerState,
 			checkinRiskCalculation: checkinRiskCalculation,
--- a/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/EditEntries/DiaryEditEntriesViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ContactDiary/EditEntries/DiaryEditEntriesViewController.swift
@@ -11,11 +11,12 @@ class DiaryEditEntriesViewController: UIViewController, UITableViewDataSource, U
 
 	init(
 		entryType: DiaryEntryType,
-		store: DiaryStoringProviding,
+		diaryStore: DiaryStoringProviding,
+		eventStore: EventStoringProviding,
 		onCellSelection: @escaping (DiaryEntry) -> Void,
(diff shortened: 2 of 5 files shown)


src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -263,7 +263,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		#else
 		return RiskProvider(
 			configuration: .default,
-			store: store,
+			diaryStore: diaryStore,
```
pascal-brause, line 266: Wait, what? Do we have an extension on `RiskProvider` that allows passing instead of the `SecureStore` the `DiaryStore`?
Otherwise, I don't think that this will work because the `RiskProvider` needs a `SecureStore`, not a `DiaryStore`.
And also, I do not see in this PR why this is need for the `RiskProvider` (because there are no changes in him self).

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/EditEntries/DiaryEditEntriesViewModel.swift
```diff
@@ -12,10 +12,12 @@ class DiaryEditEntriesViewModel {
 
 	init(
 		entryType: DiaryEntryType,
-		store: DiaryStoringProviding
+		diaryStore: DiaryStoringProviding,
```
pascal-brause, line 15: good renaming!

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/EditEntries/DiaryEditEntriesViewController.swift
```diff
@@ -11,11 +11,12 @@ class DiaryEditEntriesViewController: UIViewController, UITableViewDataSource, U
...
 	init(
 		entryType: DiaryEntryType,
-		store: DiaryStoringProviding,
+		diaryStore: DiaryStoringProviding,
+		eventStore: EventStoringProviding,
 		onCellSelection: @escaping (DiaryEntry) -> Void,
 		onDismiss: @escaping () -> Void
 	) {
-		self.viewModel = DiaryEditEntriesViewModel(entryType: entryType, store: store)
+		self.viewModel = DiaryEditEntriesViewModel(entryType: entryType, diaryStore: diaryStore, eventStore: eventStore)
```
pascal-brause, line 19:
```suggestion
		self.viewModel = DiaryEditEntriesViewModel(
		    entryType: entryType, 
		    diaryStore: diaryStore, 
		    eventStore: eventStore
		)
```

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/EditEntries/DiaryEditEntriesViewModel.swift
```diff
@@ -81,25 +83,31 @@ class DiaryEditEntriesViewModel {
...
 	private var subscriptions: [AnyCancellable] = []
 
 	private func remove(entry: DiaryEntry) {
 		switch entry {
 		case .location(let location):
-			store.removeLocation(id: location.id)
+			// first remove checkin to this location
+			if let checkinID = location.visit?.checkinId {
+				eventStore.deleteCheckin(id: checkinID)
+			}
```
pascal-brause, line 106: Should there also be a warning for the user that this will delete the checkin, too? Was this specified by any PO?
I could imagine that some users will delete here the location but are not aware that this will delete the checkin, too?

naveeddotio (approved):

pascal-brause (changes requested):
Could you add a unit or ui test for this case?

ArturFriesen (changes requested):
I'm not sure about this fix. Is this approved by a product owner? I think we should not delete checkins if the corresponding location or person is deleted. We should just prevent location or persons from recreating.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
freshking: @pascal-brause This happened while automatic renaming. I already fixed it. Thanks for pointing this out.

freshking:
Closing for now since @30mar has fixed this issue a better way in a different PR.

PR #2878, opened 2021-06-09
Hide button when already warned others

naveeddotio:
## Description
We don't want to show remove test button if user has already warned others

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7691

## Screenshots
WhatsApp Image 2021-06-09 at 15 47 00


src/xcode/ENA/ENA/Source/Scenes/Home/Cells/TestResult/HomeShownPositiveTestResultCellModel.swift
```diff
@@ -60,6 +60,7 @@ class HomeShownPositiveTestResultCellModel {
 	@OpenCombine.Published var statusFootnote: String?
 	@OpenCombine.Published var homeItemViewModels: [HomeImageItemViewModel] = []
 	@OpenCombine.Published var isButtonHidden = false
+	@OpenCombine.Published var isRemoveTestButtonHidden = false
```
nickguendling, line 63:
```suggestion
	@OpenCombine.Published var isWarnOthersButtonHidden = false
	@OpenCombine.Published var isRemoveTestButtonHidden = false
```

nickguendling (approved):
Just an idea to make it clearer on the existing property which button it is referring to

KaiTeuber (approved):

ArturFriesen (approved):

PR #2913, opened 2021-06-14
Add failure logs to URL session requests

freshking:
## Description

Add failure logs to URL session requests

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7705


src/xcode/ENA/ENA/Source/Client/HTTP Client/URLSession+Convenience.swift
```diff
@@ -59,21 +59,31 @@ extension URLSession {
...
+				Log.error("No server request response", log: .api)
 				completion(.failure(.noResponse))
 				return
 			}
 
+			guard Response.acceptableStatusCodes ~= response.statusCode else {
+				Log.error("Invalid server response with code \(response.statusCode)", log: .api)
+				completion(.failure(.invalidResponse))
+				return
+			}
```
pascal-brause, line 77: This will pre-handle the status codes, but every response handling in our http methods does it already and proofs for more codes. So I think if you capture here the codes this will destroy the handling in the corresponding http calls and their specific logs.
Addionally: We have a ticket to refactor the hole http stuff. Perhaps we should do this there and in this PR just add Logs, nothing more.

pascal-brause (changes requested):

src/xcode/ENA/ENA/Source/Client/HTTP Client/URLSession+Convenience.swift
freshking: I specifically put this in here because I had an issue where 404 codes are not handled. Any response with data was considered successful.

src/xcode/ENA/ENA/Source/Client/HTTP Client/URLSession+Convenience.swift
```diff
@@ -103,7 +113,7 @@ extension URLSession {
 			type(of: self).acceptableStatusCodes.contains(statusCode)
 		}
 
-		private static let acceptableStatusCodes = (200 ... 299)
+		static let acceptableStatusCodes = (200 ... 299)
```
marcussc, line 116:
```suggestion
		fileprivate static let acceptableStatusCodes = (200 ... 299)
```

Just an Idea

marcussc (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/URLSession+Convenience.swift
ArturFriesen: Maybe it would make sense, to handle only the 404 specifically then?

ArturFriesen: Handling the 404 specifically would still be a Problem. For example, see "func dccRegisterPublicKey()" it handles the 404 in its own way. This handling would be dead code.

ArturFriesen: Please see this doc [(Debugging HTTP Server-Side Errors)](https://developer.apple.com/documentation/network/debugging_http_server-side_errors) about the difference between transportation and server errors.
At the moment the pattern is, that the extension is only handling the transportation errors.
The server errors are handled by the calling component.
This makes sense, because the server side errors are the ones which needed to be interpreted in context of the requirements.

ArturFriesen (changes requested):
PR breaks error handling.

ArturFriesen (approved):

pascal-brause (approved):

PR #2928, opened 2021-06-14
Fix/7854 Handle the result of registerHealthCertificate()

ArturFriesen:
## Description
Result of registerHealthCertificate was not handled.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7854


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -465,11 +465,21 @@ class HealthCertificateService {
...
 
 				remove(testCertificateRequest: testCertificateRequest)
-				registerHealthCertificate(base45: healthCertificateBase45)
-				completion?(.success(()))
+				let registerResult = registerHealthCertificate(base45: healthCertificateBase45)
+
+				switch registerResult {
+				case .success:
+					Log.info("[HealthCertificateService] Certificate assembly succeeded", log: .api)
+					completion?(.success(()))
```
nickguendling, line 475: I think the `remove(testCertificateRequest: testCertificateRequest)` needs to move down here so we don't delete the request in the error case. I think there was a problem in the unit tests that made me move it above the registration, we can discuss it if there are problems with this change.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
ArturFriesen: moved

pascal-brause (approved):

nickguendling (approved):

PR #2932, opened 2021-06-15
improve storage a bit

KaiTeuber:
## Description
Double entries may appear in the diary, improved check to get the latest state

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7702

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
@@ -674,7 +674,16 @@ class CoronaTestService {
 					}
 
 					// only store test result in diary if negative or positive
-					if (testResult == .positive || testResult == .negative) && !coronaTest.journalEntryCreated {
+					// Warning: check the current coronaTest so that changes are not overlooked
+					//
+					if let journalEntryCreated = self.coronaTest(ofType: coronaTestType)?.journalEntryCreated,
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -674,7 +674,16 @@ class CoronaTestService {
...
+					// Warning: check the current coronaTest so that changes are not overlooked
+					//
+					if let journalEntryCreated = self.coronaTest(ofType: coronaTestType)?.journalEntryCreated,
+					   (testResult == .positive || testResult == .negative) && !journalEntryCreated {
+						switch coronaTestType {
+						case .pcr:
+							self.pcrTest?.journalEntryCreated = true
+						case .antigen:
+							self.antigenTest?.journalEntryCreated = true
+						}
```
pascal-brause, line 686: why does this block has to be put above the `addCoronaTest` call?

pascal-brause (approved):

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
KaiTeuber: the add coronTest call will always succeed, to make sure it won't gat called twice and give the sql database some time

nickguendling (approved):

PR #2948, opened 2021-06-16
Quicktest Profile Tile gets not reset on app reset

freshking:
## Description

Reset antigen test profile publisher when wiping db

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7871


whiskey (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -45,6 +45,7 @@ final class SecureStore: Store, AntigenTestProfileStoring {
 	func wipeAll(key: String?) {
 		do {
 			try kvStore.wipeAll(key: key)
+			antigenTestProfileSubject.send(nil)
```
pascal-brause, line 48: I think this is the real fix. The changes in the `ExpsoureSubmissionCoordinator` are not necessary and confuses me more then it helps. Do we really need the changes in the `ExpsoureSubmissionCoordinator`?

freshking: The changes there are not necessary but I found that the store attribute is passed into that class unnecessarily. I just took it out to mitigate the confusion.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -155,7 +153,7 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 					self?.showAntigenTestProfile()
 				}
 			},
-			antigenTestProfileStore: antigenTestProfileStore
+			antigenTestProfileStore: store
```
ArturFriesen, line 156: We want to reduce the usage of RequiresAppDependencies, because its tight coupling.
The injection of antigenTestProfileStore allows us to provide different implementations at runtime.

ArturFriesen (dismissed):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
freshking: This was not used in any tests. Should we still leave it?

ArturFriesen: Yes, we want to get rid of RequiresAppDependencies, so we should not use it.

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
freshking: Reverted

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
freshking: Done

marcussc (approved):

PR #2949, opened 2021-06-16
Test Certificate Details - First Name Last Name instead of other way around

freshking:
## Description

Reverse name order in certificate

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7823


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
```diff
@@ -8,28 +8,41 @@ import HealthCertificateToolkit
 extension Name {
 
 	var fullName: String {
-		var givenName = self.givenName
-		var familyName = self.familyName
+		return [resolvedGivenName, resolvedFamilyName].formatted()
+	}
+	
+	var reversedFullName: String {
+		return [resolvedFamilyName, resolvedGivenName].formatted()
```
nickguendling, line 15: I think you need an additional comma in the separator in this case

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
freshking: Not according to the ticket

nickguendling: I see the comma in the ticket:

> **Expected result:**
> Last Name, First Name
> **Actual result:**
> First Name Last Name

nickguendling: I don't see example screenshots, only the photo of the devices with the wrong formatting.

Regardless, I found that the tech spec has been updated with the new desired formatting: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/7e46800f8d9b4c0f0efb09d624960a6476b2f534/docs/spec/dgc-overview-client.md#determining-the-formatted-name-of-the-person-from-a-dgc

pascal-brause (changes requested):
I think Nick is right here, the comma is missing.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
freshking: Done

ArturFriesen (approved):

pascal-brause (approved):

nickguendling (approved):

PR #2951, opened 2021-06-16
Fix/#7862 black white screen

pascal-brause:
## Description
This fixes perhaps the problem of black/white screen at startup.
Suggestion for this fix is that the store property isOnboarded could not be read and is set to its fallback (false), which can be read in the log files.
The flag is read out at the refactored occurrence in didFinish() and should return something, but did not so.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7862

## Screenshots
x


marcussc (approved):

freshking (approved):

KaiTeuber (approved):

whiskey (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -120,10 +120,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		DeadmanNotificationManager(coronaTestService: coronaTestService).scheduleDeadmanNotificationIfNeeded()
 
 		consumer.didFailCalculateRisk = { [weak self] error in
-			guard self?.store.isOnboarded == true else {
-				return
+			if self?.store.isOnboarded == true {
+				self?.showError(error)
 			}
-			self?.showError(error)
```
ArturFriesen, line 126: Just wondering what is the difference now? The logic flow is the same isn't it? 🤔

mlenkeit:
@pascal-brause could you reproduce this by manipulating the state of the device (prior to the fix)?

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
pascal-brause: No, because in the guard we just returned without nothing, but in the normal flow, we expect `handleQuickActions(with: launchOptions)`

pascal-brause:
> @pascal-brause could you reproduce this by manipulating the state of the device (prior to the fix)?

Unfortunately no. I tried it to corrupt the isOnboarded (because in the log files we see that there is a problem with it, but I got not stuck. 
See this fix as a try.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
ArturFriesen: handleQuickActions will not be reached either way. Because the return is in a closure.
But I think the if is better readable. So let's keep it.

pascal-brause: you are completely correct.

PR #2962, opened 2021-06-17
Feature/7648 family/7840 preferred person and certificate

nickguendling:
## Description
Adds a preferred person flag and the most relevant health certificate to the `HealthCertifiedPerson`.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7840


KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/__test__/HealthCertificateArrayMostRelevantTests.swift
```diff
@@ -0,0 +1,200 @@
...
+			olderSeriesCompletingVaccinationCertificate,
+			mostRecentOtherVaccinationCertificate,
+			olderOtherVaccinationCertificate,
+			mostRecentOutdatedRecoveryCertificate,
+			olderOutdatedRecoveryCertificate,
+			mostRecentOutdatedPCRTest,
+			olderOutdatedPCRTest,
+			mostRecentOutdatedAntigenTest,
+			olderOutdatedAntigenTest
+		].shuffled()
```
marcussc, line 59: Finally a usecase for `shuffled()` 😁

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertificate.swift
```diff
@@ -87,6 +93,17 @@ struct HealthCertificate: Codable, Equatable, Comparable {
 	}
 
 	var type: CertificateType {
```
ArturFriesen, line 95: Whats the reason for having type? Cant we just only use entry? If you read the entry and you don't need the associated value, you can omit it.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertificate.swift
nickguendling: With associated value I couldn't just write e.g. `$0.entry == .test`, as the default `Equatable` implementation always requires an associated value to equate to. And the `if case .test = $0.entry` variant is very inconvenient for `.filter`ing etc.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertificate.swift
ArturFriesen: Ok, I unterstand!

ArturFriesen (approved):

PR #2977, opened 2021-06-21
CoronaTestService Refactoring

freshking:
## Description

Introducing the `AsyncOperation` class that can be subclassed to wrap any asynchronous task and execute it serially on an operation queue. Only when the asynchronous task is marked as finished will it be completed and the next task on the queue is executed.

The first task that now uses this schema is the `getTestResult()` function in the `CoronaTestService` class.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7868


KaiTeuber (changes requested):
I think this solution is only one part of the problem here. Dispatching to a serial queue will resolve a race condition. But we also have the problem, that structs are used. So not every update leads to the expected behavior - we get copies but expect the values to get updated correctly. We need to find a mechanism to provide clean updates. You can see on the failing unit test that things are not right now.
Maybe we need to think about moving from structs to a classes here?

freshking:
> I think this solution is only one part of the problem here. Dispatching to a serial queue will resolve a race condition. But we also have the problem, that structs are used. So not every update leads to the expected behavior - we get copies but expect the values to get updated correctly. We need to find a mechanism to provide clean updates. You can see on the failing unit test that things are not right now.
> Maybe we need to think about moving from structs to a classes here?

The tests that are failing now is because they themselves are faulty. I am fixing them at the moment. How would switching to classes help here? When there are no more race conditions, how do we run into stale data?

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
```diff
@@ -1677,7 +1677,8 @@ class CoronaTestServiceTests: CWATestCase {
 		}
 		waitForExpectations(timeout: .short)
 
-		XCTAssertEqual(mockNotificationCenter.notificationRequests.count, 2)
+		// has to be 6 because each call to "updateTestResults()" add one notification for antigen and one for pcr
```
ArturFriesen, line 1680: Are you sure that it should be 6?
Another comment above states this "// Updating two more times to check that notification are only scheduled once".
Looks like the notifications are expected to only trigger once .

src/xcode/ENA/ENA/Source/Models/Async Operation/AsyncOperation.swift
```diff
@@ -0,0 +1,85 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+class AsyncOperation: Operation {
```
ArturFriesen, line 7: Why is it called "Async"? I don't see the property "isAsynchronous" set to true.
As far as I know, if its not set, then the operation will be executed synchronous if its called directly or the queue will decide about async or sync.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
freshking: Yes, it's because of the implementation of the `MockUserNotificationCenter`. It saves each notification request to the array when presenting it. Thats all the logic inside of it. So therefore it has to be 6 because `updateTestResults(presentNotification: true)` is called three times and that presents one notification for PCR and one for Antigen. The assumption it would only save 2 requests was wrong.

src/xcode/ENA/ENA/Source/Models/Async Operation/AsyncOperation.swift
freshking: Because it only sets the operation to completed when the task is finished. It allows you to execute async operations serially. Otherwise you could add operations serially to a queue and they would execute in that order but they would not wait for the previous operation to complete first.

freshking: With this new class we can get rid of all the completion block inceptions we have throughout the app. You could just chain these operations together.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
ArturFriesen: This test is not testing MockUserNotificationCenter. The mock is used to figure out wether the CoronaTestService will trigger the right amount of notifications.

Looked into details of CoronaTestService and need to block this PR.
Your change is altering the expected behavior.
It could lead to redundant triggered notifications (and maybe also other unexpected behavior).

ArturFriesen (dismissed):
change is altering the expected behavior

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/CoronaTestServiceTests.swift
ArturFriesen: This failing test also indicates, that the operations are not executed serially.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -698,7 +699,7 @@ class CoronaTestService {
 
 					}
 
-					if coronaTest.finalTestResultReceivedDate == nil {
+					if self.coronaTest(ofType: coronaTestType)?.finalTestResultReceivedDate == nil {
```
freshking, line 702: @ArturFriesen This was the reason the test was failing. It was referencing a stale value. With this change it now works as expected.

ArturFriesen: Ok nice!

ArturFriesen: Referring to @KaiTeuber comment, this shows, that there is still the problem with stale data.

freshking:
> I think this solution is only one part of the problem here. Dispatching to a serial queue will resolve a race condition. But we also have the problem, that structs are used. So not every update leads to the expected behavior - we get copies but expect the values to get updated correctly. We need to find a mechanism to provide clean updates. You can see on the failing unit test that things are not right now.
> Maybe we need to think about moving from structs to a classes here?

@KaiTeuber This bug is also fixed with this PR:

IMG_3102

naveeddotio (approved):

KaiTeuber (approved):

ArturFriesen (approved):

PR #2990, opened 2021-06-22
Vaccination certificate accepted with empty mandatory fields

freshking:
## Description

All three field "vp", "mp" an "ma" must be non-empty. Updated schema validation for this.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8012


30mar (approved):

naveeddotio (approved):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Model/DigitalGreenCertificate.swift
```diff
@@ -77,6 +77,37 @@ public struct VaccinationEntry: Codable, Equatable {
...
+        vaccineMedicinalProduct = try container.decode(String.self, forKey: .vaccineMedicinalProduct)
+        marketingAuthorizationHolder = try container.decode(String.self, forKey: .marketingAuthorizationHolder)
+        doseNumber = try container.decode(Int.self, forKey: .doseNumber)
+        totalSeriesOfDoses = try container.decode(Int.self, forKey: .totalSeriesOfDoses)
+        dateOfVaccination = try container.decode(String.self, forKey: .dateOfVaccination)
+        countryOfVaccination = try container.decode(String.self, forKey: .countryOfVaccination)
+        certificateIssuer = try container.decode(String.self, forKey: .certificateIssuer)
+        uniqueCertificateIdentifier = try container.decode(String.self, forKey: .uniqueCertificateIdentifier)
+        // EU Spec says "Exactly 1 (one) non-empty "vp", "mp" or "ma" field MUST be provided"
+        if vaccineOrProphylaxis.isEmpty && vaccineMedicinalProduct.isEmpty && marketingAuthorizationHolder.isEmpty {
```
ArturFriesen, line 107: I think it would be easier to do this check in "convertCBORToStruct" after the struct was decoded?
Because its not really a decoding issue.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Model/DigitalGreenCertificate.swift
```diff
@@ -77,6 +77,37 @@ public struct VaccinationEntry: Codable, Equatable {
...
+            throw CertificateDecodingError.HC_JSON_SCHEMA_INVALID(.VALIDATION_FAILED(Error.oneNonEmptyFieldNotProvided))
```
ArturFriesen, line 108: I think it could be confusing that the error is thrown as a `HC_JSON_SCHEMA_INVALID` error here.
Because its not really arise from the json validation.

I would suggest to create a separate error. Something like `HC_ENTRY_MISSING` and put it into `CertificateDecodingError` enum.

ArturFriesen (changes requested):
Please see comments.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Model/DigitalGreenCertificate.swift
mlenkeit: @freshking I would also rather adjust the JSON schema here to fix this... with a simple pattern such as `.+`, it should work fine.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Ressources/CertificateSchema.json
```diff
@@ -149,15 +149,18 @@
         },
         "vp": {
           "description": "vaccine or prophylaxis",
-          "$ref": "#/$defs/vaccine-prophylaxis"
+          "$ref": "#/$defs/vaccine-prophylaxis",
+          "minLength": 1
         },
```
ArturFriesen, line 154: @mlenkeit Are you gonna update this also in the tech-spec?

ArturFriesen (approved):

PR #2999, opened 2021-06-24
PCR Redeemed has issue with certificate screenflow

freshking:
## Description

Minor logic fix to secondary button deciding when it should be disabled.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7892


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/TestCertificate/ExposureSubmissionTestCertificateViewController.swift
```diff
@@ -64,7 +64,7 @@ class ExposureSubmissionTestCertificateViewController: DynamicTableViewControlle
 			didTapSecondaryButton { [weak self] isLoading in
 				guard let self = self else { return }
 
-				self.footerView?.setLoadingIndicator(false, disable: isLoading ? true : self.viewModel.isPrimaryButtonEnabled, button: .primary)
+				self.footerView?.setLoadingIndicator(false, disable: isLoading ? true : !self.viewModel.isPrimaryButtonEnabled, button: .primary)
```
ArturFriesen, line 67: I don't understand how this fix is resolving the issue. The state of `isPrimaryButtonEnabled` is not changed based on the test result only after the birthday changes.

ArturFriesen (dismissed):

ArturFriesen:
Link to Jira is wrong.

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/TestCertificate/ExposureSubmissionTestCertificateViewController.swift
freshking: Yes but this code is only called when selecting the secondary button. And then the state of the primary button is updated and it is incorrect that is should be disabled if `isPrimaryButtonEnabled` == true

ArturFriesen: So the logic is not depending in some other state, like testresult, but only if `isPrimaryButtonEnabled` was enabled, it should we disabled after loading? What about errors which could happen and after the error, the button should be still enabled for retry?

freshking: This fix is handling the error state. The user tries to register a redeemed PCR test and gets an error popup. Check out the branch and try it out.

ArturFriesen: You are right! I was confused by the other wrong code line. Which you have also fixed. Thanks!

ArturFriesen (approved):

nickguendling (approved):

PR #3002, opened 2021-06-24
Feature/7087 statistic data local 7 day incidence/7692 fetch local statistics

naveeddotio:
## Description
API fetching and caching for local statistics

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7692

## Screenshots
N/A


src/xcode/ENA/ENA/Source/Services/LocalStatistics/LocalStatisticsProvider.swift
```diff
@@ -0,0 +1,73 @@
...
+	private let client: LocalStatisticsFetching
+	private let store: LocalStatisticsCaching
+
+	private func fetchLocalStatistics(administrativeUnit: String, eTag: String? = nil) -> Future<SAP_Internal_Stats_LocalStatistics, Error> {
+		return Future { promise in
+			self.client.fetchLocalStatistics(administrativeUnit: administrativeUnit, eTag: eTag) { result in
+				switch result {
+				case .success(let response):
+					// cache
+					self.store.localStatistics = LocalStatisticsMetadata(with: response)
```
30mar, line 42: The response of this API is usually very big as it has several States with tens of districts and we are only trying to fetch one district so it will be better to filter the response and only save requested District .
other with we have to filter the whole response every time we open the app

30mar (changes requested):

src/xcode/ENA/ENA/Source/Services/LocalStatistics/LocalStatisticsProvider.swift
naveeddotio: Good point, let's discuss in our alignment call tomorrow.

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -173,6 +186,44 @@ class CachingHTTPClient: AppConfigurationFetching, StatisticsFetching, QRCodePos
...
+
+	// MARK: LocalStatisticsFetching
+	
+	/// Fetches local statistics
+	/// - Parameters:
+	///   - administrativeUnit: string to pass administrative unit to the API to get local statistics
+	///   - etag: an optional ETag to download only versions that differ the given tag
+	///   - completion: result handler
+	func fetchLocalStatistics(
+		administrativeUnit: String,
```
whiskey, line 198: I would suggest we create a `typealias` for *administrativeUnit* to simplify function calls later:
```
typealias AdministrativeUnit = String
```
This should prevent devs from putting random strings into functions like this.

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -173,6 +186,44 @@ class CachingHTTPClient: AppConfigurationFetching, StatisticsFetching, QRCodePos
...
+		eTag: String?,
+		completion: @escaping LocalStatisticsCompletionHandler
+	) {
+		// Manual ETagging because we don't use native cache
```
whiskey, line 202: Life could be so much better without tinfoil…

src/xcode/ENA/ENA/Source/Client/__tests__/Mocks/CachingHTTPClientMock.swift
```diff
@@ -48,6 +48,18 @@ final class CachingHTTPClientMock: CachingHTTPClient {
...
 		return configMetadata
 	}()
+	
+	static let staticLocalStatistics: SAP_Internal_Stats_LocalStatistics = {
+		guard
+			let url = Bundle(for: CachingHTTPClientMock.self).url(forResource: "default_local_statistics", withExtension: "bin"),
+			let data = try? Data(contentsOf: url),
+			let localStatistics = try? SAP_Internal_Stats_LocalStatistics(serializedData: data)
+		else {
+			fatalError("Cannot initialize static test data")
```
whiskey, line 58: Please don't use `fatalError` in tests! This might/will kill the test runner and lead to all kinds of follow up issues. Create a function that throws an error in case of resources not found.

src/xcode/ENA/ENA/Source/Services/LocalStatistics/LocalStatisticsMetadata.swift
```diff
@@ -0,0 +1,64 @@
...
+		
+	func encode(to encoder: Encoder) throws {
+		var container = encoder.container(keyedBy: CodingKeys.self)
+		
+		try container.encode(lastLocalStatisticsETag, forKey: .lastLocalStatisticsETag)
+		try container.encode(lastLocalStatisticsFetchDate, forKey: .lastLocalStatisticsFetchDate)
+
+		let localStatistics = try localStatistics.serializedData()
+		try container.encode(localStatistics, forKey: .localStatistics)
+	}
```
whiskey, line 53: Lesson learned: Protobuf should implement (and generate) the`Codable` protocol to prevent this mess :(

src/xcode/ENA/ENA/Source/Services/LocalStatistics/LocalStatisticsProvider.swift
whiskey: @30mar define 'big'? Caching the latest state for all districts seems tempting if we can reduce the number of follow up calls – given the server content (ETag) didn't change.

whiskey (changes requested):
Small issues.

src/xcode/ENA/ENA/Source/Client/HTTP Client/CachingHTTPClient.swift
```diff
@@ -195,7 +197,7 @@ class CachingHTTPClient: AppConfigurationFetching, StatisticsFetching, LocalStat
 	///   - etag: an optional ETag to download only versions that differ the given tag
 	///   - completion: result handler
 	func fetchLocalStatistics(
-		administrativeUnit: String,
+		administrativeUnit: AdministrativeUnit,
```
whiskey, line 200: ❤️

whiskey (approved):

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/LocalStatistics/LocalStatisticsProvider.swift
```diff
@@ -0,0 +1,73 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+import OpenCombine
+
+class LocalStatisticsProvider: LocalStatisticsProviding {
```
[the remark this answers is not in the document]
naveeddotio: Both are totally different things, no connection whatsoever, discussed it with Nick as well and decided to separate it from other statistics.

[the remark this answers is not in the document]
30mar: ok

30mar (approved):

PR #3030, opened 2021-06-29
UA Bugfix (EXPOSUREAPP-8155)

NataliaLemmerth:
Alignment to Android text change.
https://github.com/corona-warn-app/cwa-app-android/pull/3576

### DIFF
--- a/src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
+++ b/src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
@@ -2810,7 +2810,7 @@ Die Angaben werden statistisch ausgewertet, sie werden nicht zu einem Profil ges
 
 /* Recovery Certificate Delete Alert */
 
-"RecoveryCertificate_Alert_title" = "Wollen Sie das Genesenenzertifikat wirklich entfernen?";
+"RecoveryCertificate_Alert_title" = "Wollen Sie das Zertifikat entfernen?";
 
 "RecoveryCertificate_Alert_message" = "Wenn Sie das Genesenenzertifikat entfernen, können Sie es nicht mehr als Nachweis in der App verwenden.";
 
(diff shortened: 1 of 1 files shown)


ChristinaTsigganos (approved):
approved

naveeddotio (approved):

pascal-brause (approved):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -2810,7 +2810,7 @@ Die Angaben werden statistisch ausgewertet, sie werden nicht zu einem Profil ges
 
 /* Recovery Certificate Delete Alert */
 
-"RecoveryCertificate_Alert_title" = "Wollen Sie das Genesenenzertifikat wirklich entfernen?";
+"RecoveryCertificate_Alert_title" = "Wollen Sie das Zertifikat entfernen?";
```
nickguendling, line 2813: This is inconsistent with the other titles above

nickguendling (changes requested):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
NataliaLemmerth: Hi Nick, yes you are right. We'll think of writing 'Wollen Sie das Zertifikat entfernen?' for all occurrences, so that it's consistent. But here we have an example of a shorter text. https://github.com/corona-warn-app/cwa-app-ios/blob/release/2.5.x/src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings#L2655

We try always to adjust both Android and iOS texts. If the texts are different and we have to change them in the future, it's not so easy to find these texts. And besides, tester care a lot about the text consistency in both Ios and Android and if there are differences, the bugs will be opened.

NataliaLemmerth: @nickguendling Would it be okay like this: 'Wollen Sie das Genesenenzertifikat entfernen?' ?

NataliaLemmerth:
I've closed this PR. I want to discuss with my UA colleagues how we are going to handle such text changes in future. Text in Android has been adjusted and it's okay for now.

PR #3036, opened 2021-06-29
Feature/#8024 dcc validation service stub

pascal-brause:
## Description
Brings in the validation service and its protocol so that the UI can call this already.
Adds also some Error and Result data structures.

## Link to Jira
x

## Screenshots
x


src/xcode/ENA/ENA/Source/Services/DCCValidationService/Model/DCCValidationProgress.swift
```diff
@@ -0,0 +1,15 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+struct DCCValidationProgress {
```
nickguendling, line 7: Isn't this more of a `DCCValidationResult`? Or maybe `DCCValidationReport`, then we don't have the naming confusion with the Swift `Result` type

src/xcode/ENA/ENA/Source/Services/DCCValidationService/Model/DCCValidationError.swift
```diff
@@ -0,0 +1,30 @@
...
+	case INVALIDATION_RULE_CLIENT_ERROR (DCCValidationProgress)
+	case INVALIDATION_RULE_JSON_ARCHIVE_FILE_MISSING (DCCValidationProgress)
+	case INVALIDATION_RULE_JSON_ARCHIVE_SIGNATURE_INVALID (DCCValidationProgress)
+	case INVALIDATION_RULE_JSON_EXTRACTION_FAILED (DCCValidationProgress)
+	case INVALIDATION_RULE_SERVER_ERROR (DCCValidationProgress)
+	case ONBOARDED_COUNTRIES_CLIENT_ERROR (DCCValidationProgress)
+	case ONBOARDED_COUNTRIES_JSON_ARCHIVE_FILE_MISSING (DCCValidationProgress)
+	case ONBOARDED_COUNTRIES_JSON_ARCHIVE_SIGNATURE_INVALID (DCCValidationProgress)
+	case ONBOARDED_COUNTRIES_JSON_EXTRACTION_FAILED (DCCValidationProgress)
+	case ONBOARDED_COUNTRIES_SERVER_ERROR (DCCValidationProgress)
```
nickguendling, line 26: Maybe we can extract these into a `DCCOnboardedCountriesError` and double up the `NO_NETWORK` error for that

src/xcode/ENA/ENA/Source/Services/DCCValidationService/DCCValidationService.swift
```diff
@@ -0,0 +1,66 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+import OpenCombine
+import HealthCertificateToolkit
+
+protocol DCCValidationProviding {
+	func availableCountries(
```
nickguendling, line 10:
```suggestion
	func onboardedCountries(
```

To match the `ONBOARDED_COUNTRIES` naming from the errors we could rename this

nickguendling (changes requested):
Just two naming proposals and an error structure idea

src/xcode/ENA/ENA/Source/Services/DCCValidationService/Model/DCCValidationProgress.swift
pascal-brause: We were not sure, but because we can return this thing also when it is not reached to the rule set validations, we thought more about a progress. Yeah but to avoid swift keywords, `Report` would be fine for me, too.

src/xcode/ENA/ENA/Source/Services/DCCValidationService/Model/DCCValidationError.swift
pascal-brause: I will have a look.

pascal-brause: Okay that makes sense to extract all the onboarded_countries_errors out from here to separate it for the 2 http calls.

nickguendling (approved):

ArturFriesen (approved):

freshking (approved):

PR #3059, opened 2021-07-01
Feature/#8096 validation onboarded countries

pascal-brause:
## Description
Part of the validation service: To request and return the onboarded countries, including caching.

## Link to Jira
Main Story:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8024
Subtask:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8208
TechSpec:
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/business-rules-dcc/docs/spec/dgc-validation-rules-client.md#downloading-onboarded-countries

## Screenshots
x


src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/HealthCertificateValidationService.swift
```diff
@@ -0,0 +1,175 @@
...
+	
+	// MARK: - Internal
+	
+	// MARK: - Private
+	
+	private let store: Store
+	private let client: Client
+	private let signatureVerifier: SignatureVerification
+
+	private func onboardedCountriesSuccessHandler(
```
ArturFriesen, line 88:
```suggestion
	private func processOnboardedCountriesResponse(
```

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/HealthCertificateValidationService.swift
```diff
@@ -0,0 +1,175 @@
...
+				store.validationOnboardedCountriesCache = receivedOnboardedCountries
+				completion(.success(countries))
+			case let .failure(error):
+				Log.error("Could not decode CBOR from package with error:", error: error)
+				completion(.failure(error))
+			}
+		})
+	}
+	
+	private func onboardedCountriesFailureHandler(
```
ArturFriesen, line 134:
```suggestion
	private func processOnboardedCountriesFailure(
```

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -615,16 +615,16 @@ final class HTTPClient: Client {
 		})
 	}
 	
-	func getDCCOnboardedCountries(
+	func validationOnboardedCountries(
```
ArturFriesen, line 618: All other "caching" requests are part of `CachingHTTPClient`.
For consistency reasons, can we also move this there?

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/Model/ValidationOnboardedCountriesCache.swift
```diff
@@ -0,0 +1,43 @@
...
+        onboardedCountries: [Country],
+        lastOnboardedCountriesETag: String
+    ) {
+        self.onboardedCountries = onboardedCountries
+        self.lastOnboardedCountriesETag = lastOnboardedCountriesETag
+    }
+    
+    // MARK: - Overrides
+    
+    // MARK: - Protocol Codable
```
ArturFriesen, line 21: I guess implementing codable is not needed here, it should be inferred by the compiler.

ArturFriesen (changes requested):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
```diff
@@ -6,6 +6,12 @@ import Foundation
 import SwiftCBOR
 
 public struct OnboardedCountriesAccess {
+    
+    // MARK: - Init
+
+    public init() {}
+    
+    // MARK: - Public
```
whiskey, line 14: Why is `extractCountryCodes` not static then? This would prevent this initializer.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesCBORDataFake.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public var onboardedCountriesCorruptCBORDataFake: Data {
+    let cborCountries = CBOR.array(
+        [
+            CBOR.null,
+            CBOR.unsignedInt(42)
+        ]
+    )
+    return Data(cborCountries.encode())
+}
```
whiskey, line 26: Do we really need global variables for a very limited scope? If this is just used in the test below, `internal` is fine.

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/Model/ValidationOnboardedCountriesCache.swift
pascal-brause: It is needed for the SecureStore.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
pascal-brause: Hm I only added it to be consistent to the rest in the ToolKit. Why all these Access struct does not have static methods: I don't know. @ArturFriesen what was your approach here?

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesCBORDataFake.swift
pascal-brause: yeah, true 👍

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
ArturFriesen: Because there is a big chance that we want to have a mock for that and replace it during runtime, for example for testing. And it increases composability.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/Model/ValidationOnboardedCountriesCache.swift
ArturFriesen: Yes right, codable is needed, but do you need to implement it? Isn't it enough to just add the protocol?

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
ArturFriesen: I even think the other way, we should put a protocol on top and inject it into the service. But lets not do it on this PR. There will be others following with more testing which will require this.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
```diff
@@ -6,6 +6,12 @@ import Foundation
 import SwiftCBOR
 
 public struct OnboardedCountriesAccess {
+    
+    // MARK: - Init
+
+    public init() {}
```
ArturFriesen, line 12: Are there now 2 public inits? 🙈

pascal-brause: oh yeah 😅

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Validation/OnboardedCountriesAccess.swift
pascal-brause: yeah, sounds like a plan (Protocol + other PR).

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
pascal-brause: I implemented it in the "normal" client, because:
1. The CachingHTTPClient does not handle the hole fake request stuff so far. So I had to copy a lot from the normal client over to the CachingHTTPClient
2. The error handling of the tech specs is more familiar with the error handling of the normal client. If I move this to the CachingHTTPClient, the error handling would be much more different then the other calls.
3. We have a ticket to refactor the client and the caching client to make these clients more beautiful. We could clean up this here in that ticket, too.

whiskey (approved):

ArturFriesen (approved):

30mar (approved):

PR #3071, opened 2021-07-02
Fix/8217 app crash re-checkin

ArturFriesen:
## Description
Use recent indexpath instead of outdated one.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8217


marcussc (approved):

nickguendling (commented):
Can you check the `CheckinsOverviewViewController` (List of check-ins) if we have the same issue there, too? It's very similar in structure, so I think we might also use the wrong index path on button taps there.

naveeddotio (approved):

ArturFriesen:
> Can you check the `CheckinsOverviewViewController` (List of check-ins) if we have the same issue there, too? It's very similar in structure, so I think we might also use the wrong index path on button taps there.

Fixed it there too!

nickguendling (approved):

PR #3084, opened 2021-07-02
Fix/certificate header optionality

ArturFriesen:
## Description
'iat' does not need to be optional, because it cant be nil.
Changed it from optional to non-optional.


mlenkeit (approved):
LGTM

whiskey (approved):

pascal-brause (approved):

PR #3109, opened 2021-07-06
Feature/8025 dcc validation ux/8095 coordinator

nickguendling:
## Description
Adds the validation coordinator and a dummy view controller as a starting point for the screen implementations

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8095

## Screenshots
IMG_0133
IMG_0134


Ein-Tim:
Wrong milestone?

nickguendling:
> Wrong milestone?

Yep, used the wrong one, corrected it now. Thanks!

ChristinaTsigganos (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/HealthCertificateValidationCoordinator.swift
```diff
@@ -0,0 +1,150 @@
...
+
+import UIKit
+import class CertLogic.ValidationResult
+
+final class HealthCertificateValidationCoordinator {
+	
+	// MARK: - Init
+	
+	init(
+		parentViewController: UIViewController,
```
ArturFriesen, line 13: Would it be possible to expose a root viewcontroller and let the calling coordinator decide how to present the root of this coordinator. Just hypothetically the coordinator then could be used in different navigation scenarios.

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/HealthCertificateValidationCoordinator.swift
```diff
@@ -0,0 +1,150 @@
...
+						switch error {
+						case .TECHNICAL_VALIDATION_FAILED:
+							self.showTechnicalErrorScreen()
+						default:
+							self.showErrorAlert(
+								title: AppStrings.HealthCertificate.ValidationError.title,
+								error: error
+							)
+						}
+					}
```
pascal-brause, line 91: 👌

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/Model/HealthCertificateValidationError.swift
```diff
@@ -22,4 +22,35 @@ enum HealthCertificateValidationError: Error {
...
+			return "\(AppStrings.HealthCertificate.ValidationError.tryAgain) (INVALIDATION_RULE_JSON_ARCHIVE_SIGNATURE_INVALID)"
+		case .INVALIDATION_RULE_JSON_EXTRACTION_FAILED:
+			return "\(AppStrings.HealthCertificate.ValidationError.tryAgain) (INVALIDATION_RULE_JSON_EXTRACTION_FAILED)"
+		case .INVALIDATION_RULE_SERVER_ERROR:
+			return "\(AppStrings.HealthCertificate.ValidationError.tryAgain) (INVALIDATION_RULE_SERVER_ERROR)"
+		case .NO_NETWORK:
+			return "\(AppStrings.HealthCertificate.ValidationError.noNetwork) (NO_NETWORK)"
+		case .VALUE_SET_SERVER_ERROR:
+			return "\(AppStrings.HealthCertificate.ValidationError.tryAgain) (VALUE_SET_SERVER_ERROR)"
+		}
```
pascal-brause, line 53: Note: While writing on the rule validation, it turned out that there will be two more error cases, for each rule type one. I will add a jira subtask as a reminder that we don't forget to add them here.

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/HealthCertificateValidationCoordinator.swift
nickguendling: I used the same pattern here that we use for the exposure submission coordinator. I see the theoretical freedom of exposing the root view controller, but in the end the whole flow needs to be either presented modally or pushed. And that exposed root view controller would need to already be a navigation controller, and couldn't be pushed onto another one. So the only viable option is to present it modally, and there I don't see an up- or downside for both approaches. So I'm just gonna leave it like it is for now.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidationService/Model/HealthCertificateValidationError.swift
nickguendling: I think we will run into a compile-time error anyway if we are missing cases, but it's good to have the task just to make sure 👍

PR #3112, opened 2021-07-07
Add correct message in case of no network

30mar:
## Description
Add readable error message for no network connection in error reports

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7193

## Screenshots
IMG_DEE4689084D4-1


NataliaLemmerth (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/ErrorReporting/ErrorReportCoordinator.swift
```diff
@@ -167,7 +167,16 @@ final class ErrorReportsCoordinator: ErrorReportsCoordinating, RequiresAppDepend
...
 					case .failure(let error):
 						Log.error("ELS submission error: \(error)", log: .els, error: error)
-						self.showErrorAlert(with: error)
+						switch error {
+						case .otpError(let otpError):
+							if otpError == .noNetworkConnection {
+								self.showErrorAlert(with: AppStrings.Common.noNetworkConnection)
+							} else {
+								self.showErrorAlert(with: error.localizedDescription)
+							}
```
pascal-brause, line 176:
```suggestion
						otpError == .noNetworkConnection ? self.showErrorAlert(with: AppStrings.Common.noNetworkConnection) : self.showErrorAlert(with: error.localizedDescription)
```

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -306,7 +306,12 @@ final class HTTPClient: Client {
...
-				completion(.failure(.invalidResponseError))
+				switch error {
+				case .noNetworkConnection:
+					completion(.failure(.noNetworkConnection))
+				default:
+					completion(.failure(.invalidResponseError))
+				}
 			}
 		})
 	}
```
pascal-brause, line 317: It would be great if you can add a new unit test to the test file of the authorize call in which you test the new error case

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
pascal-brause: Nice!

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ErrorReporting/ErrorReportCoordinator.swift
```diff
@@ -167,7 +167,12 @@ final class ErrorReportsCoordinator: ErrorReportsCoordinating, RequiresAppDepend
 						self.topViewControllerViewModel?.updateViewModel(isHistorySectionIncluded: true)
 					case .failure(let error):
 						Log.error("ELS submission error: \(error)", log: .els, error: error)
-						self.showErrorAlert(with: error)
+						switch error {
+						case .otpError(let otpError):
+							otpError == .noNetworkConnection ? self.showErrorAlert(with: AppStrings.Common.noNetworkConnection) : self.showErrorAlert(with: error.localizedDescription)
```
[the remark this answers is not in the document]
30mar: for the first point: the function is geting called twice one by passing an Error and the other by passing an OTPError So im passing the error strings directly, if i pass the OTPError we will need to type cast it again into OTPError since the function parameter is of Type Error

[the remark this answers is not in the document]
30mar: regarding the error strings we are showing human readable message for no network errors, so it didn't make sense to add strings to all the enums if we only need one case or what do you think?

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AuthorizationOTPElsTests.swift
```diff
@@ -406,4 +406,50 @@ final class HTTPClientAuthorizationOTPElsTests: CWATestCase {
...
+			userInfo: nil
+		)
+
+		let session = MockUrlSession(
+			data: nil,
+			nextResponse: HTTPURLResponse(
+				url: url,
+				statusCode: 500,
+				httpVersion: nil,
+				headerFields: nil
```
whiskey, line 425: should ne nil, if there is no network connection!

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AuthorizationOTPElsTests.swift
```diff
@@ -406,4 +406,50 @@ final class HTTPClientAuthorizationOTPElsTests: CWATestCase {
...
+			case .failure(let otpError):
+				expectedOtpError = otpError
+			}
+			expectation.fulfill()
+		})
+
+		// THEN
+		waitForExpectations(timeout: expectationsTimeout)
+		XCTAssertNotNil(expectedOtpError)
+		XCTAssertEqual(expectedOtpError, .noNetworkConnection)
```
whiskey, line 453: Why no check in place?
```suggestion
```

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AuthorizationOTPElsTests.swift
```diff
@@ -406,4 +406,50 @@ final class HTTPClientAuthorizationOTPElsTests: CWATestCase {
...
+		let ppacToken = PPACToken(apiToken: "APITokenFake", deviceToken: "DeviceTokenFake")
+
+		// WHEN
+		var expectedOtpError: OTPError?
+		HTTPClient.makeWith(mock: stack).authorize(otpEls: otp, ppacToken: ppacToken, completion: { result in
+			switch result {
+			case .success:
+				XCTFail("success should not be called")
+			case .failure(let otpError):
+				expectedOtpError = otpError
```
whiskey, line 445: Why no check in place:
```suggestion
				XCTAssertEqual(otpError, .noNetworkConnection)
```

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AuthorizationOTPElsTests.swift
```diff
@@ -406,4 +406,50 @@ final class HTTPClientAuthorizationOTPElsTests: CWATestCase {
...
+		let stack = MockNetworkStack(
+			mockSession: session
+		)
+
+		let expectation = self.expectation(description: "completion handler is called without an error")
+		let otp = "OTPFake"
+		let ppacToken = PPACToken(apiToken: "APITokenFake", deviceToken: "DeviceTokenFake")
+
+		// WHEN
+		var expectedOtpError: OTPError?
```
whiskey, line 439: Not needed here
```suggestion
```

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/Survey/SurveyURLProvider.swift
```diff
@@ -28,7 +28,7 @@ enum SurveyError: Error {
 
 	init(otpError: OTPError) {
 		switch otpError {
-		case .generalError, .invalidResponseError, .internalServerError, .otherServerError, .apiTokenExpired, .deviceTokenInvalid, .deviceTokenRedeemed, .deviceTokenSyntaxError:
+		case .generalError, .invalidResponseError, .internalServerError, .otherServerError, .apiTokenExpired, .deviceTokenInvalid, .deviceTokenRedeemed, .deviceTokenSyntaxError, .noNetworkConnection:
```
whiskey, line 31: So ppac, statistics, etc. will all get their own 'no network' errors as well?
This is a global error which should be handled on a higher level.

whiskey (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/Survey/SurveyURLProvider.swift
30mar: I just asked UA about it, it make sense to update it through the app

whiskey (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/__tests__/HTTPClient+AuthorizationOTPElsTests.swift
pascal-brause: I originally implemented this and will explain why i choosed the original style: I find it easier to read. When putting the assert check in the completion handler, one can easily oversee this. When it is at the end, one will find it faster because everyone expects it at the end. Also the Given-When-Then pattern is broken because the assert is "inside" the Then-check.
You can let the code now as it is or revert it, your choice.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/Survey/SurveyURLProvider.swift
pascal-brause: We already have the no network error, but almost every call is not handling it 😅.
But with that big refactoring ticket for the HTTPClient, we could face that, too.

ArturFriesen (approved):

PR #2946, opened 2021-06-15
display issues on TAN-Eingabe

freshking:
## Description

Fix layout issues when using a scroll view in the top view controller

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7112

PR #3114, opened 2021-07-07
Add keyboard subscriptions only to screens with textfields

30mar:
## Description
We needed to add keyboard subscription for handling a previous bug 
https://github.com/corona-warn-app/cwa-app-ios/pull/2946/files
The Problem is even screens without textfields and keyboard also call the sink block of the keyboard subscription and this creates the layout glitches.

The proposed solution is to only subscribe to the keyboard notifications if the TopBottomVC contains textfields so I added a flag for that and subscribe to notifications only for screens that have keyboard like: TAN, Antigen Profile, etc

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8285

## Videos

BEFORE:
https://user-images.githubusercontent.com/15270737/124764568-62df6380-df35-11eb-96e5-2fb07984ab5e.MP4

AFTER:
https://user-images.githubusercontent.com/15270737/124764670-7ee30500-df35-11eb-8734-9663ef845090.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Containers/TopBottomContainerViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Containers/TopBottomContainerViewController.swift
@@ -43,12 +43,10 @@ class TopBottomContainerViewController<TopViewController: UIViewController, Bott
 	
 	init(
 		topController: TopViewController,
-		bottomController: BottomViewController,
-		isWithKeyBoardHandling: Bool = false
+		bottomController: BottomViewController
 	) {
 		self.topViewController = topController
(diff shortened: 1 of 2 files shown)


ArturFriesen (approved):

KaiTeuber:
@30mar Did you try to use a .dropFirst inside the sink instead of that bool solution? I wonder if that might be enough here.

30mar:
@KaiTeuber  yes drop first on only the "HideKeyboardNotification" did the trick, i pushed again

freshking (approved):

KaiTeuber (approved):

PR #3131, opened 2021-07-12
scanning recoverycertificate contains empty required fields

freshking:
## Description

Prevent positively validating recovery certificates that contain empty strings in the required values by setting `minLength` as a validation requirement.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8323


src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Ressources/CertificateSchema.json
```diff
@@ -261,7 +261,8 @@
       "type": "object",
       "properties": {
         "tg": {
-          "$ref": "#/$defs/disease-agent-targeted"
+          "$ref": "#/$defs/disease-agent-targeted",
+          "minLength": 1
```
ArturFriesen, line 265: @mlenkeit could you change this in the referenced schema in the tech-spec, so that it will be there after an update. Maybe its also something Android wants to fix.

ArturFriesen (approved):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Ressources/CertificateSchema.json
mlenkeit: Will do...

mlenkeit (changes requested):
Please don't merge this yet, I need to think about whether we really want to do this change or not (i.e. if this is a bug or not). I'll update the review once we've reached a decision.

pascal-brause (approved):

freshking:
Closing PR because this is actually not a bug. Tech spec is correct.

PR #3136, opened 2021-07-12
Correct the Sorting

30mar:
## Description
Incorrect order in Contact Diary for both Locations and Persons for Special characters like: **Ö, Ä, etc**

- Note: We can also modify the sorting inside the database itself but we will need to do a migration, also changing the behavior in the future so i did the sorting in Swift after fetching from the database.

 
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8177

## Screenshots

IMG_0117
IMG_0118


pascal-brause (commented):
I would suggest to do the sorting not in the fetching function but in the updateDiaryDays function, because there we do already a map on the raw data. Then we could fusion this to not have manipulations on the raw data in two different functions.

whiskey (approved):

src/xcode/ENA/ENA/Source/Scenes/ContactDiary/Store/ContactDiaryStore.swift
```diff
@@ -1004,7 +1004,12 @@ class ContactDiaryStore: DiaryStoring, DiaryProviding, SecureSQLStore {
 			return .failure(dbError(from: database))
 		}
 
-		return .success(contactPersons)
+		let arrangedContactPersons = contactPersons.sorted(by: {
+			$0.name.lowercased().folding(options: .diacriticInsensitive, locale: .current)
+				<
+				$1.name.lowercased().folding(options: .diacriticInsensitive, locale: .current)
+		})
```
nickguendling, line 1011:
```suggestion
		let sortedContactPersons = contactPersons.sorted { $0.name.localizedCompare($1.name) == .orderedAscending }
```

I think that's a bit easier to read

nickguendling (changes requested):

KaiTeuber:
Are you sure the sorting works correctly? In the screenshots Orlando is before Österreich - but Ö is equal to Oe, so it should the opposite order right?

ArturFriesen (approved):

nickguendling (approved):

PR #3140, opened 2021-07-13
Feature/#8382 change value set behavior

pascal-brause:
## Description
For the DCC validation the behavior of fetching the value sets is different: We only return something cached for .notModified case. Otherwise we fail the download because we need for the validation always very fresh value sets.
So i added a second function to the protocol which ignores cached value sets and also the "timeout" when to fetch new value sets.
See also here: https://github.com/corona-warn-app/cwa-app-ios/pull/3122#discussion_r667888205

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8382

## Screenshots
x


src/xcode/ENA/ENA/Source/Services/VaccinationValueSets/VaccinationValueSetsProvider.swift
```diff
@@ -49,9 +60,14 @@ class VaccinationValueSetsProvider: VaccinationValueSetsProviding {
...
-					if let cachedValuesSets = self.store.vaccinationCertificateValueDataSets {
-						promise(.success(cachedValuesSets.valueDataSets))
+					
+					if cachedFallback {
+						// return cached if it exists
+						if let cachedValuesSets = self.store.vaccinationCertificateValueDataSets {
+							promise(.success(cachedValuesSets.valueDataSets))
+						} else {
+							promise(.failure(error))
+						}
```
nickguendling, line 70:
```suggestion
					if cachedFallback, let cachedValuesSets = self.store.vaccinationCertificateValueDataSets {
						promise(.success(cachedValuesSets.valueDataSets))
```

freshking (approved):

nickguendling (approved):

PR #3146, opened 2021-07-13
Cell selection doesn't disappear on FAQ card

freshking:
## Description

Manually deselect cell to prevent then retaining the selected state

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8290


pascal-brause (approved):
Can this behavior appear every time we leave the app for an external link? Perhaps we should check that...

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -242,6 +242,7 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 	
 	// swiftlint:disable:next cyclomatic_complexity
 	override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+		tableView.deselectRow(at: indexPath, animated: true)
```
KaiTeuber, line 245: since we use a TableViewController here isn't `clearsSelectionOnViewWillAppear` the better option?

freshking: That does not work in this case

KaiTeuber (approved):

30mar (approved):

PR #3152, opened 2021-07-14
Positive Test Result - swipe down moves the buttons

freshking:
## Description

Refactoring of the footer view mechanic to solve quite a few issues involving these buttons. Code is now a lot simpler and should be less error prone. 

For example there is no more need to set the height the footer manually. This is solved by using autolayout. 

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8302

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8388


pascal-brause (commented):
Because you heavily change the behavior of the FooterView I cannot really say if this will breaking change something or not. 
Please test on small and big devices as much of the app as you can.

whiskey (approved):
Cool!

src/xcode/ENA/ENA/Source/Scenes/FooterViewController/FooterViewController.swift
```diff
@@ -233,17 +204,35 @@ class FooterViewController: UIViewController {
...
 			.store(in: &subscription)
 
 		viewModel.$backgroundColor
 			.receive(on: DispatchQueue.main.ocombine)
 			.assign(to: \.backgroundColor, on: view)
 			.store(in: &subscription)
 	}
+	
+	private func animateHeightChange() {
+		let animator = UIViewPropertyAnimator(duration: 0.35, curve: .easeInOut) { [weak self] in
```
ArturFriesen, line 228: I think UIView.animate would be enough here.

ArturFriesen (approved):
🚀

src/xcode/ENA/ENA/Source/Scenes/FooterViewController/FooterViewController.swift
freshking: I just used the same animation code that was used before this change.

pascal-brause (approved):

PR #3164, opened 2021-07-15
Feature/8409 fix ui tests for local statistics

whiskey:
## Description
This PR fixes some issues with accessibility elements in local statistics and the matching UI test.
It also fixes the case that we didn't use the `MockStore` for UI tests.

## Link to Jira
[EXPOSUREAPP-8409](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8409)

## Screenshots
n/a

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/HomeStatisticsTableViewCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/HomeStatisticsTableViewCell.swift
@@ -9,6 +9,17 @@ class HomeStatisticsTableViewCell: UITableViewCell {
 
 	// MARK: - Overrides
 
+	override var accessibilityElements: [Any]? {
+		get {
+			stackView.arrangedSubviews
+		}
+
+		// swiftlint:disable:next unused_setter_value
+		set {
+			preconditionFailure("This property is computed by the contents of the stack view")
+		}
+	}
+
     override func awakeFromNib() {
         super.awakeFromNib()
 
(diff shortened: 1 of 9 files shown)


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/HomeStatisticsTableViewCell.swift
```diff
@@ -9,6 +9,17 @@ class HomeStatisticsTableViewCell: UITableViewCell {
...
 	// MARK: - Overrides
 
+	override var accessibilityElements: [Any]? {
+		get {
+			stackView.arrangedSubviews
+		}
+
+		// swiftlint:disable:next unused_setter_value
+		set {
+			preconditionFailure("This property is computed by the contents of the stack view")
```
30mar, line 19: Maybe just LOG the error? wont this crash the app ?

whiskey: This is a developer error because of the overall idea to pile all elements onto one big heap. The user should never see this because DEVs should handle this properly, i.e. not `set` that value.

30mar: No i mean just logging it not showing an alert this way we can debug later but don't block the functionality of the app

30mar (approved):

freshking (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/HomeStatisticsTableViewCell.swift
whiskey: Devs are lazy. If it's not exploding into their face, they won't touch it.

PR #3194, opened 2021-07-19
Feature/8430 invalid signature scanner errror

KaiTeuber:
## Description
Added an alert to scanner if signature wasn't correct

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8430

## Screenshots
Screen Shot 2021-07-19 at 11 46 05


src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -133,6 +136,10 @@ class HealthCertificateService {
 		Log.info("[HealthCertificateService] Registering health certificate from payload: \(private: base45)", log: .api)
 
 		do {
+			if case .failure = signatureVerifying.verify(certificate: base45, with: [], and: Date()) {
+				return .failure(.invalidSignature)
+			}
```
nickguendling, line 141: I think the signature check is too early here. According to the tech spec we should 1. Check and remove prefix, 2. Decode, 3. Verify signature, 4. Check JSON schema, 5. Check content, ...

I think 1, 2 and 4 are actually all contained in the `try HealthCertificate(base45: base45)`, so maybe check the signature right after that?

nickguendling (changes requested):

nickguendling (approved):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -2877,6 +2877,12 @@ Die Angaben werden statistisch ausgewertet, sie werden nicht zu einem Profil ges
 
 "HealthCertificate_Error_HC_QR_CODE_ERROR" = "Der QR-Code konnte nicht generiert werden. Sollte der Fehler weiterhin bestehen, kontaktieren Sie die Hotline über App-Informationen -> Technische Hotline.";
 
+"HealthCertificate_Error_invalidSignature_title" = "Signatur ungültig";
+
+"HealthCertificate_Error_invalidSignature_msg" = "Die Signatur des Zertifikates ist ungültig. Das Zertifikat kann nicht als Nachweis verwendet werden und daher nicht importiert werden.";
```
Ein-Tim, line 2882: In other places we write "Zertifikats", it should be clarified what we should use "Zertifikates" with an "e" or "Zertifikats" without an "e".

@NataliaLemmerth

ChristinaTsigganos: Hi @Ein-Tim , thanks very much for that. Indeed, we had discussed this several times, we will go ahead and haarmonize the use of "Zertifikats" and "Zertifikates".

30mar (approved):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
KaiTeuber: @ChristinaTsigganos will you update the text used here?

ChristinaTsigganos, line 2882: We decided to go for "Zertifikats" throughout the app. Strings well be updated soon. Please take "Zertifikats" over for this PR.

KaiTeuber: @ChristinaTsigganos I updated the text as discussed - please have a look.

ChristinaTsigganos (approved):
UA approved

Issue #2955, opened 2021-06-16
White screen - app not starting


vwdss:
## Technical details

- Device name: iPhone 11 Pro Max
- iOS version: iOS 14.6
- App version: 2.3.3

## Describe the bug

1. Cannot click 'Activate Exposure Logging' without going back.
2. Cannot enable exposure logging
3. App shows white screen after first use

## Steps to reproduce the issue

https://user-images.githubusercontent.com/3357905/122245931-6e0b1a80-cec6-11eb-8542-ab40e02d6711.mp4

--- 

Internal Tracking-ID: [EXPOSUREAPP-7862](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7862)

Ein-Tim:
Hey @vwdss 

#2951 is the fix for this, it has been merged into `release/2.4.x`, so this should be fixed in the next version. 

Please leave this issue open until you can confirm that the fix in version 2.4 (which is still in development) is working. 

--- 

Internal Tracking-ID: [EXPOSUREAPP-7862](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7862)

Ein-Tim:
@dsarkar 

I assume the `fix 2.0` label was added here by mistake. Please remove it and add the `fix 2.4` label here. 

Thank you!

thomasaugsten:
@vwdss can you please retest after manually update on v2.4?

vwdss:
@thomasaugsten unfortunately it did not help. I removed the app and reinstalled - also did not help.

heinezen:
@vwdss Thanks for checking. We'll take another look

----
Corona-Warn-App Open Source Team

Ein-Tim:
PR fo 2.5: #3037

dsarkar:
@vwdss have you tried this?

- DE https://www.coronawarn.app/de/faq/#ios_black_screen
- EN https://www.coronawarn.app/en/faq/#ios_black_screen

---
Corona-Warn-App Open Source Team

Ein-Tim:
@dsarkar 

I don't think this is related here since the white or black screen back then was shown in the iOS settings and not in CWA. But give it a try @vwdss.

dsarkar:
@Ein-Tim OK, your are probably right, I might have mixed that up here. Thanks for pointing it out!

vwdss:
> @vwdss have you tried this?
> 
[...]

PR #3258, opened 2021-07-27
Fix/#7862 black white screen

pascal-brause:
## Description
This is another try to fix the bug with the white screen. Our assumption is now that the ENF is not responding and we wait for ever for some answer.
To fix that i added a timer who checks if the ENF was activated after 3 seconds. After this time, we proceed to the Home or throw an alert to indicate that something is wrong. The fix could be to restart the phone, so for this i added a new text. @UA, please review.

If the ENF is stuck and we proceed, we will enter the the App with the status of ENF being not activated (see screenshot below). 

Only side effect I noticed is when scanning a positive test result, you cannot share this. At the point to tap the "share with others" button, an alert will occure that the ENF is not running.

To test the ENF failure behavior, comment out in `ExposureManger` the lines 229-236 and lines 280-287 in this branch.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7862

## Screenshots
App is stuck with white screen:
7862-AppStuckAtStart

App enters Home after 5 seconds of waiting and is in this status:
7862-HomeWithENFDisabled

When trying to activate the ENF, we show an alert:
7862-AlertWhenTryingToActivate


pascal-brause:
refers to #2955

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -220,14 +221,37 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
+				timer.invalidate()
+				completion(nil)
+			} else if retries >= timeout {
+				timer.invalidate()
+				Log.error("Could not activate ENF within \(retries) seconds. Proceed with deactivated ENF")
+				completion(ExposureNotificationError.notResponding)
+			} else {
+				Log.warning("Could not activate ENF within \(retries) seconds. Retry in 1 second.")
+				retries += 1
+			}
```
nickguendling, line 254: A solution to not check just once every second could be to make `activated` a publisher and use [timeout](https://developer.apple.com/documentation/combine/fail/timeout(_:scheduler:options:customerror:)) to wait x seconds max for it to publish. That way the execution would continue immediately.

Or an even simpler solution could be to just keep the old `completion(nil)` where it was and just use the timer to check once after x seconds if it was called and if not, call `completion(ExposureNotificationError.notResponding)`

pascal-brause: Good idea (last one), then i can also use GDC for the timer.

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -246,17 +259,29 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 				Log.error("Failed to change ENManager.setExposureNotificationEnabled to \(status): \(error.localizedDescription)", log: .api)
 				self.handleENError(error: error, completion: completion)
 				return
 			}
+			Log.error("Successfully changed ENManager.setExposureNotificationEnabled to \(status).", log: .api)
+			changed = true
 			completion(nil)
 		}
+		
+		// Sometimes the ENF is broken. So we wait 1 seconds to ensure the changed applied. Mostly, the ENF responding instantly, so we check after 2 seconds and show then an alert.
```
naveeddotio, line 276:
```suggestion
		// Sometimes the ENF framework is broken. So we wait 1 seconds to ensure that the changed has been applied. Mostly, the ENF framework responds instantly, so we check after 2 seconds and show then an alert.
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -246,17 +259,29 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	}
 
 	private func changeEnabled(to status: Bool, completion: @escaping CompletionHandler) {
+		Log.info("Trying to change ENManager.setExposureNotificationEnabled to \(status).")
+
+		var changed = false
```
naveeddotio, line 264:
```suggestion
		var isChanged = false
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -246,17 +259,29 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 		manager.setExposureNotificationEnabled(status) { error in
 			if let error = error {
 				Log.error("Failed to change ENManager.setExposureNotificationEnabled to \(status): \(error.localizedDescription)", log: .api)
 				self.handleENError(error: error, completion: completion)
 				return
 			}
+			Log.error("Successfully changed ENManager.setExposureNotificationEnabled to \(status).", log: .api)
+			changed = true
```
naveeddotio, line 272:
```suggestion
			isChanged = true
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -220,14 +221,26 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	/// Needs to be called before `ExposureManager.enable()`
 	func activate(completion: @escaping CompletionHandler) {
 		Log.info("Trying to activate ENManager.")
+		
+		var activated = false
```
naveeddotio, line 225:
```suggestion
		var isActivated = false
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -220,14 +221,26 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 		manager.activate { activationError in
 			if let activationError = activationError {
 				Log.error("Failed to activate ENManager: \(activationError.localizedDescription)", log: .api)
 				self.handleENError(error: activationError, completion: completion)
 				return
 			}
-			Log.info("Activated ENManager succesfully.")
+			Log.info("Activated ENF successfully.")
 			completion(nil)
+			activated = true
```
naveeddotio, line 235:
```suggestion
			isActivated = true
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -220,14 +221,26 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	/// Needs to be called before `ExposureManager.enable()`
 	func activate(completion: @escaping CompletionHandler) {
 		Log.info("Trying to activate ENManager.")
```
naveeddotio, line 223:
```suggestion
		Log.info("Trying to activate ENF")
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -220,14 +221,26 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 			}
-			Log.info("Activated ENManager succesfully.")
+			Log.info("Activated ENF successfully.")
 			completion(nil)
+			activated = true
+		}
+
+		// Sometimes the ENF is broken. So we check after 3 seconds if it was activated until we proceed with a deactivated ENF and log an error. Mostly, the ENF is activated instantly, so 3 seconds should be enough time to wait.
+		DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+			if !activated {
```
naveeddotio, line 240:
```suggestion
			if !isActivated {
```

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -246,17 +259,29 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 				return
 			}
+			Log.error("Successfully changed ENManager.setExposureNotificationEnabled to \(status).", log: .api)
+			changed = true
 			completion(nil)
 		}
+		
+		// Sometimes the ENF is broken. So we wait 1 seconds to ensure the changed applied. Mostly, the ENF responding instantly, so we check after 2 seconds and show then an alert.
+		DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+			if !changed {
```
naveeddotio, line 278:
```suggestion
			if !isChanged {
```

naveeddotio (approved):
Small suggestions

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: Hmm what about hasChanged?

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: See above

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: Proposal: isActive?

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: See above

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: See above

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: See above

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
naveeddotio: even better

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
naveeddotio: even better

jucktnich:
Are you sure you want wait 5 secs without any feedback of the app. If heard that people often restart the app from 3 secs on. Maybe show a little waiting graphic.

pascal-brause:
> Are you sure you want to wait 5 secs without any feedback from the app. I've heard that people often restart the app from 3 secs on. Maybe show a little waiting graphic.

Thanks for your feedback. First, I decreased the time we wait to 3 seconds. Second, we can not show here a loading indicator without refactoring parts of the app start, which will take a while. And because we can not reproduce the real error, every bug fix is just an assumption so we should be minimal invasive at changing the UI for the app.

jucktnich:
Ah yeah sometimes you should check the code and not only the description 🙈. Not showing a graphic even with only 3 secs isn't optimal, but it's really not worth the work

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -219,15 +220,27 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	/// Activates `ENManager`
 	/// Needs to be called before `ExposureManager.enable()`
 	func activate(completion: @escaping CompletionHandler) {
-		Log.info("Trying to activate ENManager.")
```
marcussc, line 222: Hm the old log was more correct, wasn't it?
We activate the ENManager, which will then activate ENF right?

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -219,15 +220,27 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
+		var isActive = false
+		
 		manager.activate { activationError in
 			if let activationError = activationError {
 				Log.error("Failed to activate ENManager: \(activationError.localizedDescription)", log: .api)
 				self.handleENError(error: activationError, completion: completion)
 				return
 			}
-			Log.info("Activated ENManager succesfully.")
+			Log.info("Activated ENF successfully.")
```
marcussc, line 233: Same here

marcussc (approved):

ChristinaTsigganos (changes requested):
please take over the UA proposal for the error message

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: I can revert it.

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: I can revert it.

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -246,17 +259,29 @@ final class ENAExposureManager: NSObject, ExposureManager {
...
 				self.handleENError(error: error, completion: completion)
 				return
 			}
+			Log.error("Successfully changed ENManager.setExposureNotificationEnabled to \(status).", log: .api)
+			hasChanged = true
 			completion(nil)
 		}
+		
+		// Sometimes the ENF framework is broken. So we wait 1 seconds to ensure that the changed has been applied. Mostly, the ENF framework responds instantly, so we check after 2 seconds and show then an alert.
+		DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
KaiTeuber, line 277: The comment say 2 seconds but the code only 1? What is the correct value herer?

KaiTeuber (approved):

pascal-brause:
> please take over the UA proposal for the error message

Done.

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
pascal-brause: the two is outdated. I'll fix that.

ChristinaTsigganos (approved):
UA approved

PR #3269, opened 2021-07-29
Title issue on certificate validation result screen

freshking:
## Description

 Refactor HealthCertificateValidationViewController to use navigation item title 

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8512

## Screenshots

IMG_3242


src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -109,6 +109,16 @@ extension DynamicTableViewController {
 			view?.titleLabel.accessibilityLabel = title
 			view?.titleLabel.accessibilityTraits = .header
 			view?.title = title
+			
+			// add optional background gradient
+			if let backgroundGradient = backgroundGradient, let view = view {
```
ArturFriesen, line 114: I think it would be nicer to have this inside of `DynamicTableViewHeaderImageView`. There could be a property `backgroundView` to avoid searching for subviews.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
freshking: The problem with putting it in `DynamicTableViewHeaderImageView` is that it has a title label that is placed above the the image which we don't need in this case.

ArturFriesen: I would not expect that the label is overlapped by the backgroundGradient if I use it.

freshking: I mean that the title label occupies space above the image which conflicts with the layout need in this case.

pascal-brause (approved):

KaiTeuber:
The scrolling of the title was requested from UX. We should discuss what's the expected solution here. i think the main issue is that the transparent navigation bar did get restored. As the testers are not aware of the UX wish with the scrolling title?

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/Result/HealthCertificateTechnicalValidationFailedViewModel.swift
```diff
@@ -17,15 +17,15 @@ struct HealthCertificateTechnicalValidationFailedViewModel: HealthCertificateVal
 	}
 
 	// MARK: - Internal
+	
+	var title = AppStrings.HealthCertificate.Validation.Result.TechnicalFailed.title
```
KaiTeuber, line 21: UX requested a scrolling title in this screen - so this might break this request

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/Result/HealthCertificateValidationFailedViewModel.swift
```diff
@@ -26,6 +26,8 @@ struct HealthCertificateValidationFailedViewModel: HealthCertificateValidationRe
 
 	// MARK: - Internal
 	
+	var title = AppStrings.HealthCertificate.Validation.Result.Failed.title
```
KaiTeuber, line 29: title not scrollable as requested by UX

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/Result/HealthCertificateValidationOpenViewModel.swift
```diff
@@ -26,6 +26,8 @@ struct HealthCertificateValidationOpenViewModel: HealthCertificateValidationResu
 
 	// MARK: - Internal
 	
+	var title = AppStrings.HealthCertificate.Validation.Result.Open.title
```
KaiTeuber, line 29: same as above

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Validation/Result/HealthCertificateValidationPassedViewModel.swift
```diff
@@ -22,14 +22,13 @@ struct HealthCertificateValidationPassedViewModel: HealthCertificateValidationRe
 
 	// MARK: - Internal
 
+	var title = AppStrings.HealthCertificate.Validation.Result.Passed.title
```
KaiTeuber, line 25: one more

KaiTeuber (changes requested):
UX requested the scrollable titles, so we added them here.
Maybe the testers are not aware of that change?

pascal-brause (changes requested):
As Kai mentioned, change would break UX requirement

PR #3280, opened 2021-07-30
Feauture/8420 dcc signature validation

ArturFriesen:
## Description
Add signature check and corresponding CBOR decoding.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8420


src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/SignatureVerification/Model/DCCSigningCertificate.swift
```diff
@@ -3,8 +3,11 @@
 //
 
 import Foundation
+import Security
+import CryptoKit
```
marcussc, line 7: Does that work with iOS 12.5? 🤔

ArturFriesen: Yes, it is compiling, because the import is not used 🙈 removed it.

marcussc (approved):

30mar (approved):

KaiTeuber (approved):

Issue #2758, opened 2021-05-23
Fehler 13 still operative even after 24 hours


mtrinkau:
EDIT 210523: Corona-Warn-App Open Source Team:

Version 2.2.1 (0)
IPhone 11 Pro 
IOS 14.5.1.

1. Did this occur just after updating CWA from 2.1 to 2.2?
2. Did this issue occur when receiving a test result, and if so positive or negative?
3. Any other occurrence at the same time, i.e. another error notification, i.e. NoExposureWindow?
4. Does the error occur when you open the app, or on its own.
5. When was the last risk check?
6. Did you have the same error before using 2.2, and if yes in which context (updating CWA, QR code scan, getting test result)

Answers: 

1. After the update. Never had this issue before.
2. No. Just wanted to check my status.
3. No
4. Only when I open the app and try to update the status.
5. Day before yesterday at 01:15 am
6. No. I never had this issue before.

---
Internal Tracking ID: [EXPOSUREAPP-5836](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5836)
error report ID ID: 2EC7ECO7FECO3E9B6E3A

Ein-Tim:
Hey @mtrinkau 

Could you please let us know which version of the Corona-Warn-App you're using? 

Thank you!

mtrinkau:
Hi - am using Version 2.2.1 (0)

Best

Ein-Tim:
@mtrinkau 

Okay, thank you for the information.
Could you please also provide us with your iOS version and your iPhone Model? 

Thank you very much and have a nice Sunday!

mtrinkau:
Thank you for the prompt response.
IPhone 11 Pro and IOS 14.5.1. 

Best & a nice Sunday too

Ein-Tim:
Okay, thank you very much @mtrinkau!

We had some more users reporting this issue in version 1.15 as well as version 2.0, but this should have been fixed in version 2.1. 
One of the community managers (cc @dsarkar) will soon come back to you with further instructions or requests.

Thank you for reporting this issue and helping to improve the Corona-Warn-App!

dsarkar:
@mtrinkau Thank you for reporting! 
@Ein-Tim Thanks!

@mtrinkau Could you provide us with an error report:

1. start the error report logging
[...]

PR #3284, opened 2021-07-30
Fix ENError13

Ein-Tim:
## Description 
This PR resolves an issue with missing rate limit checks (ENError 13) for subsequent ENF API calls, after the exposure detection returned with an error code.
See more details at https://github.com/corona-warn-app/cwa-app-ios/issues/2758

The solution introduces an additional persistent timestamp (exposureDetectionDate), which is updated directly before the call to ENManager.detectExposures(), and is then used by the RiskProvider in the rate limit check.
We have tested the solution manually (with a modified MockExposureManager returning error codes) and with a unit test.

## Link to JIRA
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5836

## Credits
All work in this PR was done by @ndegendogo!


dsarkar:
@Ein-Tim @ndegendogo Thank you ! Forwarded to internal ticket and pinged a dev team member.

ArturFriesen (changes requested):
Thanks for your contribution!
The check for having a min. time distance between detection calls is already implemented in `RiskProvider`. The implementation also depends on a timeframe which is configurable over the `AppConfiguration`.

Please see this codeline:
https://github.com/corona-warn-app/cwa-app-ios/blob/426ffaba0194165e6e48a51ffbde4c0ec37b0cda/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift#L333

I propose to solve the problem at this place. I think we should consider to compare against a date, which we set on detection call instead of the calculation date. Because the calculation date is only set if the ENF returns a success and the calculation was successful.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -331,7 +331,7 @@ final class RiskProvider: RiskProviding {
 
 	private func previousRiskIfExistingAndNotExpired(userInitiated: Bool) -> Risk? {
 		let enoughTimeHasPassed = riskProvidingConfiguration.shouldPerformExposureDetection(
-			lastExposureDetectionDate: store.enfRiskCalculationResult?.calculationDate
+			lastExposureDetectionDate: store.exposureDetectionDate
```
ndegendogo, line 334: @ArturFriesen 
> The check for having a min. time distance between detection calls is already implemented in RiskProvider. The implementation also depends on a timeframe which is configurable over the AppConfiguration.
I propose to solve the problem at this place. I think we should consider to compare against a date, which we set on detection call instead of the calculation date. Because the calculation date is only set if the ENF returns a success and the calculation was successful.

Do I understand your proposal correctly that we should not only check but also setup the timestamp here in this function? E.g. [before this return?](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift#L357)
**Pro**: big improvement with regard to locality of the change
**Con**: currently I see a few exit paths after this check but before the call to ENF, which would unnecessarily extend the time for next detection although ENF was not called ([here](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift#L368), [here](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Services/Exposure%20Transaction/ExposureDetection.swift#L77), [here](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Services/Exposure%20Transaction/ExposureDetection.swift#L83), [here](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Services/Exposure%20Transaction/ExposureDetection.swift#L47), and [here](https://github.com/corona-warn-app/cwa-app-ios/blob/34dc7d3c5acea27dd74829ce0d4199f9e38eb032/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift#L304)). As far as I see these execution paths are currently related to reasons like duplicate risk task execution or wrong device time, so maybe we can just ignore that?

thomasaugsten:
The feedback from Apple at the beginning of the project was that errors are not count into the quota.  
An error 13 can also caused by uninstall and reinstall of the app and an update from <v1.9

Adding a rate limit can lead to a malfunction for all 15 Mio clients if there is mistake in the implementation.
I'm not sure if the rare occurrence of Error 13 justifies such significant change of the exposure check. 

We could implement a soft rate limit which only logs the rate limit to see in case of an error 13 the rate limit would prevent this error.

Ein-Tim:
@thomasaugsten 

> The feedback from Apple at the beginning of the project was that errors are not count into the quota.

Could you maybe check if this has changed in the meantime? Because IIRC it's possible to get the Error 13 by hard closing the app while an Exposure Check is running. The Exposure Check will fail and as soon as you open the app again it will run again. If you now close the app again while it's running the check will fail again, and so on. After max. 6 times of doing this you will get Error 13, although the Exposure Checks weren't successful.

thomasaugsten:
I don't think they touch this in the last year. This means the exposure check didn't fail there is only no app to receive the result

ndegendogo:
@thomasaugsten we have (at least) [two reported occurrences of error 13 with cwa 2.5.1](https://github.com/corona-warn-app/cwa-app-ios/issues/2758), so it still happens, although not very often.
I am not aware of any changes in the implementation of the risk checking procedure since cwa 2.5.

> The feedback from Apple at the beginning of the project was that errors are not count into the quota.

Is this still the same strategy with ENF v2? 
Apple specifies [3 different numbers for their rate limit](https://developer.apple.com/documentation/exposurenotification/enmanager/3586331-detectexposures) for different iOS and ENF versions.
Who knows if they have maybe also changed their strategy of counting?

> An error 13 can also caused by uninstall and reinstall of the app and an update from <v1.9

Well - in my case it was obviously a different reason. I did neither reinstall cwa, nor did I any update of the app. Also I did not hard-close it. I saw the message at 08:33 when I opened the app in the morning; the previous check at 06:35 was successful, and should have prevented another call of the API. But the check at 02:35 before was missing - so maybe some kind of error code from ENF at that time? We can only guess, I didn't have logging active at that moment...

> Adding a rate limit can lead to a malfunction for all 15 Mio clients if there is mistake in the implementation.

***I fully agree that cwa must be as robust as possible with regard to the exposure detection and risk check procedure. 
This is the core functionality of the app, and there is no other app available in Germany that could replace it.***

> I'm not sure if the rare occurrence of Error 13 justifies such significant change of the exposure check.

So your suggestion is to leave it alone and untouched, in the assumption that even if the error 13 occurs, it is no permanent failure, and will hopefully recover itself, right?
Note: In my case it indeed recovered itself after I left it alone for a few hours; and I have the impression that the background procedure for the check is meanwhile indeed really very robust. 

*No idea if it would have recovered itself if I had kept the app open and busy and had triggered the error 13 repeatedly - but next time I will test that, for sure :-)*

> We could implement a soft rate limit which only logs the rate limit to see in case of an error 13 the rate limit would prevent this error.

Good idea. 
Do you plan to implement such a "soft rate limit"? 
Or should I modify this PR into that direction?

thomasaugsten:
There are no changes on ENF from Apple side.
It is not reproducible and is self healing after 24h.
Yes please modify your PR

ndegendogo:
@ArturFriesen I am still busy with this, but I have a question now.
As a first step I have extended your Logging system for "mockability", so that I am able to write unit tests for my implementation of the "soft rate limit" (no blocking of exposure detection, only log entries).

**Question:** Shall I create a separate PR for that, or do you prefer that I just include it in this PR together with the "soft rate limit"?

ArturFriesen:
> @ArturFriesen I am still busy with this, but I have a question now.
> As a first step I have extended your Logging system for "mockability", so that I am able to write unit tests for my implementation of the "soft rate limit" (no blocking of exposure detection, only log entries).
> 
> **Question:** Shall I create a separate PR for that, or do you prefer that I just include it in this PR together with the "soft rate limit"?

@ndegendogo I think it's a good idea to split it.

Ein-Tim:
As this PR won't get merged, I'll close it now.

PR #3293, opened 2021-08-02
Feature/8693 update determination of most relevant certificate

30mar:
## Description
Update determination of most relevant certificate

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8693


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificateArray+MostRelevant.swift
```diff
@@ -117,31 +131,28 @@ extension Array where Element == HealthCertificate {
...
 
-		if let outdatedAntigenTestCertificate = sortedHealthCertificates.last(where: { $0.testEntry?.coronaTestType == .antigen }) {
+		if let outdatedAntigenTestCertificate = last(where: { $0.testEntry?.coronaTestType == .antigen }) {
 			return outdatedAntigenTestCertificate
 		}
 
-		// Fallback
-
-		return first
+		return nil
```
marcussc, line 156: fallback to nothing, is this really a good idea?

30mar: The fall back is moved to the calling function, this way we go to the else if block, other wise the function would return the first element which is incorrect unless all the cases fail
Screenshot 2021-08-02 at 14 14 33

pascal-brause: then perhaps leave a comment here. The next one is questioning the same ;)

marcussc (approved):

pascal-brause (approved):

naveeddotio (approved):

PR #3295, opened 2021-08-02
Positive Test Result - swipe down moves the buttons

freshking:
## Description

Refactor `FooterViewController` to `UIView` instead of `UIViewController`. 

When presented modally, the `TopBottomContainerViewController` changes the safearea insets of the child view controller when attempting to dismiss the parent my swiping down. This results in weird UI behavior in the bottom view controller. 

To prevent this I made the `FooterViewController` a `UIView` instead of `UIViewController`. Which makes more sense anyway since we don't need all the cruft that is associated with a controller here.

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8302


pascal-brause (commented):
Seems for me a really breaking change. Perhaps it would be better (if not too late) to split up the refactoring from ViewController to View in a separate PR and afterwards the usage to fix the bug in another PR.
So we eventually get a hole new bugs only for the refactoring.

src/xcode/ENA/ENA/Source/Scenes/ErrorReporting/MainErrorReportViewControllers/BottomErrorReportView.swift
```diff
@@ -0,0 +1,217 @@
+////
+// 🦠 Corona-Warn-App
+//
+
+import UIKit
+import OpenCombine
+
+class BottomErrorReportView: UIView {
```
KaiTeuber, line 8: I wonder what this has in common with the problem of the keys? This is about a bug in the coordinator, but not responsible for the scrolling problem?
That seems to be a different bug.

KaiTeuber (changes requested):
In general, I do not understand why it is necessary to change the concept here. The changes are very large and could have more impact than hoped. Has this been sufficiently tested in all situations?

src/xcode/ENA/ENA/Source/Scenes/FooterView/FooterView.swift
```diff
@@ -37,25 +37,63 @@ class FooterViewController: UIViewController {
...
-	// MARK: - Overrides
+	// MARK: - Internal
+
+	var viewModel: FooterViewModel {
+		didSet {
+			updateViewModel()
+		}
+	}
+
+	// MARK: - Private
```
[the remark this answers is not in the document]
freshking: This is not my code

src/xcode/ENA/ENA/Source/Scenes/ErrorReporting/MainErrorReportViewControllers/BottomErrorReportView.swift
freshking: > Seems for me a really breaking change. Perhaps it would be better (if not too late) to split up the refactoring from ViewController to View in a separate PR and afterwards the usage to fix the bug in another PR.
> So we eventually get a hole new bugs only for the refactoring.

Basically all I did is Change a child view controller to a view. I think the changes are rather minimal.

freshking:
> Seems for me a really breaking change. Perhaps it would be better (if not too late) to split up the refactoring from ViewController to View in a separate PR and afterwards the usage to fix the bug in another PR.
> So we eventually get a hole new bugs only for the refactoring.

Basically all I did is change a child view controller to a view. I think the changes are rather minimal.

src/xcode/ENA/ENA/Source/Scenes/ErrorReporting/MainErrorReportViewControllers/BottomErrorReportView.swift
freshking: I just changes this class from a view controller to a view. Nothing else

freshking:
> In general, I do not understand why it is necessary to change the concept here. The changes are very large and could have more impact than hoped. Has this been sufficiently tested in all situations?

Basically all I did is change a child view controller to a view. I think the changes are rather minimal.

src/xcode/ENA/ENA/Source/Scenes/FooterView/FooterView.swift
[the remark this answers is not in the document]
KaiTeuber: But you changed the logic from the ViewController (who owns the viewModel) to a view. So the change isn't perfect right here?

marcussc:
As discussed in the daily, closing this PR for now.

PR #3300, opened 2021-08-03
"Certificates" becomes slower with each certificate (in one personal wallet)

freshking:
## Description

Prevent unnecessary array and class reconstructions for better scroll performance

## Link to Jira

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8567


src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/HealthCertificate/HealthCertificateCellViewModel.swift
```diff
@@ -105,9 +108,9 @@ final class HealthCertificateCellViewModel {
 		} else {
 			return nil
 		}
-	}
+	}()
 
-	var image: UIImage {
+	lazy var image: UIImage = {
```
nickguendling, line 113: I think if we don't reevaluate the properties every time they are not updating anymore if e.g. the `validityState` of the certificate or the `vaccinationState` of the person changes while the user is on the screen. We probably need publishers then to update the views individually if a reload doesn't trigger those updates anymore.

nickguendling (changes requested):

Ein-Tim:
What's the status of this PR?

PR #3305, opened 2021-08-03
Fix/6908 prevent loading UI on bg task

ArturFriesen:
## Description
The solution for the bug and the main change of this PR is to postpone `setupUI(...)` from `didFinishLaunchingWithOptions` to `didBecomeActive`.

`didFinishLaunchingWithOptions` is called when the app state changes from `not running` / `suspended` to `background`.
`setupUI(...)` leads to some `viewDidLoad` calls, which lead to updating the test results (and also other calls).

In case of a background task execution there was a race condition between the update starting from `viewDidLoad` and the update starting from the background task. Due to this race condition the alert was not triggered sometimes.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6908


src/xcode/ENA/ENA/Resources/Info_Debug.plist
```diff
@@ -56,6 +56,8 @@
 	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
+	<key>UIMainStoryboardFile</key>
```
marcussc, line 59: Do we need this change in the other Info_*.plist as well?

marcussc (approved):
Nice finding!

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -118,8 +120,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		}
 
 		// Check for any URLs passed into the app – most likely via scanning a QR code from event or antigen rapid test
-		let route = routeFromLaunchOptions(launchOptions)
-		setupUI(route)
+		// Route will be executed in 'applicationDidBecomeActive'
+		self.route = routeFromLaunchOptions(launchOptions)
```
KaiTeuber, line 124:
```suggestion
		route = routeFromLaunchOptions(launchOptions)
```

src/xcode/ENA/ENA/Resources/Info_Debug.plist
ArturFriesen: Good point!

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -156,7 +160,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 	}
 
 	func applicationDidBecomeActive(_ application: UIApplication) {
-		Log.info("Application did become active.", log: .background)
+		Log.info("Application did become active.", log: .appLifecycle)
+
+		if !didSetupUI {
+			setupUI(route)
+			didSetupUI = true
```
KaiTeuber, line 167: i think you should also reset `route` after handling it ti avoid it gets handled twice

KaiTeuber (approved):
Did you check if it handles the `postOnboardingRoute` correctly? I guess it will, but I'm not sure about that.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
ArturFriesen: Good point. Done.

src/xcode/ENA/ENA/Resources/Info_Debug.plist
ArturFriesen: done

30mar (approved):

PR #3306, opened 2021-08-04
Feature/8378 Implement exceptions from schema validation

30mar:
## Description

- Implement Exceptions from schema validation for certificates as proposed by the EU
- Date of birth schema validations are removed now any dob can be accepted in the certificate according to the tech specs
- Remove white spaces and nil value from the CBOR certificate if the validation is success

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8378


src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Utils/CBOR+AnyMap.swift
```diff
@@ -41,12 +41,15 @@ extension CBOR {
...
         case .utf8String(let stringValue):
-            return stringValue as Any
+            return stringValue.trimmingCharacters(in: .whitespacesAndNewlines) as Any
         case .array(let arrayValue):
             return arrayValue.map { $0.anyValue }
         case .map(let mapValue):
             return mapValue.anyMap as Any
-        case .null, .undefined, .tagged, .break:
+        case .null:
+            // if the value is null we will remove both the key and value from the dictionary
```
ArturFriesen, line 50: I don't understand this change. 
Why is this done?

A null value is a valid value.
Removing the key alters the structure of the object, in the way that it is not the same object anymore.

ArturFriesen (changes requested):
- Please see comment.
- The tests of HealthCertificateToolkit are not green anymore

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Ressources/CertificateSchema.json
```diff
@@ -38,7 +38,7 @@
     },
     "v": {
       "description": "Vaccination Group",
-      "type": ["array", null],
+      "type": ["array", "null"],
```
ArturFriesen, line 41: Does that work now?
I needed to change "null" to null to make the schema validation work.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Utils/CBOR+AnyMap.swift
30mar: The tech specs were updated, now we need to remove null values from the DGC CBOR:
Screenshot 2021-08-04 at 10 01 05

30mar: the DGC from the previous preset should be valid now if we try to scan the qr code

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Ressources/CertificateSchema.json
30mar: I only tested with the certificate generated from here as per the tech spec, **cwa tc gen --preset 25**
the validation works but the certificate has an invalid signature, so it should work, i will retest it

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Utils/CBOR+AnyMap.swift
30mar: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/5d0c104b512eb601d22a1926e204222d8feddf6c/docs/spec/dgc-overview-client.md#decoding-a-dgc-from-cose

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/__test__/HealthCertificateTests.swift
```diff
@@ -86,14 +86,10 @@ class HealthCertificateTests: XCTestCase {
...
-		XCTAssertNotNil(error)
-
-		guard case let .HC_JSON_SCHEMA_INVALID(validationError) = error,
-			  case .VALIDATION_RESULT_FAILED = validationError else {
-			XCTFail("Unexpected error returned")
-			return
-		}
+		// According to the tech specs now any value can be accepted as a date
+		//https://github.com/corona-warn-app/cwa-app-tech-spec/blob/5d0c104b512eb601d22a1926e204222d8feddf6c/docs/spec/dgc-overview-client.md#sample-data-for-formatting-the-date-of-birth
+		XCTAssertNotNil(healthCertificate)
```
ArturFriesen, line 91: The test is testing as `Base45WrongDGCEncoded`. So instead of adjusting the assert, I think something else should be used to create an wrong DGC, instead of birthdate.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/DigitalCovidCertificateAccess.swift
```diff
@@ -227,11 +227,23 @@ public struct DigitalCovidCertificateAccess: DigitalCovidCertificateAccessProtoc
         guard  let healthCertificateCBOR = healthCertificateMap[1] else {
             return .failure(.HC_CBORWEBTOKEN_NO_DIGITALGREENCERTIFICATE)
         }
-
+        
+        guard let trimmedHealthCertificateCBOR = cborMapWithTrimming(certificateCBOR: healthCertificateCBOR) else {
+            return .failure(.HC_JSON_SCHEMA_INVALID(.DECODING_FAILED))
```
ArturFriesen, line 232: The error is wrong, not the decoding failed at this place.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Utils/CBOR+CBORMap.swift
```diff
@@ -0,0 +1,41 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import SwiftCBOR
+
+extension Dictionary where Key == CBOR, Value == CBOR {
+    var cborMapWithTrimmingWhiteSpaces: [CBOR: CBOR] {
```
ArturFriesen, line 8:
```suggestion
    var trimmed: [CBOR: CBOR] {
```

ArturFriesen, line 8: Maybe add some comments what is trimmed.

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/Utils/CBOR+CBORMap.swift
```diff
@@ -0,0 +1,41 @@
...
+        var anyMap = [CBOR: CBOR]()
+        for (key, value) in self {
+            anyMap[key] = value.cborValue
+        }
+        return anyMap
+    }
+}
+
+extension CBOR {
```
ArturFriesen, line 17: private or fileprivate?

src/xcode/ENA/HealthCertificateToolkit/Tests/HealthCertificateToolkitTests/CertificateAccess/DigitalCovidCertificateAccessTests.swift
```diff
@@ -178,6 +178,30 @@ final class DigitalCovidCertificateAccessTests: XCTestCase {
             XCTAssertEqual(keyIdentifier, expectedKeyIdentifier)
         }
     }
+    
+    func test_When_CBORHasNullValues_Then_SchemaValidationSucceeds() {
```
ArturFriesen, line 182: Can we have tests for the nil values are really stripped?
There could be the case, that validation succeeds, result is a success, but still the nil values are not trimmed.

ArturFriesen (changes requested):
Please see comments.

ArturFriesen (approved):

marcussc (approved):

PR #3310, opened 2021-08-04
Feature/8396 encrypt event check in/8819 update proto buf add new request

KaiTeuber:
## Description
* updated protobuf generated file
* added new requests for encrypted downloading (uploading not needed)

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8819


marcussc (approved):

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient+Configuration.swift
```diff
@@ -39,6 +39,7 @@ extension HTTPClient {
 
 			return Configuration(
 				apiVersion: "v1",
+				encryptedApiVersion: "v2",
```
ArturFriesen, line 42: I think if we refactor the http stuff, we should take into account that not every endpoint will operate with the same version. Like we already see with the encrypted API.

ArturFriesen (approved):

PR #3315, opened 2021-08-04
Feature/8396 encrypt event check in/8812 extend app configuration

KaiTeuber:
## Description
App Configuration will get an optional feature flag to disable encryption for event checkins.
Extracted an extension to handle all feature flags equally. This required some refactpring of the device time check feature flag.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8818

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
@@ -21,14 +21,6 @@ final class DeviceTimeCheck: DeviceTimeCheckProtocol {
 
 	// MARK: - Protocol DeviceTimeCheckProtocol
 
-	// MARK: - Internal
-
-	enum TimeCheckResult: Int, Codable {
-		case correct
-		case assumedCorrect
-		case incorrect
-	}
-
 	var isDeviceTimeCorrect: Bool {
 		return store.deviceTimeCheckResult == .correct || store.deviceTimeCheckResult == .assumedCorrect
 	}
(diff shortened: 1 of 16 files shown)


src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -172,6 +171,12 @@ final class MockTestStore: Store, PPAnalyticsData {
 	// MARK: - Protocol DSCListCaching
 
 	var dscList: DSCListMetaData?
+
+	// MARK: - Protocol AppFeaturesStoring
+	#if !RELEASE
```
nickguendling, line 176: I think it's not useful to have a `#if !RELEASE` condition inside of the `#if DEBUG` scope, or is it?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
```diff
@@ -15,20 +15,12 @@ final class DeviceTimeCheck: DeviceTimeCheckProtocol {
 
 	// MARK: - Init
 
-	init(store: AppConfigCaching & DeviceTimeChecking) {
+	init(store: AppConfigCaching & DeviceTimeChecking & AppFeaturesStoring) {
 		self.store = store
 	}
 
 	// MARK: - Protocol DeviceTimeCheckProtocol
```
nickguendling, line 22:
```suggestion
	// MARK: - Protocol DeviceTimeChecking
```

Seems to have been renamed later on, let's just clean it up while we're here 🧹

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/DeviceTimeCheck.swift
KaiTeuber: done

nickguendling (approved):

marcussc (approved):

PR #3324, opened 2021-08-05
Feature/8396 encrypt event check in/8815 select correct downloading

KaiTeuber:
## Description
all downloads of TraceWarnings will now react on the app feature .unencryptedCheckinsEnabled
App Feature now can be decorated to make different behaviors much easier to realize.
This made it possible to add it to the debug menu very easy. Some more refactoring to the DeviceTime check was needed.
A bunch of Unittests have be updated to use the decorated app features - all ran green locally. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8815


src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -157,6 +158,7 @@ final class RiskProvider: RiskProviding {
 
 	private let store: Store
 	private let appConfigurationProvider: AppConfigurationProviding
+	private let appFeatureProvider: AppFeatureProviding
```
marcussc, line 161: This is only used for the DeviceTimeCheck right? 
Why not store the DeviceTimeCheck directly during init?

marcussc (approved):
Small thing that good be improved otherwise LGTM

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
KaiTeuber: Good point, I'll try it.

KaiTeuber: I did some more refactoring to get rid of a dependency cycle, please have another look

ArturFriesen (approved):

PR #3337, opened 2021-08-09
Fix/8867 federal state name localization

naveeddotio:
## Description
Adding localization to the list of federal state names

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8867

## Screenshots

P.S. the placeholders are shown because the translation delivery needs to provide the English names

Simulator Screen Shot - iPhone 12 - 2021-08-09 at 15 56 18
Simulator Screen Shot - iPhone 12 - 2021-08-09 at 15 56 28
Simulator Screen Shot - iPhone 12 - 2021-08-09 at 15 56 34


NataliaLemmerth (approved):
UA approved

nickguendling (changes requested):
I tried it out, I can't add local statistics anymore if my App is set to English, and for the data donation the federal state gets lost if I select it in English and then switch the app language to German, might not be saved properly at all.

nickguendling (approved):

PR #3340, opened 2021-08-09
Fix/8404 disappearing certificates

nickguendling:
## Description
While we were not able to reproduce the issue of disappearing health certificates ourselves so far, these changes attempt to minimize the probability of certificates disappearing.

Main Changes:
- If CBOR decoding of one of the certificates fails, all other certificates do not disappear with it anymore.
- The certificates are only written back to the store if there are actual changes. If they stay unchanged, the store content is kept and decoding certificates that previously failed might work again.

Additional Changes:
- The validity state is now persisted as well.
- Errors that are passed to the `log()` method are actually logged now (should make debugging easier in the future).

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8404


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertifiedPerson.swift
```diff
@@ -30,7 +30,7 @@ class HealthCertifiedPerson: Codable, Equatable, Comparable {
 	required init(from decoder: Decoder) throws {
 		let container = try decoder.container(keyedBy: CodingKeys.self)
 
-		healthCertificates = try container.decode([HealthCertificate].self, forKey: .healthCertificates)
+		healthCertificates = try container.decode([HealthCertificateDecodingContainer].self, forKey: .healthCertificates).compactMap { $0.healthCertificate }
```
ArturFriesen, line 33: Is this backwards compatible? Is it possible to encode a type of `[HealthCertificate]` and then decode a type of `[HealthCertificateDecodingContainer]`

nickguendling: Yes, because what we're storing is basically just `[{base45: "HC1:xxx"}, {base45: "HC1:yyy"}]`, so as long as we keep the same property names (or rather implicit `CodingKeys` in this case) it decodes just fine.

ArturFriesen: Ok, I See. And because `validityState` is optional it will just be nil. 👍

ArturFriesen (approved):

marcussc (approved):

PR #3349, opened 2021-08-11
Fix/8919 failing background exposure checks

nickguendling:
## Description
Activates the exposure manager in the background task if it wasn't activated yet. Needed because previously it was only activated in `AppDelegate.showHome()` which is not called on the background task anymore due to the changes in #3305 

(Contains an unrelated typo-fix that additionally moves the code that is not part of the protocol to the private section)

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8919


marcussc (approved):
Nice finding, also nice Description of the PR!

ArturFriesen (approved):

ArturFriesen (approved):

PR #3373, opened 2021-08-17
Feature/8691 show complete vaccination for recovered people immediately after first vaccination

KaiTeuber:
## Description
For vaccines which require usually two doses (Astra, Biontech, Moderna), the vaccination shall be shown as completed immediately, if the dose is 1/1
For Johnson&Johnson, which has always dose 1/1, the 14 days rule still applies

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8691

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/VaccinationProductType.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/VaccinationProductType.swift
@@ -1,0 +1,27 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+enum VaccinationProductType {
+	case biontech
+	case moderna
+	case astraZeneca
+	case unknown
+
+	// MARK: - Init
+
+	init(value: String) {
+		switch value.uppercased() {
+		case "EU/1/20/1528":
+			self = .biontech
+		case "EU/1/20/1507":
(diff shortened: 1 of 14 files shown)


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/VaccinationProductType.swift
```diff
@@ -0,0 +1,27 @@
...
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+enum VaccinationProductType {
+	case biontech
+	case moderna
+	case astraZeneca
+	case unknown
```
marcussc, line 11: not sure if compiler allows this but:
```suggestion
	case other
```
I think this would make it easier to understand the guard let in `HealthCertifiedPerson.swift`: `VaccinationProductType(value: vaccinationEntry.vaccineMedicinalProduct) != .unknown else`

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/VaccinationProductType.swift
KaiTeuber: Done

marcussc (approved):

PR #3387, opened 2021-08-20
Feature/8356 warning on behalf/8980 request

nickguendling:
## Description
Introduces the `OnBehalfCheckinSubmissionService` and implements the actual submission of a checkin on behalf of an attendee that doesn't use the Corona-Warn-App.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8980


NataliaLemmerth (approved):
UA approved

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Client/__tests__/Mocks/ClientMock.swift
```diff
@@ -53,6 +53,7 @@ final class ClientMock {
 
 	var onGetTestResult: ((String, Bool, TestResultHandler) -> Void)?
 	var onSubmitCountries: ((_ payload: SubmissionPayload, _ isFake: Bool, _ completion: @escaping KeySubmissionResponse) -> Void) = { $2(.success(())) }
+	var onSubmitOnBehalf: ((_ payload: SubmissionPayload, _ isFake: Bool, _ completion: @escaping KeySubmissionResponse) -> Void) = { $2(.success(())) }
```
naveeddotio, line 56: sounds a bit strange :D

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
```diff
@@ -152,21 +152,19 @@ class ENAExposureSubmissionService: ExposureSubmissionService {
 					with: self.symptomsOnset
 				)
 
-				let unencryptedCheckinsEnabled = self.appConfigurationProvider.featureProvider.value(for: .unencryptedCheckinsEnabled)
+				let unencryptedCheckinsEnabled = AppFeatureProvider(appConfig: appConfig).value(for: .unencryptedCheckinsEnabled)
```
naveeddotio, line 155: 👍

ArturFriesen: This change would not take into account the settings from the dev-menu. Please see the decorater pattern implemented to add the dev-menu functionality. Maybe we should change the pattern. Because its not obvious.

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Submission/ExposureSubmissionService.swift
nickguendling: Ah, thanks for that hint! I've now reversed that change and instead made sure that `self.appConfigurationProvider.featureProvider` is always up-to-date whenever the app config publisher comes back. Maybe it makes sense to get rid of the `AppFeatureProvider(appConfig: appConfig)` again as the publisher is now more reliable?

Issue #2156, opened 2021-03-05
Clicking on icon on the tab bar should bring to the top


Ein-Tim:
## Avoid duplicates
* [X] This issue has not already been raised before
* [ ] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation
Clicking on for example the "Start Screen" icon on the tab bar does nothing if the start screen is already opened. Even if the user is on the very bottom of the start screen, nothing happens.

## Suggested Enhancement
Scroll to the top of the screen.
Native iOS Apps (try i.e. the App Store or Photos) are scrolling back to the top if the user is clicking on the icon from the currently opened menu on the tab bar.

## Expected Benefits
Consistency with other and mainly the native iOS Apps.

---

## Related Issue

- #2159 

---
Internal Tracking ID: [EXPOSUREAPP-5572](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5572)

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-5572](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5572)

Ein-Tim:
Just to follow up here:

Although I didn't consider this as a bug, some users might think it's a bug. I really hope that one of the devs is on this, since iOS wide I didn't found **a single** application which doesn't support this.
Is this a story in JIRA @dsarkar?

dsarkar:
@Ein-Tim It is indeed, Internal Tracking ID: EXPOSUREAPP-5572 as a wishlist item. Considered to be a great addition, but right now, not in development. Currently prio high on wishlist (be aware, this might change.).

tens0rfl0w:
@Ein-Tim @dsarkar Hey there! I checked the mentioned behavior on multiple views and can't reproduce it right now, seems fixed👍🏼

fynngodau:
In Android, touching the Start Screen icon reloads the Start Screen, thereby jumping (not smooth scrolling) to the top. I can't tell to what extend this is intentional, but this is indeed the behavior I'd expect from a tab bar on Android.

Ein-Tim:
@fynngodau 

Touching the icon on a tab bar should scroll to the top of the view. Additionally the view can be reloaded. 
This is the behavior of every single iOS app on my phone, native and third party ones. 
CWA is really an exception there.

ndegendogo:
[...]

PR #3411, opened 2021-08-24
Scroll to top when clicking on current tab (Implements #2156) (COMMUNITY)

Ein-Tim:
## Description
This PR implements that the view scrolls to the top when the user clicks on the currently opened tab.

## Link to Jira
* Github: #2156
  * Jira: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5572

## Screen recording

https://user-images.githubusercontent.com/67682506/130691614-1705569b-b85d-4f00-b6c0-ff268f22b551.mp4

## Credits
All work in this PR was done by the amazing @ndegendogo! Thank you so much! 🎉


src/xcode/ENA/ENA/Source/RootCoordinator.swift
```diff
@@ -245,3 +246,51 @@ extension RootCoordinator: ENStateHandlerUpdating {
...
+}
+
+private extension Array {
+	func first<T>(ofType _: T.Type) -> T? {
+		first(where: { $0 is T }) as? T
+	}
+
+	func last<T>(ofType _: T.Type) -> T? {
+		last(where: { $0 is T }) as? T
+	}
```
KaiTeuber, line 295: I can't see that `last` gets used - if not we should remove it.

ndegendogo: @KaiTeuber 
Indeed.
Another thought. Actually, I copy-pasted this from AppNavigationController.
Maybe I should move the new code over to AppNavigationController? Then I don't need to duplicate it?
In that case AppNavigationController would take responsibility as delegate for the tab bar?

ndegendogo: @KaiTeuber I moved now the solution into AppNavigationController:
- AppNavigationController takes now responsibility as delegate for the tab bar
- Array extensions already available here, no need to duplicate them

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/RootCoordinator.swift
```diff
@@ -144,6 +144,7 @@ class RootCoordinator: RequiresAppDependencies {
 		diaryCoordinator.viewController.tabBarItem = diaryTabBarItem
 
 		tabBarController.tabBar.tintColor = .enaColor(for: .tint)
+		tabBarController.delegate = homeCoordinator.rootViewController
```
ArturFriesen, line 147: Does the feature only work for the home screen?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
```diff
@@ -34,7 +34,7 @@ class HomeCoordinator: RequiresAppDependencies {
 
 	// MARK: - Internal
 
-	let rootViewController: UINavigationController = AppNavigationController(rootViewController: UIViewController())
+	let rootViewController: AppNavigationController = AppNavigationController(rootViewController: UIViewController())
```
ArturFriesen, line 37: This change brings other functionality and configuration to the rooViewController, which can have side effects we don't see yet.
Would it be possible to separate the the "scroll up" functionality and use it without the use of AppNavigationController?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/RootCoordinator.swift
ndegendogo: @ArturFriesen actually, it is implemented for all 4 tabs. The scrolling is intended only on the root view controller of the respective tab; this follows the behaviour of native apps like the Health app. 
It works only if that controller has a table view; this was tested for both a native UITableViewController and a cwa DynamicTableViewController.
I have tested it on simulator; not tested for iOS 12.5.
I have tested on all 4 tabs; however, the Certificate tab I have tested only with its initial info screen (I don't know how to add certificates with the simulator - is there a trick for this?).

src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
ndegendogo: @ArturFriesen Good point.
Actually, I wanted to participate in the private extensions for Array; but I can duplicate this code elsewhere.

ArturFriesen: Instead of implementing the functionality in the `AppNavigationController`, I would rather move the extension out and make it reusable. A good thing would be: separate the responsibilities and keep the scope of `AppNavigationController` small.

src/xcode/ENA/ENA/Source/RootCoordinator.swift
Ein-Tim: Please re-review @ArturFriesen. Thanks!

src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
Ein-Tim: Please re-review @ArturFriesen. Thanks!

src/xcode/ENA/ENA/Source/RootCoordinator.swift
ArturFriesen: This line was confusing. `homeCoordinator.rootViewController` is the root controller of the home tab. And it was handling the scrolling of the other tabs. From hierarchy perspective this is confusing. Only by looking at this line, it looks like only the root of home screen will be handled.

However, I like the solution with the `TabBarScrolling` a lot more! It is not confusing from hierarchy perspective and encapsulates the behavior in an own component. Thanks!

ArturFriesen (approved):
Thanks a lot for your efforts! I really appreciate it!

Issue #2743, opened 2021-05-20
[Exposure logging active screen] headline changes appearance


jucktnich:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: iPhone SE 1
- iOS version: iOS 14.5.1
- App version: 2.2.1 (0)

## Describe the bug

The headline changes appearance
## Steps to reproduce the issue

Watch the video:
![Image](https://user-images.githubusercontent.com/73582096/118995613-949d6a80-b987-11eb-933b-4241b8ac6815.MOV)

## Expected behaviour

No changes
---
Internal Tracking ID: [EXPOSUREAPP-7385](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7385)

dsarkar:
@jucktnich Thanks. Internal Tracking ID: [EXPOSUREAPP-7385](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7385)

Cannot reproduce on iPhone 6s / iOS 14.4 / CWA 2.2.1


---
Corona-Warn-App Open Source Team

Ein-Tim:
Tested on:

- iPhone SE 1. Gen (Emulator)
- iPhone SE 2. Gen (Emulator)
- iPhone 6s (real device)
- iPhone 12 Pro (Emulator)

Reproducible on all these devices.

Issue #3372, opened 2021-08-17
UI: Headline disappears


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone XR
- iOS version: 14.7.1
- App version: 2.7.1

## Describe the bug

When swiping back from the "My QR Codes" screen, the headline disappears: 

https://user-images.githubusercontent.com/67682506/129716156-14b20983-b9e7-4c51-ac15-b90754d8e427.mov

## Steps to reproduce the issue

1. Open the Corona-Warn-App
2. Scroll down to the "Planning an event?" box
3. Click on "Create QR Code"
4. Swipe back as shown in the video

## Expected behaviour

The headline should not disappear

---
Internal Tracking ID: [EXPOSUREAPP-9029](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9029)

dsarkar:
@Ein-Tim Thanks. Internal Tracking ID: [EXPOSUREAPP-9029](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9029)

ndegendogo:
UIKit provides the following elements for handling of large titles:
- the navigation bar can be configured to support large titles (UINavigationBar.prefersLargeTitles). 
Note that the **navigation bar is shared** between the sibling view controllers (these siblings are pushed in from the side).
- each of these view controller **has its own navigation item** which it configures with its title and layout preferences (UINavigationItem.largeTitleDisplayMode).
- in the current implementation the view controllers permanently re-configure the shared navigation bar to their own preferences in their viewWillAppear method.

*A better solution seems to setup the navigation bar only from its owner (the navigation controller), and the view controllers don't fiddle with it and only setup their navigation item. 
BUT: it seems that the UIKit API has a bug (or hard-to-understand usage) if this view controller is a UITableViewController 
(if you google you find tons of Voodoo recipes and all seem to help in specific circumstances only)...*

I continue to investigate this...

dsarkar:
@ndegendogo forwarded all to internal ticket. Thanks.

ndegendogo:
[...]

PR #3430, opened 2021-08-26
Fix/7385 #2743 large titles change appearance

ndegendogo:
## Description

This PR fixes an issue with mixed normal and large titles in sibling view controllers under the same navigation controller.
When the user starts to swipe a view controller with large title from the stack, but then cancels the action, the large title property is already lost.
The issue was reported in github as #2743, see more details at that ticket.

Note: additionally I found an inconsistency in the implementation of the DataDonation screen - it was seemingly implemented with a large title, but shown only with normal title. 
Not sure which behaviour was intended - I changed it to "large title", see screenshot.

## Link to Jira

Internal Tracking ID: EXPOSUREAPP-7385

## Screenshots

Simulator Screen Shot - iPhone 8 - 2021-08-26 at 15 57 22


src/xcode/ENA/ENA/Source/Scenes/DataDonation/DataDonationViewController.swift
```diff
@@ -26,12 +26,17 @@ class DataDonationViewController: DynamicTableViewController, DeltaOnboardingVie
 
 	override func viewDidLoad() {
 		super.viewDidLoad()
-		parent?.navigationItem.title = AppStrings.DataDonation.Info.title
-		parent?.navigationItem.largeTitleDisplayMode = .always
-		parent?.navigationController?.navigationBar.prefersLargeTitles = true
+		navigationItem.title = AppStrings.DataDonation.Info.title
```
ArturFriesen, line 29: Would it be possible to have the 3 lines of configuration just in `viewWillAppear`?

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
```diff
@@ -38,7 +38,7 @@ final class ExposureNotificationSettingViewController: UITableViewController, Ac
 		navigationItem.title = AppStrings.ExposureNotificationSetting.title
```
ArturFriesen, line 38: Same question here: Would it be possible to have the 3 lines of configuration just in viewWillAppear?

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/InviteFriendsViewController.swift
```diff
@@ -45,6 +45,11 @@ final class InviteFriendsViewController: UIViewController, UIActivityItemSource
 		inviteButton.setTitle(AppStrings.InviteFriends.submit, for: .normal)
 	}
 
+	override func viewWillAppear(_ animated: Bool) {
```
ArturFriesen, line 48: Same question here. See above.

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettings/NotificationSettingsViewController.swift
```diff
@@ -46,6 +46,11 @@ class NotificationSettingsViewController: UIViewController, UITableViewDataSourc
 		notificationSettings()
 	}
 
+	override func viewWillAppear(_ animated: Bool) {
```
ArturFriesen, line 49: Same question here. See above.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsCoordinator.swift
```diff
@@ -135,7 +135,6 @@ class SettingsCoordinator: ENStateHandlerUpdating {
 		let dataDonationViewController = DataDonationViewController(viewModel: viewModel)
 		dataDonationViewController.navigationItem.title = AppStrings.DataDonation.Info.title
 		dataDonationViewController.navigationItem.largeTitleDisplayMode = .always
-		dataDonationViewController.navigationController?.navigationBar.prefersLargeTitles = true
```
ArturFriesen, line 138: What does this deletion change?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/DataDonation/DataDonationViewController.swift
ndegendogo: **navigation item:** all example code from apple I have seen so far perform the setup of navigation items in viewDidLoad, so I consider it as the "standard way" of implementation. But of course you can do it differently if you have reasons.
viewDidLoad is called once / at the beginning of the lifetime of the view controller. The navigation item is owned by the view controller, so there is no reason to configure it repeatedly in each 'viewWillAppear' call; a one-time configuration should be sufficient.

**navigation bar:** this is shared between the sibling view controllers. And to be honest, I am not too enthusiastic that this shared property is reconfigured by the view controllers all the time; I am currently investigating if I find a better solution (see #3372), but this investigation is still ongoing.

Note that the timing of 'viewDidLoad' and 'viewWillAppear' is slightly different, so there is a potential risk that moving these lines could have an impact on the animation. 
I tried to move all three lines into 'DataDonationViewController.viewWillAppear', and I saw no adverse effect; but I tested on simulator only, not on a true device.

ndegendogo: ***I removed an inconsistency here and changed the behaviour from normal to large titles.
Please clarify with your UI responsibles if these large titles are ok.***
*(one side effect of the large titles is that the consent box is no longer visible without scrolling - and this might be undesired...)*

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
ndegendogo: General remarks on navigation items and navigation bar: see above.

Strangely, here it is not possible to move the lines into viewWillAppear, it has an impact on the animation, and would initially be shown with normal title.

I cannot tell for sure which fact causes this difference in behaviour, although I suspect that is because ExposureNotificationSettingViewController is inherited from UITableViewController.

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettings/NotificationSettingsViewController.swift
ndegendogo: See answer on DataDonationViewController.
NotificationSettingsViewController does not inherit from UITableViewController, and I saw no adverse effect when I moved these lines.

src/xcode/ENA/ENA/Source/Scenes/FriendsInvite/InviteFriendsViewController.swift
ndegendogo: See answer on DataDonationViewController.
InviteFriendsViewController does not inherit from UITableViewController, and I saw no adverse effect when I moved these lines.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsCoordinator.swift
ndegendogo: Actually, this line had no effect, because the dataDonationViewController has no navigation controller before it is pushed on the stack. I removed the line mainly because it is / was confusing and misleading.

By the way, the other two lines above are duplicated in DataDonationViewController.viewDidLoad(). Currently this duplication is aligned, but could easily run out of synch in future. For sake of maintainability, I suggest to remove those lines from one or the other location (or even move them into DataDonationViewController.viewWillAppear if you prefer that).

ndegendogo:
@ArturFriesen I have answered all your questions above.
Please clarify with your UI responsible if the large titles are ok for the DataDonation screen, or if I should revert the behaviour to normal titles.

src/xcode/ENA/ENA/Source/Scenes/DataDonation/DataDonationViewController.swift
ArturFriesen: @ndegendogo: Regarding separation of configuration in`viewDidLoad ` and `viewWillAppear`: keep it like that. Because only the configuration of the `navigationBar` can be reconfigured during a navigation to the next view controller. So it makes sense to only reconfigure `navigationBar` on reappearance.

Regarding the large titles: It would be very nice if you provide me screenshots of the different states. This way I can directly ask a designer about the change.

src/xcode/ENA/ENA/Source/Scenes/Settings/SettingsCoordinator.swift
ArturFriesen: The duplication was also something I stumbled over. I would also prefer to remove it at one place.

ndegendogo: > The duplication was also something I stumbled over. I would also prefer to remove it at one place.

Done.

src/xcode/ENA/ENA/Source/Scenes/DataDonation/DataDonationViewController.swift
ndegendogo: First screenshot shows data donation screen as of current implementation (normal title).
Second screenshot shows same screen but with large title.
Both screenshot with iPhone SE (shorter display size).

Simulator Screen Shot - iPhone SE (2nd generation) - 2021-09-01 at 10 54 22

Simulator Screen Shot - iPhone SE (2nd generation) - 2021-09-01 at 10 41 32

ArturFriesen: @ndegendogo I had a chat with a responsible designer. Its a little more complicated. This view controller is used in the onboarding and also in the settings. The ideal solution would be: a small title in the onboarding and a large title in the settings.

ndegendogo: lol 😅😅

I'll try ...

ndegendogo: @ArturFriesen unfortunately, its a little more complicated .....

In the current branch release/2.10.x all onboarding screens have a large free space at the top - except the Data Donation, which has a large title. And ... yes, the Data Donation screen from Settings has a normal title.

*I hope your designer won't cry if he sees this ...*

I have submitted a ticket #3459 with screenshots.

My suggestion is now as follows:
1) confine the scope of this PR. I cannot save the world in one day. I'll role back all the changes related to the data donation screen, so at least I have no regression here.
2) fix the bug in the onboarding screens (separate PR). Maybe you guys have already seen this and started to work on the fix. If not, I could volunteer, or maybe one of you is faster.
3) still another PR: make your designer happy, as discussed above

ndegendogo: @ArturFriesen I have now reverted all changes releated to the Data Donation screen as suggested.

Additionally I found that this issue occurs also in the 'Background App Refresh' screen (from Settings).
I have fixed also this occurence.

ArturFriesen (approved):

pascal-brause (approved):

PR #3442, opened 2021-08-30
Fix/6354 - Too much free space in "Transnational Exposure Logging" screen

### LINKED ISSUE #2414: Unnecessary free space on top of "Transnational Exposure Logging" screen
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 6s / iPhone XR
- iOS version: 14.5 Public Beta 7
- App version: 1.15.2

## Describe the bug

There's an unnecessary free space above of the illustration in the "Transnational Exposure Logging" screen. This looks very strange in dark mode, because the dark color from the illustration isn't the same as the dark color of the background.

1. Open the Corona-Warn-App
2. Click on "Exposure Logging Active/Stopped" 
3. Click on "Transnational Exposure Logging"
4. See the free space above of the illustration

## Expected behaviour

No free space.

## Screenshots

<details>
<summary>Screenshots</summary>

iPhone 6s:

Light Mode:

Dark Mode:

iPhone XR:

Light Mode:

Dark Mode:

</details>

---
Internal Tracking ID: [EXPOSUREAPP-6354](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6354)

ndegendogo:
Fix #2414 - Unnecessary free space on top of "Transnational Exposure Logging" screen

EUSettingViewController has no title, and needs no largeTitleDisplayMode.

## Description

This PR fixes a layout issue in the "Transnational Exposure Logging" screen.
This screen has no title, but had configured the navigation item with largeTitleDisplayMode, which resulted in a wide navigation bar and unnecessary free space above the banner image. 
Note: Additionally I have moved the configuration of the navigation item into the viewDidLoad method; so the view controller will now use its own navigation item and no longer that of its sibling view controller.
Note: I have tested it on simulator with iPhone 8 and iPhone SE (normal and small screen).
Note: I left the code for iOS 12.5 untouched as I am not able to test this.

The issue was reported as github ticket #2414; see more details at that ticket.

## Link to Jira

Internal Tracking ID: EXPOSUREAPP-6354
github ticket #2414

## Screenshots

Simulator Screen Shot - iPhone 8 - 2021-08-30 at 23 52 47

Simulator Screen Shot - iPhone 8 - 2021-08-30 at 23 53 07


dsarkar:
@ndegendogo Many thanks. PR forwarded to internal ticket and pinged a few developers.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/ENSetting/EUSettings/EUSettingsViewController.swift
```diff
@@ -33,6 +33,9 @@ class EUSettingsViewController: DynamicTableViewController {
 
 	override func viewDidLoad() {
 		super.viewDidLoad()
+		if #available(iOS 13, *) {
```
marcussc, line 36: is the `if #available(iOS 13, *)` really needed?

ndegendogo: @marcussc thanks for your review, and especially for mentioning this one.

Yes, I think this line of iOS 12 code https://github.com/corona-warn-app/cwa-app-ios/blob/d79316e5804952e0734f73036d97dc55e8516166/src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift#L125 can also just be moved over to here.

The main reason that I was reluctant to move it in the first place: **I don't know how to test this.** And I know that you have several pieces of strange-looking code in your project that are actually work-arounds and have their reasons ...

Actually, I am wondering how you guys are testing for iOS 12, and very interested to learn more about this :-) . 
I have seen check-in comments in your code like "fix the build for iOS 12.4" - is this configuration available somewhere? Or can it be enabled manually / locally somehow?

TLDR: I am happy to clean this up.

marcussc: iOS 12 is a little tricky and by tricky I mean impossible to test in the Simulator, since Apple never provided a Simulator for 12.5. On our dev side we either had an old iPhone with 12.x laying around that we could upgrade to 12.5 or even had to buy some used devices (: So if we want to see the code running we have to test on a real device. 

Regarding your issue the compiler should throw an error if you use code that is not available in iOS 12.5. That is why we have all those 
`if #available(iOS 13, *) {
 	//Do stuff that is only available on iOS 13 and higher
 } else {
 	// Fallback to iOS 12 APIs
 }`
scattered all over the code. 

So if a am not mistaken it would be enough to just write:
`navigationItem.largeTitleDisplayMode = .never` without the `if #available(iOS 13, *) {`

ndegendogo: > iOS 12 is a little tricky and by tricky I mean impossible to test in the Simulator, since Apple never provided a Simulator for 12.5. 

Yes, I am aware of that. It's a pity ...
The delta between iOS 12.4 and 12.5 is mainly the Exposure Notification Framework, right? So in theory you could have a configuration for iOS 12.4 with mocks for all layers below the visible user interface (or at least for the ENF), which would allow at least to test the screens and the animations without a true device ... and actually, I was hoping that you have already prepared something similar ... (and I have no idea how much effort it would be to prepare such a configuration ...)

ndegendogo: > On our dev side we either had an old iPhone with 12.x laying around that we could upgrade to 12.5 or even had to buy some used devices (: So if we want to see the code running we have to test on a real device.

Another thought. Actually, I have a device with iOS 12.5, so in theory I could support the project with testing.
Of course I don't have the entitlements for ENF. But the idea of opening the TestFlight for interested beta testers from the community has been brought up several times (although so far always been rejected ...)
Do you think it would be worth to open again a github ticket on this idea (or to revive the discussion on one of the already available tickets)?

ndegendogo:
@marcussc I have now cleaned up the iOS 12 switch as per our discussion, but blindly, not tested, not even compiled.
*I have to rely on your CI robots now.*

marcussc (approved):

PR #3443, opened 2021-08-31
Feature/9211 download booster notifications rules

30mar:
## Description

- download booster notifications rules
- separate download logic into a separate class
- PS: Unit tests will be done in a separate task

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9211


src/xcode/ENA/ENA/Source/Client/Client.swift
```diff
@@ -238,6 +238,12 @@ protocol Client {
 		ruleType: HealthCertificateValidationRuleType,
 		completion: @escaping DCCRulesCompletionHandler
 	)
+	
+	func getBoosterNotificationRules(
```
naveeddotio, line 242: Please add function signature comments like above

NataliaLemmerth (approved):
UA approved - no text changes.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/BoosterNotificationsService.swift
```diff
@@ -0,0 +1,37 @@
...
+		self.store = store
+		self.client = client
+	}
+	
+	// MARK: - Protocol BoosterNotificationsServiceProviding
+	
+	func downloadBoosterNotifications(
+		completion: @escaping (Result<[Rule], HealthCertificateValidationError>) -> Void
+	) {
+		self.rulesDownloadService = RulesDownloadService(store: self.store, client: self.client)
```
pascal-brause, line 26: the `RulesDownloadService` should not be initialized in the download method. Otherwise we would init every time we trigger a download. 
Better would be to inject it in the init(). Additionally, it would make it easier for testing later because you can inject a mock and also the logic is decoupled.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/HealthCertificateValidationService.swift
```diff
@@ -152,10 +145,13 @@ final class HealthCertificateValidationService: HealthCertificateValidationProvi
 							completion(.failure(.VALUE_SET_CLIENT_ERROR))
 						}
 					}
-					
 				}, receiveValue: { [weak self] valueSets in
+					guard let self = self else { return }
+
 					Log.info("Successfully received value sets. Proceed with downloading acceptance rules...", log: .vaccination)
-					self?.downloadAcceptanceRules(
+					self.rulesDownloadService = RulesDownloadService(store: self.store, client: self.client)
```
pascal-brause, line 152: here again, the rulesDownloadService should not be initialized everytime we value sets changes.

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -271,6 +271,13 @@ final class SecureStore: Store, AntigenTestProfileStoring {
 		set { kvStore["invalidationRulesCache"] = newValue }
 	}
 	
+	// MARK: - Protocol HealthCertificateBoosterNotificationCaching
+	
+	var boosterRulesCache: ValidationRulesCache? {
+		get { kvStore["boosterRulesCache"] as ValidationRulesCache? ?? nil }
+		set { kvStore["boosterRulesCache"] = newValue }
```
pascal-brause, line 278: Could not find in the tech spec something about caching. Did i oversee something? If so, it would explain why you added the 304-http-code to the downloading.

pascal-brause (changes requested):
You copied over a lot of stuff from the ValidationService to a new `DownloadRuleService`. But you copied also the functions where the validation is done (`validateRules()`). Seems for me confusion because the naming of the new service suggests it is only about downloading. And also there is mainly nothing left in the ValidationService. 
So i am wondering, why do we really need this new service?

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -702,6 +702,53 @@ final class HTTPClient: Client {
...
+						completion(.failure(.invalidResponse))
+						return
+					}
+					let etag = response.httpResponse.value(forCaseInsensitiveHeaderField: "ETag")
+					let packageDownloadResponse = PackageDownloadResponse(package: sapPackage, etag: etag)
+					Log.info("Successfully got rules for ruleType: Booster")
+					completion(.success(packageDownloadResponse))
+				case 304:
+					Log.info("Content was not modified - 304.", log: .api)
+					completion(.failure(.notModified))
```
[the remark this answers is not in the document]
30mar: I we have the new Booster rules that have the same structure as the validation rules, I wanted to keep the Validation Service focused on the Validation logic so i created a new service to download all the rules, the original health validation service is now 300 lines , the idea is to separate the downloading from the rest of the logic, but i agree i will move the preparation part back to the service and keep the downloading and caching part in the download service.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/BoosterNotificationsService.swift
30mar: Yes this was just a temporary code, as the actual injection is handled in another task, but i will update it now

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/HealthCertificateValidationService.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import OpenCombine
 import HealthCertificateToolkit
-// Do not import everything, just the datatypes we need to make a clean cut.
```
pascal-brause, line 8: bring back the comment?

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/HealthCertificateValidationService.swift
```diff
@@ -109,6 +110,7 @@ final class HealthCertificateValidationService: HealthCertificateValidationProvi
 	
 	// MARK: - Private
 	
+	private var rulesDownloadService: RulesDownloadServiceProviding?
```
pascal-brause, line 113: Why is the service optional when it it passed in the initalizer?
Why is it private?
Is it set from outside?
Otherwise, you could declare it as let and remove the optionality, so you do not have to think about what error is thrown if the service is nil when trying to trigger a download ;)

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/__tests__/HealthCertificateValidationServiceValidationTests.swift
```diff
@@ -1493,132 +1493,133 @@ class HealthCertificateValidationServiceValidationTests: XCTestCase {
...
-		XCTAssertEqual(mappedSet["covid-19-lab-result"]?.first, tcTrKey)
-		XCTAssertEqual(mappedSet["covid-19-lab-test-manufacturer-and-name"]?.first, tcMaKey)
-		XCTAssertEqual(mappedSet["covid-19-lab-test-type"]?.first, tcTtKey)
-		XCTAssertEqual(mappedSet["disease-agent-targeted"]?.first, tgKey)
-		XCTAssertEqual(mappedSet["sct-vaccines-covid-19"]?.first, vpKey)
-		XCTAssertEqual(mappedSet["vaccines-covid-19-auth-holders"]?.first, maKey)
-		XCTAssertEqual(mappedSet["vaccines-covid-19-names"]?.first, mpKey)
-	}
-	
+	// TODO move this test to the RulesDownloadServiceTests
```
pascal-brause, line 1496: is it done in another task?

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
pascal-brause: Okay we need some caching so my comment can be ignored.

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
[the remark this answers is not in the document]
pascal-brause: Okay we need some caching so my comment can be ignored.

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/HealthCertificateValidationService.swift
30mar: I removed the optional and made it a constant, but i think it should remain private as long as we don't need to access it from outside right ?

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/__tests__/HealthCertificateValidationServiceValidationTests.swift
30mar: yeah the tests are now giving errors and i have another task to create the mocks and update the tests

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/HealthCertificateValidationService.swift
pascal-brause: Yes absolutely, it should stay `private`

src/xcode/ENA/ENA/Source/Services/HealthCertificateValidation/__tests__/HealthCertificateValidationServiceValidationTests.swift
pascal-brause: Then you should remove all the comments because the TODO will fail the CI.
When you start over with the subtask just copy the tests from this PR or an old state or something like this.

pascal-brause (approved):

naveeddotio (approved):

PR #3475, opened 2021-09-06
feature/5572-click tabbar scroll to top shows large titles

ndegendogo:
Improve implementation and add unit tests.

## Description

This PR improves the implementation that the view scrolls to the top when the user clicks on the currently opened tab.
The previous implementation (as was integrated in PR #3411) scrolls only up to the first line of the embedded table view.
The new implementation scrolls further up and shows also the large title.
Besides it is more robust, it will not crash if the embedded table view is empty, and it supports also other scroll views besides table views.

## Link to Jira

Jira: Internal Tracking ID: EXPOSUREAPP-5572
github: #2156


Ein-Tim:
*Service*

## Screenrecording

https://user-images.githubusercontent.com/67682506/132257891-8982ecf8-4bbf-41a7-a10b-5906409883d0.mp4

You can see what has improved when comparing this video with the one in #3411.

*/ Service*

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
```diff
@@ -179,34 +179,28 @@ class TabBarScrolling: NSObject, UITabBarControllerDelegate {
...
+			scrollViewToTop(embeddedScrollView)
 		}
 	}
 
-	private func scrollTableToTop(_ tableView: UITableView) {
-		guard !tableView.isEditing else {
-			return
+	private func scrollViewToTop(_ scrollView: UIScrollView) {
+		scrollView.scrollRectToVisible(CGRect(x: 0, y: 0, width: 1, height: 1), animated: true)	// no large title
+		DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
```
marcussc, line 202: Where does the magic number come from? 🤔

ndegendogo: @marcussc indeed, a magic number. And if you activate slow motion in the simulator it will break ...

Apple's [Core Animation Programming Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreAnimation_guide/AnimatableProperties/AnimatableProperties.html#//apple_ref/doc/uid/TP40004514-CH11-SW1) mentions 0.25 sec as duration for default property-based animations (CABasicAnimation) and default transition-based animations (CATransition); see tables B-2 and B-3. 
**Not sure what this means for a scrolling animation.** And to be on the safe side, I want slightly more than this duration; so I took the magic 0.3, and it seems to work, at least on the simulator. 
Note: I have not tested on a true device, although I am aware that this is strongly recommended.... but the damage in case of wrong timing is not so high.
If the time is too short it will stop at the first table row and not reveal the large title.
If the time is too long the animation will look less smooth.

*Do you know of any system constant or similar property that I could / should use instead?
Or should I add a comment with a short explanation?*

Oh, wait, I wrote a unit test ... if the time is indeed too short the unit test will break or be fragile ... maybe I should add a comment there...

ndegendogo:
@marcussc I have added two comments

src/xcode/ENA/ENA/Source/Scenes/App/__test__/NaviControllerViewScrollingTest.swift
```diff
@@ -0,0 +1,312 @@
...
+
+// Hint: if this test turns out fragile, try to increase the timing in UINavigationController.scrollViewToTop
+
+private protocol Scrolling {
+	func scrollToBottom()
+}
+
+private typealias ScrollingViewController = UIViewController & Scrolling
+
+class NaviControllerViewScrollingTest: CWATestCase {
```
pascal-brause, line 16:
```suggestion
class NavigationControllerViewScrollingTest: CWATestCase {
```

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
pascal-brause: Yeah, add a comment, that would be nice

marcussc: Yeah a comment would be nice our you could implement the `scrollViewShouldScrollToTop(_:)` from `UIScrollViewDelegate` (https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619382-scrollviewdidscrolltotop)

ndegendogo: @marcussc - UIScrollViewDelegate - cool idea 😎👍

ndegendogo: Unfortunately, my first tries on this really cool idea didn't work out as expected ....
I tried scrollViewDidEndScrollingAnimation, but on non-table scrollview it created a huge empty space below the title. For tableview it looked ok, only it did not reveal the large title if the table was already at the top row.

So I reverted to the previous solution (with the magic number) and only try to document this number better.

src/xcode/ENA/ENA/Source/Scenes/App/__test__/NaviControllerViewScrollingTest.swift
ndegendogo: done

ndegendogo:
@marcussc @pascal-brause please see my latest commit.

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
marcussc: Thanks for giving it a try! :)
With the new Comment it looks good to me.

Thank you again for your contribution.

marcussc (approved):

pascal-brause (approved):

PR #3477, opened 2021-09-06
Fix/9268 crash corona test service

naveeddotio:
## Description
A try to solve the possible crash in corona test service. The crash is happening because of unbalanced enter and leaves in the dispatch queue, perhaps due to the non returning closure.  

(i) Capturing weak self
(ii) Adding more logs so if the above approach fails the issue can be traced 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9268


src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -291,7 +291,8 @@ class CoronaTestService {
 
 		for coronaTestType in CoronaTestType.allCases {
 			group.enter()
-
+			Log.info("Dispatch group entered for updateTestResults")
```
pascal-brause, line 294: You could enter in the log the coronaTestType, too.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -301,6 +302,7 @@ class CoronaTestService {
 					break
 				}
 
+				Log.info("Dispatch group exited for updateTestResults")
```
pascal-brause, line 305: You could enter in the log the coronaTestType, too.

pascal-brause (approved):

marcussc (approved):

PR #3496, opened 2021-09-08
Remove misleading about disabled certificate pinning

rugk:
## Description

Fixes https://github.com/corona-warn-app/cwa-app-ios/issues/3495

This removes the ambiguity about when certificate pinning is disabled.
It assures people to show that it is never done automatically on any build variant (which was the claim before) and thus removes the misleading examples.

## Link to Jira

N/A

## Screenshots

N/A

PR #3497, opened 2021-09-08
Remove misleading about disabled certificate pinning (closes #3495)

### LINKED ISSUE #3495: DISABLE_CERTIFICATE_PINNING doc comment mentions TestFlight and other build variants where it does not apply
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: N/A
- iOS version: N/A
- App version: d25bf3875af1e553021780e1eeaced63dc576531

## Describe the bug

Due to some discussion I checked your code around certificate pinning and I noticed a minor doc issue.

Here you have a bypass/conditionally disable certificate pinning via the `DISABLE_CERTIFICATE_PINNING` compilation variable:
https://github.com/corona-warn-app/cwa-app-ios/blob/d25bf3875af1e553021780e1eeaced63dc576531/src/xcode/ENA/ENA/Source/Extensions/URLSession%2BDefault.swift#L12-L15

You say there, among others, it is disabled/used for `TestFlight`, UI testing and some debug/community release versions.

Your [doc](https://github.com/corona-warn-app/cwa-app-ios/blob/d25bf3875af1e553021780e1eeaced63dc576531/COMPILATION_CONDITIONS.md) then says:
> This is **NEVER EVER** set for builds that will end up in the App Store. We only use this for debugging purposes.

Now [TestFlight](https://apps.apple.com/de/app/testflight/id899247664) may be considered to be an official release, even though it may not be a real "app store" release.
Furthermore, you have separate conditions for `COMMUNITY` and `DEBUG` – and yet another `XCUI` env variable for UI testing. I doubt it sets `DISABLE_CERTIFICATE_PINNING` implicitly.

Look into the code.

## Expected behaviour

The documentation/code comments should be coherent.

Also:
* clarification may be needed whether TestFlight versions disable certificate pinning. IMHO they should not do so, and I have no evidence that they would, but your comment suggests otherwise, so better check that.
  If that would be true, this could be considered to contract your "**NEVER EVER**" claim in the doc.
* clarify whether any debug/community/… build respectively build conditions implicitly sets `DISABLE_CERTIFICATE_PINNING` or somehow else disables certificate pinning -> again, I guess, no.

## Possible Fix

As far as I see all your listed examples in that code comment are somewhat misleading, because nothing sets `DISABLE_CERTIFICATE_PINNING` unless one does it explciity.

Solution may be:
* to just remove that comment
* or clarify when `DISABLE_CERTIFICATE_PINNING` is used
* and/or clarify the Markdown doc on when the build condition is used

I have proposed one such possible solution in:
* #3496

## Additional context

Please note I'm no iOS dev nor do i know Swift, so bear with me should I interpret some details wrong.

Also, some dev stuff: Code comments tend to rot, and I guess this was just outdated and may cause some panic for users though (:wink:). So try to be explicit in code and avoiding comments is really a best practice here.

rugk:
## Description

Fixes https://github.com/corona-warn-app/cwa-app-ios/issues/3495

This removes the ambiguity about when certificate pinning is disabled.
It assures people to show that it is never done automatically on any build variant (which was the claim before) and thus removes the misleading examples.

## Link to Jira

N/A

## Screenshots

N/A

---
second try after https://github.com/corona-warn-app/cwa-app-ios/pull/3496 :upside_down_face:


ArturFriesen (approved):

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Extensions/URLSession+Default.swift
```diff
@@ -10,8 +10,7 @@ extension URLSession {
 		delegateQueue: OperationQueue? = nil
 	) -> URLSession {
 		#if DISABLE_CERTIFICATE_PINNING
-		/// Disable certificate pinning while app is running on:
-		/// Community, Debug, TestFlight, UITesting modes
+		/// Disable certificate pinning with build condition for manual debugging if set explicitly.
```
marcussc, line 13: But this is not set manual, it is set in the build settings for the different targets.

ArturFriesen: I see that its set on `Community` and `Debug` configuration but not on `TestFlight`

rugk: Ah okay, thanks. So I've adjusted the comment. :slightly_smiling_face:

marcussc (approved):

PR #3527, opened 2021-09-14
fix/9434-crash-when-export

pascal-brause:
## Description
When traversing between the consent and the generated certificate screens, there may occur in some cases a crash. This happens because we generate the pdf async and push the view controller when we receive the result. This leads to non-UIThread-call and the app crashes.

I also noticed a console assert that we init the PDFView with nothing, we should avoid this.

I also removed after discussion of #3526 the default value of the scale factor in our function to generate the qrImage so that every dev has to think about the property he is passing into.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9434

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/AntigenTestProfile/AntigenTestProfile/Cells/QRCodeCellViewModel.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/AntigenTestProfile/AntigenTestProfile/Cells/QRCodeCellViewModel.swift
@@ -30,7 +30,7 @@ struct QRCodeCellViewModel {
 			with: vCardV4(revDate: revDate),
 			encoding: .utf8,
 			size: CGSize(width: 280.0, height: 280.0),
-			scale: 1,
+			scale: UIScreen.main.scale,
 			qrCodeErrorCorrectionLevel: .medium
 		)
 		else {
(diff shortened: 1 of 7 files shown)


src/xcode/ENA/ENA/Source/Scenes/AntigenTestProfile/AntigenTestProfile/Cells/QRCodeCellViewModel.swift
```diff
@@ -30,6 +30,7 @@ struct QRCodeCellViewModel {
 			with: vCardV4(revDate: revDate),
 			encoding: .utf8,
 			size: CGSize(width: 280.0, height: 280.0),
+			scale: 1,
```
nickguendling, line 33:
```suggestion
			scale: UIScreen.main.scale,
```

As this is rendered for the screen, I think it should be set to screen scale

nickguendling (changes requested):

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/AntigenTestProfile/AntigenTestProfile/Cells/QRCodeCellViewModel.swift
pascal-brause: After testing both implementations, i could not determine any difference. But i adopt your suggest because it makes more sense :)

nickguendling (approved):

PR #3532, opened 2021-09-15
Additional PPA Logging

mlenkeit:
## Description
This PR adds additional logging to some PPA-related functionality to support further analysis of recent observations.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9437
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9450
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9451


src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -392,6 +392,7 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 				switch result {
 				case .success:
 					Log.info("Analytics data succesfully submitted", log: .ppa)
+					Log.info("Analytics submission post-processing self-reference is nil: \(self == nil)", log: .ppa)
```
pascal-brause, line 395: we could add here a guard with a strong self reference, too. So we would be sure, that self is not nil and all the post-processing that depends on self will be executed correctly.

mlenkeit: If you can still get that into 2.10... otherwise, let's stick only with logging. According to @nickguendling, `self` shouldn't be `nil` because you are holding a strong reference elsewhere.
/fyi @marcussc

marcussc: Lets stick with the logging for now, so short before last RC.

marcussc (approved):

pascal-brause (approved):

PR #3553, opened 2021-09-21
Feature/8395 qrscanner/badge notification start tab

naveeddotio:
## Description
Adding the functionality to show a badge on start tab and respective unit tests.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9395

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
@@ -492,6 +492,10 @@ class CoronaTestService {
 		return healthTuple
 	}
 
+	func resetUnseenTestResultCount() {
+		unseenTestResultCount.value = 0
+	}
+
 	// MARK: - Private
 
 	private let client: Client
@@ -510,6 +514,8 @@ class CoronaTestService {
 	private var antigenTestOutdatedDate: Date?
 
 	private var subscriptions = Set<AnyCancellable>()
+
+	private(set) var unseenTestResultCount = CurrentValueSubject<Int, Never>(0)
 
 	private func setup() {
 		updatePublishersFromStore()
@@ -538,6 +544,12 @@ class CoronaTestService {
 				if let antigenTest = antigenTest {
 					self?.setupOutdatedPublisher(for: antigenTest)
 				}
+			}
+			.store(in: &subscriptions)
+		
+		unseenTestResultCount
+			.sink { [weak self] in
+				self?.store.unseenTestResultCount = $0
 			}
 			.store(in: &subscriptions)
 	}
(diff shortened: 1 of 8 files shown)


src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -130,6 +131,8 @@ class CoronaTestService {
 					Log.info("[CoronaTestService] PCR test registered: \(private: String(describing: self?.pcrTest), public: "PCR Test result")", log: .api)
 
 					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), registrationToken, .pcr)))
+					// updating badge count for home tab
+					self?.unseenTestsCount.value += 1
```
nickguendling, line 135: Can we add a parameter `markAsUnseen` to this func so we don't add a badge count if the user adds their test from the home screen? (We have the same thing for health certificates already in place)

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -183,6 +187,9 @@ class CoronaTestService {
 					Analytics.collect(.testResultMetadata(.updateTestResult(.positive, registrationToken, .pcr)))
 					Analytics.collect(.keySubmissionMetadata(.submittedWithTeletan(true, .pcr)))
 					
+					// updating badge count for home tab
+					self?.unseenTestsCount.value += 1
+
```
nickguendling, line 192: There is no way currently to add a teletan from the universal scanner so I think it should not be required to mark it as unseen here.

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -267,6 +274,9 @@ class CoronaTestService {
 
 					Analytics.collect(.testResultMetadata(.registerNewTestMetadata(Date(), registrationToken, .antigen)))
 
+					// updating badge count for home tab
+					self?.unseenTestsCount.value += 1
+
```
nickguendling, line 279: Here I also suggest to add the `markAsUnseen` parameter to the func

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
naveeddotio: Make sense

30mar (approved):

nickguendling (approved):

PR #3590, opened 2021-09-28
Fix/xxxx missing texts

KaiTeuber:
## Description
added missing text placeholders to localizable strings


NataliaLemmerth (approved):
UA approved

naveeddotio (approved):

PR #3591, opened 2021-09-28
Fix broken UI tests

marcussc:
## Description
Since we now have one scanner to rule all QRCodes instead of many, our UI-Tests starting to fail all over the place since each of the old QRCodesScanners had different "hacks" to get the Simulator to scan a QRCode.
When the QRCodeScanner is opened now in the Simulator you don't get an error message but an Alert where you can choose which QRCode you want to fake/scan. 
The failing tests are now using this alert to choose the right QRCode they need to be scanned. 
With 2.12 we would have the chance to remove this "hack" and use the image picker and equip the simulator with the needed QRCodes. This would improve our tested codepaths a lot (@KaiTeuber)

I also removed two performance tests that stole 2 minutes from our CI Runtime without providing any meaningful insights since the results never got tracked.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9627


src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerViewController.swift
```diff
@@ -348,4 +357,36 @@ class QRScannerViewController: UIViewController {
 		previewLayer.mask?.addSublayer(throughHoleLayer)
 		previewLayer.mask?.addSublayer(backdropLayer)
 	}
+
```
pascal-brause, line 360: And for warning on-behalf we do not need an action, correct?

src/xcode/ENA/ENA/Source/Client/Security/__tests__/CryptoFallbackTests.swift
```diff
@@ -23,26 +23,6 @@ class CryptoFallbackTests: iOS13TestCase {
 		XCTAssertEqual(customDigest, referenceDigest)
     }
 
-    func testPerformanceCryptoKitSHA256() throws {
```
ArturFriesen, line 26: 🚮❤️

pascal-brause (commented):
For all occurrences of `#if targetEnvironment(simulator)`: Was it intentionally or why not using `#if DEBUG if isUITesting {`?

marcussc:
> For all occurrences of `#if targetEnvironment(simulator)`: Was it intentionally or why not using `#if DEBUG if isUITesting {`?

@pascal-brause I think `#if targetEnvironment(simulator)` is nicer since this enables us to test all flows that involve a scanned QRCode in the Simulator and not on the device. I also think this makes debugging easier for the Community. But as a mentioned this should get removed once move to 2.12 and we can just use QRCodes saved to the CameraRoll/Files.

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerViewController.swift
marcussc: You need an event for warning on-behalf :)

pascal-brause: What I meant was that we do not need an UIActionSheet action here for scanning an event for the warning on-behalf?

marcussc: There is one in line 382 🕵️‍♂️

pascal-brause: 🤦‍♂️

pascal-brause (approved):

ArturFriesen (approved):

PR #3618, opened 2021-09-30
Feature/#9113 notification rework/#9676 settings screen

pascal-brause:
## Description
The new notification settings screen. 
For this i completely rebuild the screen programmatically with the dynamicTableViewController and it's cells. The logic is now that in the NotificationSettingsViewModel there exists two tableViewModels, one for notifications on and one for notifications off.
 
To achieve all different UX styles, i had to change slightly some cell's and their behavior:
- Added new cell: DoubleLabel. This are two labels in one line. Left one is aligned to left, the right one is aligned to right.
- Icon cell has now a property to set the imageAlignment to left or right.
- Section can now have a rounded background. This background is set accordingly to every cell and in dependence of it's position (top, middle, bottom)
- Rounded cell can now have nil texts and for this case, the spaces have to be removed


src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
```diff
@@ -95,6 +102,34 @@ extension DynamicCell {
 			configure?(viewController, cell, indexPath)
 		}
 	}
+	
+	static func doubleLabels(
+		text1: String,
```
ArturFriesen, line 107: Maybe would be a nice case to use a tuple?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -307,14 +308,40 @@ extension DynamicTableViewController {
...
+		
+		cell.removeBackground()
+		
+		switch section.background {
+		case .none:
+			cell.backgroundColor = .clear
+		case .greyBoxed:
+			
+			// Give all subviews some insets to the border
+			if let subview = cell.subviews.first?.subviews.first {
```
ArturFriesen, line 326: Some magic code here. Which view is getting the inset here? Is it the content view?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -307,14 +308,40 @@ extension DynamicTableViewController {
...
 		let content = dynamicTableViewModel.cell(at: indexPath)
 
 		let cell = tableView.dequeueReusableCell(withIdentifier: content.cellReuseIdentifier, for: indexPath)
+		
+		
+		cell.removeBackground()
+		
+		switch section.background {
+		case .none:
+			cell.backgroundColor = .clear
```
ArturFriesen, line 322: Can this maybe break some styling of existing cells?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -382,3 +409,51 @@ private extension UITableViewCell {
 		}
 	}
 }
+
+/// Stuff to draw a background to the cell, depending on the position of the cell in the section it gets rounded corners.
+
+private extension UITableViewCell {
+	enum BackgroundLocation: Int {
+		case top = 200_001
```
ArturFriesen, line 417: In first sight I was wondering about the magic numbers here. Then I saw, that they have a dependency to other magic numbers (seperators). Would it make sense to place them together at a central point?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -382,3 +409,51 @@ private extension UITableViewCell {
...
+		case bottom = 200_002
+		case inBetween = 200_003
+	}
+	
+	func removeBackground() {
+		viewWithTag(BackgroundLocation.top.rawValue)?.removeFromSuperview()
```
ArturFriesen, line 423: Adding and removing views via tagging makes me a little bit uncomfortable. I know the pattern with the tagged views is copied from the seperators. I'm wondering if there is a better object oriented solution (oh my god, I wrote the forbidden phrase 🙈) Just for discussion.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewDoubleLabelViewCell.swift
```diff
@@ -0,0 +1,123 @@
...
+	override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+		super.init(style: style, reuseIdentifier: reuseIdentifier)
+		setup()
+	}
+
+	override func prepareForReuse() {
+		super.prepareForReuse()
+
+		resetMargins()
+//		configureDynamicType()
```
ArturFriesen, line 28: Can be removed?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewDoubleLabelViewCell.swift
```diff
@@ -0,0 +1,123 @@
...
+			accessibilityIdentifier1: nil,
+			accessibilityIdentifier2: nil,
+			accessibilityTraits1: .none,
+			accessibilityTraits2: .none
+		)
+	}
+	/*
+	// MARK: - Protocol DynamicTableViewTextCell
+	
+	func configureDynamicType(size: CGFloat = 17, weight: UIFont.Weight = .regular, style: UIFont.TextStyle = .body) {
```
ArturFriesen, line 42: Can be removed?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewIconCell.swift
```diff
@@ -33,10 +34,14 @@ class DynamicTableViewIconCell: UITableViewCell {
...
 		contentTextLabel.textColor = .enaColor(for: .textPrimary1)
 		contentTextLabel.numberOfLines = 0
+		contentTextLabel.lineBreakMode = .byWordWrapping
 		contentTextLabel.setContentCompressionResistancePriority(.required, for: .vertical)
 		stackView.addArrangedSubview(contentTextLabel)
 
+		// We need to make the contentTextLabel bigger when we use a bigger font e.g. title1 or title2
+		let size = (contentTextLabel.style?.fontSize ?? width) * 4
 		NSLayoutConstraint.activate([
+			contentTextLabel.heightAnchor.constraint(greaterThanOrEqualToConstant: size),
```
ArturFriesen, line 44: Will this affect other cells besides the settings screen?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewRoundedCell.swift
```diff
@@ -85,16 +85,20 @@ class DynamicTableViewRoundedCell: UITableViewCell {
...
+		let bodyTopConstraintValue: CGFloat = body.attributedText == nil ? 0 : 16
+		
 		NSLayoutConstraint.activate([
 			insetView.leadingAnchor.constraint(equalTo: marginGuide.leadingAnchor),
 			insetView.topAnchor.constraint(equalTo: marginGuide.topAnchor),
 			insetView.trailingAnchor.constraint(equalTo: marginGuide.trailingAnchor),
 			insetView.bottomAnchor.constraint(equalTo: marginGuide.bottomAnchor),
 			title.leadingAnchor.constraint(equalTo: insetView.leadingAnchor, constant: 16),
-			title.topAnchor.constraint(equalTo: insetView.topAnchor, constant: 20),
+			title.topAnchor.constraint(equalTo: insetView.topAnchor, constant: titleTopConstraintValue),
```
ArturFriesen, line 99: Will this affect other cells besides the settings screen?

ArturFriesen (changes requested):
Some changes looks like they would affect other cells as well. Needs to be clarified.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewCell.swift
pascal-brause: After trying around, i'd like to not change it, because for the texts we could use a tuple, but for the accessibilityIdentifer and the accessibilityTraits we cannot because of the default values.
And so we would expect tuples and not-tuples, what would confuse we if i use this element.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
pascal-brause: It's about the views inside the content view. We cannot change the insets of the content view. I added a comment for better understanding.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
pascal-brause: Indeed, i will remove that.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
pascal-brause: Hmm i was thinking about this, too, but decided to put these view tags there. Now we have two extension (separators and background) and they capsulate their view tags and the logic inside. If we put the tags together at one place, one could be confused why they are not aligned with the logic.... Do you agree or disagree?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
pascal-brause: I will think about the forbidden phrase ;)

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewDoubleLabelViewCell.swift
pascal-brause: Yes!

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewDoubleLabelViewCell.swift
pascal-brause: Yes we can :)

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewIconCell.swift
pascal-brause: No, the others screen using the IconCell never pass a nil text and body.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewRoundedCell.swift
pascal-brause: No, the others screen using the IconCell never pass a nil text and body.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
ArturFriesen: Its potentially dangerous. Some future developer could just add a case like `topleft`. And for whatever reasons, think its a good idea to name it "100_001". And then there will be a very weird bug, hard to track down.

pascal-brause: That's a point. I extracted both view tags (background and separators) to a common central place. (see file  UITableViewCell+ViewTags)

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
pascal-brause: Actually, nothing really good comes to my mind. I will create a discussion for our dev camp.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -371,14 +369,92 @@ private extension UITableViewCell {
...
+	) {
+		self.removeBackground()
+		
+		switch section.background {
+		case .none:
+			break
+		case .greyBoxed:
+			
+			// Give all subviews in the content view of the cell some insets to the border. If we change the contentView's constraints, it would not affect the subviews.
+			if let subview = self.subviews.first?.subviews.first {
```
ArturFriesen, line 400: Is `subview` the root view inside the content view?

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/DynamicTableViewController.swift
```diff
@@ -371,14 +369,92 @@ private extension UITableViewCell {
...
+		at indexPath: IndexPath
+	) {
+		self.removeBackground()
+		
+		switch section.background {
+		case .none:
+			break
+		case .greyBoxed:
+			
+			// Give all subviews in the content view of the cell some insets to the border. If we change the contentView's constraints, it would not affect the subviews.
```
ArturFriesen, line 399:
```suggestion
			// Give the root view in the content view of the cell some insets to the border. If we change the contentView's constraints, it would not affect the subviews.
```

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewIconCell.swift
pascal-brause: I reverted this and restored the old state. I retested it and it worked without my change. I think it was something from the beginning i forgot to remove it.
Good eye!

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettings/NotificationSettingsViewController.swift
```diff
@@ -23,26 +26,17 @@ class NotificationSettingsViewController: UIViewController, UITableViewDataSourc
...
-
-		tableView.delegate = self
-		tableView.dataSource = self
-		tableView.separatorColor = .enaColor(for: .hairline)
-
-		navigationItem.title = AppStrings.NotificationSettings.navigationBarTitle
-		navigationItem.largeTitleDisplayMode = .always
-		navigationController?.navigationBar.prefersLargeTitles = true
+		
+		setupView()
```
ArturFriesen, line 30: Is it correct, that `setupView ` is called twice? See line 39.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewIconCell.swift
```diff
@@ -76,7 +77,18 @@ class DynamicTableViewIconCell: UITableViewCell {
 		iconImageView.tintColor = customTintColor ?? tintColor
 		iconImageView.image = image
 		iconImageView.isHidden = image == nil
-
+		
+		// swap label and image so the image is set to the right. Because it is initialized to the left, we need this only for .right
+		if imageAlignment == .right,
+		   let imageView = stackView.subviews.first {
+			stackView.removeArrangedSubview(imageView)
+			stackView.setNeedsLayout()
```
ArturFriesen, line 85: Are the `setNeedsLayout` and `layoutIfNeeded` needed?
Would it be an opportunity to pass the imageAlignment at init?
If not, configure should also handle the swap back to right.

src/xcode/ENA/ENA/Source/Scenes/Settings/NotificationSettings/NotificationSettingsViewController.swift
pascal-brause: That is one too much. Removed one.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTableViewIconCell.swift
pascal-brause: Refactored it a bit - please have a look again @ArturFriesen.
Inject it in the init is not possible because of the inheritance of the DynamicTableViewCell. I would have to make too many changes.

ArturFriesen (approved):

naveeddotio (approved):

PR #3650, opened 2021-10-06
Fix/#9824 settings out of sync

pascal-brause:
## Description
This fixes the settings overview, which was not in sync with the system settings set for the notifications because we had the two old flags in the sync condition.
I removed then the two old notification flags from our code and replaces where necessary with the system's flag for notifications.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9824

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Settings/Settings/SettingsViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Settings/Settings/SettingsViewController.swift
@@ -256,8 +256,7 @@ final class SettingsViewController: UITableViewController, ExposureStateUpdating
 		currentCenter.getNotificationSettings { [weak self] settings in
 			guard let self = self else { return }
 
-			if (settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional)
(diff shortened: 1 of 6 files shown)


src/xcode/ENA/ENA/Source/Scenes/Settings/Settings/SettingsViewController.swift
```diff
@@ -256,8 +256,7 @@ final class SettingsViewController: UITableViewController, ExposureStateUpdating
 		currentCenter.getNotificationSettings { [weak self] settings in
 			guard let self = self else { return }
 
-			if (settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional)
-				&& (self.store.allowRiskChangesNotification || self.store.allowTestsStatusNotification) {
+			if settings.authorizationStatus == .authorized || settings.authorizationStatus == .provisional {
```
pascal-brause, line 259: The real fix. We do not set the old properties any more.

marcussc (approved):

KaiTeuber (approved):

PR #3699, opened 2021-10-14
Fix/10019 home restructure layout problem

30mar:
## Description

- Fix Extra space in home screen between sections
- Add Localized strings for app information and settings screens

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10019


KathrinLuecke (approved):
UA approved

src/xcode/ENA/ENA/Source/View Helpers/AppStrings.swift
```diff
@@ -1071,6 +1066,9 @@ enum AppStrings {
 	}
 
 	enum AppInformation {
+		static let appInformationNavigationTitle = NSLocalizedString("Home_AppInformationNavigation_Title", comment: "")
+		static let appInformationVersion = NSLocalizedString("Home_AppInformationCard_Version", comment: "")
```
nickguendling, line 1070:
```suggestion
		static let appInformationNavigationTitle = NSLocalizedString("App_Information_Title", comment: "")
		static let appInformationVersion = NSLocalizedString("App_Information_Version", comment: "")
```

That way the naming would be aligned with the other ones on the same screen. And please move the second one (the renamed `Home_AppInformationCard_Version`) to the same `/* App Information */` section in the Localizable.strings file as it is not used on the Home Screen.

nickguendling (approved):

naveeddotio (approved):

PR #3706, opened 2021-10-15
Feature/9804 recycle bin/9980 certificate scan

ArturFriesen:
## Description
After a certificate was scanned, the logic now checks if the certificate is already in the recycle bin.
If this is true, an alert is shown to the user and the certificate is restored from the bin.
This PR also removes the 'canRestore' handling from certificate restoration handler, because we don't need it so far.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9980


CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -54,11 +54,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 
 		self.client = HTTPClient(environmentProvider: environmentProvider)
 		self.wifiClient = WifiOnlyHTTPClient(environmentProvider: environmentProvider)
+		self.recycleBin = RecycleBin(store: store)
 
 		self.downloadedPackagesStore.keyValueStore = self.store
 
 		super.init()
 
+		recycleBin.testRestorationHandler = TestRestorationHandlerFake()
```
marcussc, line 63: Fake in Production? :o

nickguendling (approved):

marcussc (approved):

PR #3711, opened 2021-10-18
Attempt to fix flaky FileScannerCoordinatorViewModelTests.

ArturFriesen:
## Description
Execute processing for tests on main.
Fix double callback call.
Add more information to the assertion logs.

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
@@ -78,7 +78,7 @@ class FileScannerCoordinatorViewModel: FileScannerProcessing {
 	init(
 		qrCodeDetector: QRCodeDetecting,
 		qrCodeParser: QRCodeParsable,
-		queue: DispatchQueue = .main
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -78,7 +78,7 @@ class FileScannerCoordinatorViewModel: FileScannerProcessing {
 	init(
 		qrCodeDetector: QRCodeDetecting,
 		qrCodeParser: QRCodeParsable,
-		queue: DispatchQueue = .main
+		queue: DispatchQueue = DispatchQueue.global(qos: .background)
```
marcussc, line 81:
```suggestion
		queue: DispatchQueue = DispatchQueue.global(qos: . userInitiated)
```
Feels like a better choice.
https://developer.apple.com/documentation/dispatch/dispatchqos/qosclass/userinitiated

ArturFriesen: Done

marcussc (approved):

naveeddotio (approved):

PR #3730, opened 2021-10-20
Support parsing multiple codes from the same QRCodeParser instance

nickguendling:
## Description
Fixes a bug where having multiple QR codes in one Document led to the same specialized QR code parser instance being used and not returning correctly or returning with an incorrect `.success` or `.failure`.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10012

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/QRScanner/Model/Parser/QRCodeParser.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/QRScanner/Model/Parser/QRCodeParser.swift
@@ -33,6 +33,7 @@ class QRCodeParser: QRCodeParsable {
 		qrCode: String,
 		completion: @escaping (Result<QRCodeResult, QRCodeParserError>) -> Void
 	) {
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/QRScanner/Model/Parser/QRCodeParser.swift
```diff
@@ -60,15 +61,18 @@ class QRCodeParser: QRCodeParsable {
 			)
 		}
 
-		guard let parser = parser else {
+		guard parser != nil else {
```
marcussc, line 64: Why not `guard let`?

Leaving the scope of the function should dealloc the instance as well, without having the need to set it to `nil` explicitly

marcussc (approved):

KaiTeuber (approved):

ArturFriesen (approved):

PR #3733, opened 2021-10-21
Prevent calling showHome() twice.

ArturFriesen:
## Description
This change prevents calling showHome() twice during startup.

Detailed explanation:
We can have different paths of delegate callbacks depending on the app was started with a QR-Code scan or not.
Possible paths after QR-Code was scanned:
App was suspended: didFinishLaunchingWithOptions -> applicationDidBecomeActive -> continue userActivity
App was in background: continue userActivity -> applicationDidBecomeActive
Either 'continue userActivity' or 'applicationDidBecomeActive' needs to show the UI.

Before that fix,  'applicationDidBecomeActive' was calling showHome() and after that 'continue userActivity' was also calling showHome().

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10103

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
@@ -655,7 +655,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 	}()
 
 	/// - Parameter launchOptions: Launch options passed on app launch
-	/// - Returns: A `Route` if a valid URL is passed in the launch options
+	/// - Returns: `true` if `launchOptions` contains user activity of type `NSUserActivityTypeBrowsingWeb`, returns `false` otherwhise.
 	private func appLaunchedFromUserActicityURL(_ launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -121,12 +121,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
...
+
+			// Return false, because if the app is disabled, we cannot handle URL ressources or user activity.
+			// More information: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622921-application
+			return false
 		}
 
-		// Check for any URLs passed into the app – most likely via scanning a QR code from event or antigen rapid test
-		// Route will be executed in 'applicationDidBecomeActive'
-		route = routeFromLaunchOptions(launchOptions)
+		// 'appLaunchedFromUserActivityURL' inidcates, if the app was launched through a QR-Code scan.
```
marcussc, line 130:
```suggestion
		// 'appLaunchedFromUserActivityURL' inidcates, if the app was launched through a QR-Code scan, from the System Camera.
```
Thats correct right?

marcussc (approved):

naveeddotio (approved):

PR #3736, opened 2021-10-21
Chore/fix flaky statistics tests

naveeddotio:
## Description
PR to fix flaky tests for statistics


src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -878,6 +879,41 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		}
 	}
 	
+	private func setupSelectedRegionsForTesting() {
```
marcussc, line 882: Hmm seems like our already big Appdelegate will grow and grow if we countinue to add more testing data. 
We could move all of the *ForTesting methods all together into there own extension of Appdelegate or one extension for each *ForTesting method.

marcussc (approved):

nickguendling (approved):

PR #3763, opened 2021-10-28
Fix/9486 concurrent log writes

marcussc:
## Description
Logging was not able to log from multiple Threads/Tasks without scrambling the Log file.
With the magic of using a serial Queue to coordinate the `seekToEndOfFile()` and `write` this should no longer happen.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9486

## Screenshots
Imagine nicer logs that don't interrupt each other anymore. Nice, right?


ArturFriesen (approved):
🤩 Noice!

nickguendling (approved):

PR #3778, opened 2021-11-02
Feature/9802 recycle corona test/10155 restore on reregistration

nickguendling:
## Description
Restores a corona test if it has been moved to the recycle bin and the user scans the QR code again.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10155

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
@@ -394,13 +394,18 @@ class QRScannerCoordinator {
 					switch self.recycleBin.canRestore(recycleBinItem) {
 					case .success:
 						self.qrScannerViewController?.dismiss(animated: true) {
-							self.recycleBin.restore(recycleBinItem)
-
 							switch self.presenter {
 							case .submissionFlow, .onBehalfFlow:
+								let parentPresentingViewController = self.parentViewController?.presentingViewController
+
 								// Dismiss submission/on behalf submission flow
-								self.parentViewController?.dismiss(animated: true)
-							case .checkinTab, .certificateTab, .universalScanner, .none:
+								parentPresentingViewController?.dismiss(animated: true) {
+									self.parentViewController = parentPresentingViewController
+									self.restoreAndShow(recycleBinItem: recycleBinItem)
+								}
+							case .checkinTab, .certificateTab, .universalScanner:
+								self.restoreAndShow(recycleBinItem: recycleBinItem)
+							case .none:
 								break
 							}
 						}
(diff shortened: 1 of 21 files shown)


src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -135,34 +144,78 @@ class QRScannerCoordinator {
...
+		case let .coronaTest(testRegistrationInformation):
+			showScannedTestResult(testRegistrationInformation)
+		case let .certificate(certificateResult):
+			showScannedHealthCertificate(certificateResult)
+		case let .traceLocation(traceLocation):
+			showScannedCheckin(traceLocation)
+		}
 	}
 
+	// swiftlint:disable cyclomatic_complexity
```
ArturFriesen, line 157: The warning should be resolved by moving some code into own funcs.

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -135,34 +144,78 @@ class QRScannerCoordinator {
...
-		switch presenter {
-		case .submissionFlow:
-			didScanCoronaTestInSubmissionFlow?(testRegistrationInformation)
-		case .onBehalfFlow:
-			let parentPresentingViewController = parentViewController?.presentingViewController
+		let recycleBinItemToRestore: RecycleBinItem?
+		switch testRegistrationInformation {
+		case .pcr(guid: _, qrCodeHash: let qrCodeHash),
+			.antigen(qrCodeInformation: _, qrCodeHash: let qrCodeHash):
+			recycleBinItemToRestore = store.recycleBinItems.first {
```
ArturFriesen, line 165: Can we use the func `func item(for identifier: String)` from `RecycleBin` here?

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -135,34 +144,78 @@ class QRScannerCoordinator {
...
+
+		if let recycleBinItemToRestore = recycleBinItemToRestore {
+			showTestRestoredFromBinAlert(recycleBinItem: recycleBinItemToRestore)
+			return
+		}
+
+		qrScannerViewController?.dismiss(animated: true) { [weak self] in
+			guard let self = self else { return }
+
+			switch self.presenter {
```
ArturFriesen, line 184: I wonder if the logic in this switch shouldn't be the same which is called after `showTestRestoredFromBinAlert` is finished. Do we have a redundant implementation?

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -332,5 +375,123 @@ class QRScannerCoordinator {
...
+
+		let footerViewController = FooterViewController(footerViewModel)
+		let topBottomViewController = TopBottomContainerViewController(
+			topController: overwriteNoticeViewController,
+			bottomController: footerViewController
+		)
+
+		let navigationController = NavigationControllerWithLargeTitle(rootViewController: topBottomViewController)
+
+		switch self.presenter {
```
ArturFriesen, line 452: This logic looks similar to other calls. Can we have one func to handle it at one place?

ArturFriesen (changes requested):

CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
nickguendling: Sadly not because the `recycleBinIdentifier` for a coronaTest can't just be the `qrCodeHash` as that property is optional and the recycle bin needs to work for items that have this set to nil as well.

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
nickguendling: Absolutely! Done 👍

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
nickguendling: The details are different, I think this is not easy to generalize. E.g. in this case we start the coordinator with `CoronaTestRegistrationInformation`, in `showTestRestoredFromBinAlert` the coordinator is started with the `CoronaTestType`.

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
nickguendling: The logic looks similar but differs in details, I don't see an easy way here to handle it in a uniform way.

ArturFriesen (approved):

Issue #3409, opened 2021-08-24
Digital COVID certificates validity check: List of countries is always in the language which was active during the onboarding


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details
- Device name: iPhone 6
- iOS version: 15 Beta 6
- App version: 2.8.1

## Describe the bug
The names of the countries which can be selected in the "Validity of Certificate" screen does not adapt its translation correctly. They always stay in the language which was active during the onboarding.

## Steps to reproduce the issue
1. Open the iOS Settings
2. Scroll down to "Corona-Warn"
3. Select "Sprache" 
4. Select "English"
5. Open the Corona-Warn-App
6. Click on "Certificates"
7. Click on "Add Certificate" and scan the QR code of a certificate if you don't have any certificates registered in the app
8. Click on a certificate
9. Click on "Check Validity"
10. Click on "Country"
11. See that the names of the countries appear in German instead of in English

## Expected behaviour
The names of the countries should be localized correctly.

## Additional context
1. There is no way for me (currently) to get the app back to show the English names...
2. Please let me know if you need an error log. 

## Credits
This issue was discovered by @dsarkar!

---
Internal Tracking ID: [EXPOSUREAPP-10180](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10180)

Ein-Tim:
With version 2.8 the issue is exactly the other way around. English always shows the German names...
Will adapt the OP later.

ndegendogo:
Analysis: the list of countries is stored persistently including the localised names of the countries. As long as the list does not change (e.g. additional countries participate), it uses this stored list and does not recalculate the localisation.

dsarkar:
@ndegendogo Thanks for analysing, I forward your info to the internal ticket.

ndegendogo:
I found three variables in the Store related to Countries:
[...]

PR #3802, opened 2021-11-06
fix/10180 localisation of countries for certificate validation

ndegendogo:
## Description

This PR fixes a localisation issue that was reported as #3409. The localised names of the countries are written to the persistent store. If the user later changes their language settings, the picker for the certificate validity check still uses the original localised country names.

The solution recalculates the localisation for the country names when reading them from the store.

## Link to Jira

Jira: Internal Tracking ID: Internal Tracking ID: [EXPOSUREAPP-10180](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10180)

github: #3409

## Screenshots

Simulator Screen Shot - iPhone SE (1st generation) - 2021-11-06 at 21 14 28


pascal-brause (changes requested):
Sorry that you had to wait so long for a review!
In generel is your solution nice, but i have one problem with manipulating the data of a database in the getter:
When someone uses the country code in another view controller, one would get the manipulated value but not the original one. But what if he wants the original one.
So my proposal would be to move your code from the getters into computed properties of the view model where the values are read out. There the computed (and manipulated) values are returned but the original values are not changed.

ndegendogo:
@pascal-brause thanks for your review.

I am not sure if I understand your proposal correctly - so let's try.
I identified 3 variables in the Store related to countries which are affected by the localisation issue.

**a) lastSelectedValidationCountry**
it is used as initial selected country for the spinner. HealthCertificateValidationViewModel reads it from the Store, and this is already the model for the spinner; so here is indeed a good place to correct the localisation.

**b) validationOnboardedCountriesCache**
it is used to populate the spinner with countries. HealthCertificateValidationOnboardedCountriesProvider reads it from the Store, and then it flows on several pathes till it comes as parameter in the constructor to HealthCertificateValidationViewModel. So good places to correct the localisation would be either the HealthCertificateValidationOnboardedCountriesProvider (before it fans out), or the HealthCertificateValidationViewModel (after it fans in again). So I'll use the HealthCertificateValidationViewModel according to your proposal. 
*Of course, the risk is here if anybody uses that variable later for a different purpose, they must remember to localise it, else this issue pops up again.*

**c) submissionCountries**
this is not related to certificate validation, but this list suffers from the same localisation issue.
Should I correct it, or should I better leave it alone?
I could use the computed property of ENAExposureSubmissionService for this purpose.

*it is late today - I'll continue tomorrow and change the code ...*

ndegendogo:
@pascal-brause 

Thinking again about this, it boils down to where (which layer) your architecture sees the responsibility to localize the country names.

Currently I feel tempted to expect this responsibility in your service layer, especially as your service points provide a list of `Country`, and each such country has a field with a localized name.

So, of course, you guys are the owners and the masters of your architecture. If you see the responsibility to localize these country names downstream of the service layer - fine with me. *However, in that case I would add a short comment (one-liner) to the service point to resolve this potential pitfall / ambiguity.*

As far as I see, the spinner for the country selection for certificates is currently the only consumer of the list of countries that participate in the Health Certificate system.
For the list of countries that participate in the Gateway for Exposure tracing I am aware of at least 2 consumers (visited countries in the DEK submission flow, and the list of countries shown in the Settings menu).

pascal-brause:
After thinking about your comments and investigating a bit more i understand your points:
- that there is no clear solution for this problem (not only creating a property in one view model) 
- at the end, every caller of these three properties will have the same problem. 

Sorry, i did not see the carrying width problem at my first review.  

I think, because of that, your approach is fine for me and i will approve it.

BTW: The main problem is, i think, that we do not trigger the download of the county list every time when we enter the validation screen. With the download, the list would be created new and there the correct locale would be token. But the tech spec defines here a caching to prevent too many HTTP calls on server and client side so the list is downloaded once and then is not edited later correctly.

Thank you for your performed work on this bug!

pascal-brause (approved):

marcussc:
@ndegendogo Looks like there is a linter warning forENA/ENA/Source/Workers/Store/SecureStore.swift which prevents the merge, in this case you can go ahead and disable it.

ndegendogo:
> @ndegendogo Looks like there is a linter warning forENA/ENA/Source/Workers/Store/SecureStore.swift which prevents the merge, in this case you can go ahead and disable it.

@marcussc oops - thanks for notifying me! 👍👍

naveeddotio (approved):

PR #3812, opened 2021-11-08
use current app configuration instead of waiting up to 5 minutes

KaiTeuber:
## Description
Avoid waiting for the AppConfiguration for the QR Code parser and present self created events.
Always use the current App configuration instead of force load it from the server - this may block up to 5 minutes the UI.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10453


marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/QRScanner/Model/Parser/CheckinQRCodeParser.swift
```diff
@@ -57,7 +57,7 @@ class CheckinQRCodeParser: QRCodeParsable {
 		onSuccess: @escaping((TraceLocation) -> Void),
 		onError: @escaping((CheckinQRScannerError) -> Void)
 	) {
-		appConfigurationProvider.appConfiguration().sink { [weak self] appConfig in
+		appConfigurationProvider.currentAppConfig.sink { [weak self] appConfig in
```
nickguendling, line 60: An issue in both of this cases could be that as long as they are in memory the `currentAppConfig` publisher could publish again and again and the completion handlers could be called multiple times which then could lead to unexpected behavior. I guess it would be safer to just synchronously access the current app config once.

nickguendling (changes requested):

nickguendling (approved):

PR #3830, opened 2021-11-10
Extend error handling.

ArturFriesen:
## Description
Extend the error handling according to the spec.
Error handling now recognizes wrong initialization vectors and encryption key.

### DIFF
--- a/src/xcode/ENA/ENASecurity/Sources/ENASecurity/CBCEncryption.swift
+++ b/src/xcode/ENA/ENASecurity/Sources/ENASecurity/CBCEncryption.swift
@@ -6,16 +6,23 @@ import Foundation
 import CommonCrypto
 
 public enum CBCEncryptionError: Error {
-    case EncryptionFailed(Int)
+    case AES_CBC_INVALID_KEY // if key cannot be used for AES
+    case AES_CBC_INVALID_IV // Initialization Vector is empty or does not have the length of the provided contraint ivLengthConstraint.
+    case UNKNOWN(Int)
 }
 
 public struct CBCEncryption {
 
     // MARK: - Init
 
-    public init(encryptionKey: Data, initializationVector: Data) {
+    public init(
+        encryptionKey: Data,
+        initializationVector: Data,
+        ivLengthConstraint: Int? = nil
+    ) {
         self.encryptionKey = encryptionKey
         self.initializationVector = initializationVector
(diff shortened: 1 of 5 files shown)


src/xcode/ENA/ENASecurity/Sources/ENASecurity/CBCEncryption.swift
```diff
@@ -32,8 +39,20 @@ public struct CBCEncryption {
 
     private let encryptionKey: Data
     private let initializationVector: Data
+    private let ivLengthConstraint: Int?
 
     private func crypt(data: Data, option: CCOperation) -> Result<Data, CBCEncryptionError> {
+
+        guard !initializationVector.isEmpty else {
```
marcussc, line 46: 🤔 wouldn't it be an idea to check for those constraints during `init` ?
Seems a little late to check it here.

ArturFriesen: Since init does not have type safe error handling (only optional or throws), I would like to avoid failable initializers.

src/xcode/ENA/ENASecurity/Tests/ENASecurityTests/GCMEncryptionTests.swift
```diff
@@ -167,6 +167,64 @@ final class GCMEncryptionTests: XCTestCase {
...
+              }
+    }
+
+    func test_When_Decrypt_With_ivLengthConstraint_Then_Error_AES_GCM_INVALID_IV() {
+        let testData = testDatas[0]
+
+        // initializationVector of size 13 should return an error with ivLengthConstraint of 16
+        let initializationVector = Data(capacity: 13)
+
+        print("initializationVector.isEmpty: \(initializationVector.isEmpty)")
```
marcussc, line 205: old debugging print? 😅

marcussc (approved):

src/xcode/ENA/ENASecurity/Tests/ENASecurityTests/GCMEncryptionTests.swift
ArturFriesen: removed

nickguendling (approved):

naveeddotio (approved):

PR #3831, opened 2021-11-10
Replace entfernen by löschen

CV113:
Replace entfernen by löschen


marcussc (approved):

KathrinLuecke (commented):
@CV113 Is it ok to also change the legal strings? I thought we were not supposed to change them?

CV113:
> @CV113 Is it ok to also change the legal strings? I thought we were not supposed to change them?

oh, I'm not sure- I thought it was aligned with all colleagues (even with the colleagues of the legal department) to replace entfernen by löschen. Let me ask and I will come back to you asap

naveeddotio (changes requested):
Legal text has to be approved by the lawyers. Please take Martin Hermes into the loop before changing anything even one word.

KathrinLuecke (approved):
UA approved

naveeddotio (approved):
@CV113 Checked with Martin and it's okay to change the texts.

Issue #3834, opened 2021-11-11
Link to App informations in "easy language" missing in English version


ndegendogo:
## Avoid duplicates
* [ ] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [ ] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 8
- iOS version: 14.4.1
- App version: 2.13.1 (0)

## Describe the bug
In the German version, the page "Über die App" has a link to "Informationen zur App in leichter Sprache und Gebärdensprache". This link is missing when I switch to English.

## Steps to reproduce the issue
- on the 'Status' tab, scroll down
- on the 'More' tile, select App information
- select About the app
- scroll down to the bottom 
- in the German version here is a link. But in the English version, the link is missing

see screenshots

## Expected behaviour

See the same link in the English versipn (possibly to a localised page)

## Additional context

![7D3FCBE7-907F-408B-9144-84F4A8B08005](https://user-images.githubusercontent.com/6118267/141289014-08190f07-a312-494d-b595-06b19db94d9b.png)

![60A65AFE-9A87-4D51-8531-57D22959A61D](https://user-images.githubusercontent.com/6118267/141289065-c3b87aea-05a0-437f-9142-ec8f20f4da2d.png)

Ein-Tim:
@ndegendogo 

Is there an equivalent English page to the German one?

ndegendogo:
>

dsarkar:
@ndegendogo Thanks for the report. @Ein-Tim have you checked Android, how does it look there?

Ein-Tim:
@dsarkar let me check quickly.

Ein-Tim:
Yes, there is this link under Android:
![Screenshot_20211111_163814_de rki coronawarnapp](https://user-images.githubusercontent.com/67682506/141325992-73bf080d-0041-47ad-a3bf-0e331bea525a.jpg)

It links to the German page.

ndegendogo:
Initial analysis:
- the variable name of the link text (as shown on the page) is AppStrings.AppInformation.aboutLinkText
- its value is NSLocalizedString("App_Information_About_LinkText", ...)
- its translation is available in all 6 languages
[...]

PR #3842, opened 2021-11-14
Fix/10565 link to App informations in easy language

ndegendogo:
## Description

This PR fixes an issue that was reported as #3834.
The link to app informations in easy language and sign language on the "About the App" page is shown only for German localisation. The same link is also referenced on the first Onboarding page, same issue here.

The solution removes the artificial (and supposedly unintended) blocking of this link for the other languages.
Note: the name of this link is already localised.
The link itself is pointing to the same German-language URL for all languages.
This is the same behaviour as in the Android app for the [About page](https://github.com/corona-warn-app/cwa-app-ios/issues/3834#issuecomment-966403895). 
*However, the Android app does not have this link in the [Onboarding flow](https://github.com/corona-warn-app/cwa-app-ios/issues/3834#issuecomment-968362728).*

## Link to Jira

Jira: Internal Tracking ID: [EXPOSUREAPP-10565](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10565)
github: #3834

## Screenshots

Simulator Screen Shot - iPhone 8 - 2021-11-14 at 16 08 12

---
Internal Tracking ID: [EXPOSUREAPP-10565](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10565)


ndegendogo:
Simulator Screen Shot - iPhone 8 - 2021-11-14 at 21 51 22

davidlindercwa:
Currently, this easy-to-read information is only available in German - and hence the link should only be shown when the app is set to German language (as is the current behavior).

This means that we'll close this PR for now, but thanks anyway for the proposal!

Ein-Tim:
This PR is practically identical to #4576, which will be merged into the `release/2.23` branch.

Issue #2179, opened 2021-02-15
Support link preview pop-up


Ein-Tim:
## Current Implementation
Currently (v. 1.12.1, v. 1.13, v 1.14.3), when [holding](https://user-images.githubusercontent.com/67682506/107942410-7c364200-6f8b-11eb-9c52-afdee3b2ad28.mov) links, nothing happens, except for the "New Features" screen which is showing after an app update, there the preview [works](https://user-images.githubusercontent.com/67682506/110522429-2d567500-8111-11eb-86a5-8d1738e98d48.png).

## Suggested Enhancement
Support link previews everywhere in the app, like Safari or GitHub does ([Screen recording](https://user-images.githubusercontent.com/67682506/107942163-2792c700-6f8b-11eb-8a0b-cf61e0c0836d.MP4)).

## Expected Benefits
Consistency with the "New Features" screen, the "Terms of Use" screen, the "Data Privacy" screen, the "Imprint" screen and most other iOS apps. Also, the user has quick(er) access to content. 

---
Internal Tracking ID: [EXPOSUREAPP-5624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5624)

dsarkar:
@Ein-Tim I am planning to do an issue [Overview] UX/UI with this issue in the wishlist repo, the translucent tab bar, etc. If you are happy to create such an issue go ahead, please. These overviews help quite a lot to get organized and direct the attention of the team to certain topics.

Ein-Tim:
@dsarkar 

I can't create it right now but today afternoon.

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-5624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5624)

Ein-Tim:
@dsarkar Please see https://github.com/corona-warn-app/cwa-wishlist/issues/386 for the overview.
I will extend it with Android issues soon, but feel free to already add some.

ndegendogo:
@Ein-Tim could you please test this again? I think it is at least partly resolved.
I see the current behaviour:
(cwa 2.13.1(0) / iOS 14.4.1 / iPhone 8, tested with a few links)

- I press long on the link and get a menu together with a small window that shows the link in full length, and an option to preview.
- if I opt for the preview, next time I get the menu and the small window shows the preview, and an option to not preview

[...]

PR #3843, opened 2021-11-14
fix/5624 enable link preview for onboarding

ndegendogo:
## Description

This PR starts to address an issue that was reported as #2179. The app has a very inconsistent behaviour with regard to links. Several links support a preview behaviour if you press the link long, but several links don't support this. 
The desired behaviour is to support the preview.

The PR adds preview behaviour for the "link to app informations in easy language and sign language" from the first page of the Onboarding flow, and thus aligns the behaviour of this link to the "About the App" page, where the same link is referenced (here the preview behaviour is already implemented).

~Analysis shows that two concurring mechanisms were implemented to open the URL from the Onboarding page.
(1) the link was implemented as an AttributedString in a TextView. In this case iOS opens the link. This mechanism supports the preview.
(2) at the same time, the view controller was setup as delegate for the TextView, and the view controller was extended to open the link with LinkHelper. This mechanism overrides the first mechanism, but it does not support the preview.~

~The solution removes the second mechanism (delegate and extension).
**Note:** Unfortunately this solution will have a negative impact on the robustness of the unit tests.
The LinkHelper mechanism mocks the URL access for UITesting configuration and thus decouples the execution of the unit tests from the true network.
The AttributedString mechanism, however, does not have this mocking.
I am not sure how to resolve this dilemma.
One option would be to bring back the extension, but to setup the delegate only for UITesting (although that solution would be a bit intransparent and needs to be commented well).~

**Edit:** I found an alternate and much better solution. The textview delegate takes now care of the default interaction (open the link with LinkHelper), but allows iOS to perform the alternate actions (like the action menu or the preview).
For unit tests opening the link as well as preview is mocked.

## Link to Jira

Jira: Internal Tracking ID: [EXPOSUREAPP-5624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5624)

github: #2179

## Screenshots

https://user-images.githubusercontent.com/6118267/141700658-97ad290d-fe09-4959-b418-0afcccb507d3.mp4

---
 Internal Tracking ID: [EXPOSUREAPP-5624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5624)


dsarkar:
@ndegendogo Thanks. Forwarded to Internal Tracking ID: EXPOSUREAPP-5624

ndegendogo:
I have another idea.
I am going to try if I can move this logic into LinkHelper.

***Stay tuned ...***

marcussc (approved):
Looks better and more general then the old PR 👍

ndegendogo:
@marcussc full ack 😁
I hope I can later use this solution as kind of 'blueprint' to improve handling for all the other external links in the system (some don't yet support preview, some don't yet support the mocking)

naveeddotio (approved):

PR #3845, opened 2021-11-15
Feature/10349 crypto functions part1/10437 generating an ec key pair with curve p 256

30mar:
## Description
Add ec key pair generation

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10437

The screenshot is of a command line check to check if the generated public key string is valid
## Screenshots
Screenshot 2021-11-15 at 15 36 14

### DIFF
--- a/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
+++ b/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
@@ -45,4 +45,8 @@ public struct ECKeyPairGeneration {
       }
       return (privateKey, nil)
     }
+    
+    public func generatePublicKey(from privateKey: SecKey) -> SecKey? {
+        return SecKeyCopyPublicKey(privateKey)
+    }
 }
--- a/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
+++ b/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
@@ -57,7 +57,7 @@ public struct ECKeyPairGeneration {
             let convertedError = error!.takeRetainedValue() as Error
             return (nil, convertedError.localizedDescription)
         }
-        let modifiedData = appendHeaderToData(data: keyCFData as Data)
+        let modifiedData = prependHeaderToData(data: keyCFData as Data)
         
         return (modifiedData, nil)
     }
--- a/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
+++ b/src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
@@ -68,4 +68,12 @@ public struct ECKeyPairGeneration {
       "\(Bundle.main.bundleIdentifier ?? "app").\(name)".data(using: .utf8)!
     }
     
+    private func appendHeaderToData(data: Data) -> Data {
+        var appendedData = data
+        let headerBytes = [0x30, 0x59, 0x30, 0x13, 0x06, 0x07, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x02, 0x01, 0x06, 0x08, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07, 0x03, 0x42, 0x00] as [UInt8]
+        appendedData.insert(contentsOf: headerBytes, at: 0)
+        
+        return appendedData
+    }
+
 }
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
```diff
@@ -0,0 +1,79 @@
...
+        return (modifiedData, nil)
+    }
+    
+    // MARK: - Private
+    
+    private func tag(for name: String) -> Data {
+      "\(Bundle.main.bundleIdentifier ?? "app").\(name)".data(using: .utf8)!
+    }
+    
+    private func appendHeaderToData(data: Data) -> Data {
```
marcussc, line 71:
```suggestion
    private func prependHeaderToData(data: Data) -> Data {
```

src/xcode/ENA/ENASecurity/Tests/ENASecurityTests/ECKeyPairGenerationTests.swift
```diff
@@ -0,0 +1,32 @@
...
+        let generationClass = ECKeyPairGeneration()
+        let generationOutput = generationClass.generatePrivateKey(with: "test")
+        guard let privateKey = generationOutput.0,
+              let publicKey = generationClass.generatePublicKey(from: privateKey) else {
+                  XCTFail()
+                  return
+              }
+        
+        guard let publicKeyData = generationClass.generateData(from: publicKey).0 else {
+            XCTFail()
```
marcussc, line 22: An Error message would be nice so we can see directly on the CI what went wrong here.

src/xcode/ENA/ENASecurity/Tests/ENASecurityTests/ECKeyPairGenerationTests.swift
```diff
@@ -0,0 +1,32 @@
...
+@testable import ENASecurity
+
+final class ECKeyPairGenerationTests: XCTestCase {
+    
+    func test_validPublicKeyBase64_isGenerated() {
+        let generationClass = ECKeyPairGeneration()
+        let generationOutput = generationClass.generatePrivateKey(with: "test")
+        guard let privateKey = generationOutput.0,
+              let publicKey = generationClass.generatePublicKey(from: privateKey) else {
+                  XCTFail()
```
marcussc, line 17: And Error message would be nice so we can see directly on the CI what went wrong here.

marcussc (changes requested):

src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
```diff
@@ -0,0 +1,79 @@
...
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+public struct ECKeyPairGeneration {
+    
+    // MARK: - Public
+
+    public func generatePrivateKey(with name: String? = nil) -> (SecKey?, String?) {
```
ArturFriesen, line 11: I think we should have a common API Style. In the other cases we are using Result type with explicit errors.

src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
```diff
@@ -0,0 +1,79 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+public struct ECKeyPairGeneration {
```
ArturFriesen, line 7: API which returns 3 values, like described in the tech spec, is missing.

ArturFriesen (changes requested):

src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
```diff
@@ -0,0 +1,108 @@
...
+    case dataGenerationFromKeyFailed(String?) // if key cannot be used for AES
+}
+
+public struct ECKeyPairGeneration {
+    
+    // MARK: - Public
+
+    func generateECPair(with name: String? = nil) -> Result<ECKeyPair, ECKeyPairGenerationError> {
+        let secureKey = self.generatePrivateKey(with: name)
+        guard let privateKey = secureKey.0 else {
```
ArturFriesen, line 19: Not mandatory, just for information: If the other functions would also return a Result type, you could just concatenate all the calls with flatmap. This way this func will only be a couple of lines. (https://useyourloaf.com/blog/using-swift-result-and-flatmap/)

ArturFriesen (approved):

src/xcode/ENA/ENASecurity/Sources/ENASecurity/ECKeyPairGeneration.swift
30mar: they are not returning a Result type, they return an error message or no error and based on the function i return the needed enum

marcussc (approved):

PR #3855, opened 2021-11-16
Feature/10350 ticketvalidation consent/10537 UI test

pascal-brause:
## Description
Adds new ui testplan for ticket validation to our project and to circle ci.
Adds inside the new test class a detailed UI test for the first consent screen (and here we can extend then later)
Adds a small test for the universal scanner.

## Link to Jira
Mainstory: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10350
Subtask: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10537

## Screenshots
Nothing to see it's a ui test.


src/xcode/ENA/ENAUITests/TicketValidation/ENAUITests_19_TicketValidation.swift
```diff
@@ -0,0 +1,36 @@
...
+import XCTest
+
+class ENAUITests_19_TicketValidation: CWATestCase {
+	var app: XCUIApplication!
+	
+	override func setUpWithError() throws {
+		try super.setUpWithError()
+		continueAfterFailure = false
+		app = XCUIApplication()
+		setupSnapshot(app)
```
marcussc, line 14:
```suggestion
```
No snapshots/screenshots in the tests so no setup needed.

src/xcode/ENA/ENAUITests/UniversalQRCodeScanner/ENAUITests_16_UniversalQRCodeScanner.swift
```diff
@@ -98,5 +98,18 @@ class ENAUITests_16_UniversalQRCodeScanner: CWATestCase {
 		/// Exposure Submission QR Info Screen
 		XCTAssertTrue(app.staticTexts[AccessibilityLabels.localized(AppStrings.ExposureSubmissionQRInfo.title)].waitForExistence(timeout: .short))
 	}
+	
+	func test_TicketValidationTestFromUniversalQRCodeScanner() throws {
+		app.launch()
+		
+		app.buttons[AccessibilityIdentifiers.TabBar.scanner].waitAndTap()
+		
+		/// Simulator only Alert will open where you can choose what the QRScanner should scan
```
marcussc, line 107: Do we really need this test? Seems to do the same as `test_isFirstConsentShown` in  `ENAUITests_19_TicketValidation`

If it is still needed I think it would be better to put it in `ENAUITests_19_TicketValidation:`

marcussc (changes requested):

src/xcode/ENA/ENAUITests/TicketValidation/ENAUITests_19_TicketValidation.swift
pascal-brause: aye aye

src/xcode/ENA/ENAUITests/UniversalQRCodeScanner/ENAUITests_16_UniversalQRCodeScanner.swift
pascal-brause: yeah the test is very similar but here we test if the route from the universal scanner is doing correctly and checks only if the correct screen opens. So the test is from my point of view correct in set `ENAUITests_16_UniversalQRCodeScanner`.
The test in `ENAUITests_19_TicketValidation` is more detailed and checks if all elements of the consent screen are existent.
What they have in common is the starting point, because the ticket validation can only be triggered from the universal scanner.
Still need to change @marcussc ?

marcussc: So `ENAUITests_19_TicketValidation` has the same test, but is more detailed? If I understand it correct, I would say just remove it from `ENAUITests_16_UniversalQRCodeScanner `

nickguendling (approved):

30mar (approved):

marcussc (approved):

PR #3863, opened 2021-11-19
Add encrypt and sign of DCC.

ArturFriesen:
## Description
Adds Encrypting and Signing of a DCC.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10639


src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidation.swift
```diff
@@ -43,5 +44,92 @@ final class TicketValidation: TicketValidating {
...
+		dccBarcodeData: String,
+		nonceBase64: String,
+		encryptionScheme: EncryptionScheme,
+		publicKeyForEncryption: SecKey,
+		privateKeyForSigning: SecKey
+	) -> Result<EncryptAndSignResult, EncryptAndSignError> {
+		
+		// Determine iv (initializationVector): the iv shall be derived as the byte sequence (Data / ByteArray) of nonceBase64 using base64 encoding.
+		guard let initializationVector = Data(base64Encoded: nonceBase64) else {
+			return .failure(.UNKNOWN)
```
marcussc, line 58: Maybe add a Log statement for each return what step failed? Could make it easier to debug.

src/xcode/ENA/ENASecurity/Sources/ENASecurity/Utils/SecKey+Creation.swift
```diff
@@ -0,0 +1,41 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+public extension SecKey {
+    
+    // We need this since Apple expects only the raw keys no headers allowed. 🙄 and we cant use CryptoKit since we have to support iOS 12
+    // Depending on where else we will need this in our awesome app we might have to move this extension into its own file. Currently its only need for unit testing
```
marcussc, line 10: Since its move to its own extension now you can remove the comment :D

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Services/TicketValidation/Models/TicketValidationInitializationData.swift
```diff
@@ -13,4 +13,26 @@ struct TicketValidationInitializationData: Codable {
 	let consent: String
 	let subject: String
 	let serviceProvider: String
+	
+	static func fake (
```
nickguendling, line 17:
```suggestion
	static func fake(
```

Also, I think you can move this to an extension in a separate file and for now only include it in the test target so we don't accidentally use this in production.

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidation.swift
```diff
@@ -43,5 +44,113 @@ final class TicketValidation: TicketValidating {
 	func cancel() {
 
 	}
-
+	
+	func encryptAndSignDCC(
```
nickguendling, line 48: I think this does not necessarily have be part of the public API of `TicketValidation`, maybe it makes sense to move it to a separate struct or class and just use it from the `validate()` func of the `TicketValidation`? Not sure if we can already use it, maybe we need a new subtask to integrate the steps in here once we have all the parts available.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidation.swift
ArturFriesen: Added logs.

src/xcode/ENA/ENA/Source/Services/TicketValidation/Models/TicketValidationInitializationData.swift
ArturFriesen: done

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidation.swift
ArturFriesen: Moved the code to a own component. Created task for integration.

nickguendling (approved):

Issue #3868, opened 2021-11-20
Unused link definitions in AppStrings


ndegendogo:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: n/a
- iOS version: n/a
- App version: lates source code under development as of 2021-11-20
branch release/2.15.x commit 0deabb62045f6a2f11761b0ce8f317ca65a79568

## Describe the bug

1) The definition AppStrings.Links.healthCertificateFAQ seems not referenced.
git annotate shows that it was added on 06 May.
So, either it was forgotten to add this link to some page, or it is obsolete meanwhile.
**Note:** the definition resolves to NSLocalizedString("HealthCertificate_Info_register_FAQLink",...)
and that resolves to the localised version of the general FAQ URL, like "https://www.coronawarn.app/de/faq/"

2) I just found another one.
Also AppStrings.ExposureNotificationError.learnMoreURL seems not referenced.
It resolves to NSLocalizedString("ExposureNotificationSetting_GeneralError_LearnMore_URL"...)
and that also resolves to the localised version of the general FAQ URL, like "https://www.coronawarn.app/de/faq/"
git annotate shows that it was added on 03 July 2020.

## Possible Fix
Find a good place to use these link definitions, or delete them.

ndegendogo:
@dsarkar I have created a PR #3869 under the assumption that the definition can be deleted.
Please apply the labels. Thanks!

ndegendogo:
Closing, the fix was merged.

PR #3869, opened 2021-11-20
fix/3868 remove unreferenced link definitions

ndegendogo:
## Description

This PR removes two unreferenced link definitions that were reported as #3868.

Of course I cannot know for sure the true reason for this.
The definition `AppStrings.Links.healthCertificateFAQ` was added on 06 May 2021, so I don't believe that this is an "unborn child" (work in progress).
My suspicion is rather that it is either no longer used and was just forgotten to clean up.
Or it was planned to be used for some purpose but this was forgotten to implement.

The definition `AppStrings.ExposureNotificationError.learnMoreURL` was added on 03 July 2020; also here I believe it was just forgotten to clean up after it was no longer used.

## Link to Jira

Jira: n/a
github: #3868


pascal-brause (approved):
Nice finding. Completely right.
For the future: It's sufficient to remove the localized string from the DE file, the others are synced in a internal process. Keys that are removed in the DE file are then removed in the other language files, too.

CV113 (approved):
UA approved

naveeddotio (approved):

PR #3886, opened 2021-11-24
Add possibility for detailed trust evaluation error handling.

ArturFriesen:
## Description
This PR adds the possibility to have detailed error handling for trust evaluation.


pascal-brause (approved):

nickguendling (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Security/EvaluateTrust/DynamicEvaluateTrust.swift
```diff
@@ -38,9 +38,14 @@ struct DynamicEvaluateTrust: EvaluateTrust {
 			completionHandler(.useCredential, URLCredential(trust: trust))
 		case .failure(let error):
 			Log.debug("AuthenticationChallenge failed with error \(error.localizedDescription)", log: .client)
+			trustEvaluationError = error
```
KaiTeuber, line 41: When will is set to nil? Otherwise it might block all future requests?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Security/EvaluateTrust/DynamicEvaluateTrust.swift
ArturFriesen: Good catch!!! 
Reseting it now after the completion call.

KaiTeuber (approved):

PR #3888, opened 2021-11-25
Story/10733 validation decorator unit tests

30mar:
## Description
Added unit tests for the Validation Decorator service

## Link to Jira
(https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10733)


src/xcode/ENA/ENA/Source/HTTPClient/Resources/SIDValidationDecoratorResource.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-struct ServiceIdentityDocumentValidationDecoratorResource: Resource {
+struct SIDValidationDecoratorResource: Resource {
```
nickguendling, line 7: I would prefer not to introduce abbreviations, I think the name was more precise and easier to understand before

src/xcode/ENA/ENA/Source/HTTPClient/Resources/__tests__/SIDValidationDecoratorResourceTests.swift
```diff
@@ -0,0 +1,138 @@
...
+		serviceProvider.load(resource) { result in
+			switch result {
+			case .success:
+				XCTFail("client error is expected!")
+			case .failure(let error):
+				guard case let .receivedResourceError(resourceError) = error,
+					  .VD_ID_CLIENT_ERR == resourceError else {
+						  XCTFail("unexpected error case")
+						  return
+					  }
```
pascal-brause, line 33:
```suggestion
				guard case let .receivedResourceError(. VD_ID_CLIENT_ERR) = error else {
						  XCTFail("Expected .VD_ID_CLIENT_ERR but received: \(error)")
						  return
					  }
```

pascal-brause: If you don't compare the error but just capture it (and for alle the remaining tests, too), you will not need to implement the Equatable protocol for `ServiceIdentityValidationDecoratorError`.

pascal-brause (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/SIDValidationDecoratorResource.swift
30mar: yes i just reverted it now

src/xcode/ENA/ENA/Source/HTTPClient/Resources/__tests__/SIDValidationDecoratorResourceTests.swift
30mar: yeah you are right, i had it first as i used to compare but now it is obsolete, i will remove the equatable conformance

pascal-brause (approved):

nickguendling (approved):

PR #3921, opened 2021-12-02
Update Localizable.strings

KathrinLuecke:
Change of UniversalQRScanner_InstructionDescription
Changed "Tests, Zertifikate,..." to "Tests, Ihre Zertifikate,.." to make clear that the QR-Scanner should only be used to scan own certificates, but not to check certificates of others.


marcussc (approved):

nickguendling (approved):

CV113 (approved):
UA approved

PR #3925, opened 2021-12-02
Fix/10885 ticket validation broken

KaiTeuber:
## Description
Validate a ticket twice (or more) was claimed to be broken.

It turned out that the current behavior is correct, but the error codes should be extended 400...499 & 500...599 for all errors.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10885


nickguendling (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/ServiceIdentityDocumentResource.swift
```diff
@@ -55,9 +55,9 @@ struct ServiceIdentityDocumentResource: Resource {
 			return .VS_ID_NO_NETWORK
 		case .unexpectedServerError(let statusCode):
 			switch statusCode {
-			case (400...409):
+			case (400...499):
```
ArturFriesen, line 58: @mlenkeit Might be also be updated in the tech spec?

src/xcode/ENA/ENA.xcodeproj/project.pbxproj
```diff
@@ -9843,7 +9843,7 @@
 				34079851274D3D41007E6071 /* HealthCertificateService+UITestingMethods.swift in Sources */,
 				BA5B7FD626458EF400502087 /* HealthCertificateSimpleTextCellViewModel.swift in Sources */,
 				01B7232F24FE4F080064C0EB /* OptionGroupViewModel.swift in Sources */,
-				ABC506CF274BE30700698A95 /* Locator+ServiceIdentityDocument.swift in Sources */,
+				ABC506CF274BE30700698A95 /* ServiceIdentityDocument.swift in Sources */,
```
ArturFriesen, line 9846: I think this name is misleading. It looks like its the "ServiceIdentityDocument". But it only contains the extension for the locator. I would propose to keep "Locator+ServiceIdentityDocument.swift". And rename the other files to the same schema.

ArturFriesen: Or maybe other way around ""ServiceIdentityDocument+Locator.swift""

PR #3933, opened 2021-12-03
improved numberformatters to correct rounding

KaiTeuber:
## Description
Wrong rounded values fixed in the statistic tiles.
Improved NumberFormatters a bit and added a new unit test for that special case.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10616

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/Extensions/SAP_Internal_Stats_KeyFigure+Formatting.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/Extensions/SAP_Internal_Stats_KeyFigure+Formatting.swift
@@ -5,18 +5,19 @@
 import UIKit
 
 extension SAP_Internal_Stats_KeyFigure {
+
+	// MARK: - Internal
+
 	var formattedValue: String? {
 		let decimals = max(0, Int(self.decimals))
-		let numberFormatter = NumberFormatter()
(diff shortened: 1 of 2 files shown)


naveeddotio (approved):

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/Cells/Statistics/Extensions/SAP_Internal_Stats_KeyFigure+Formatting.swift
```diff
@@ -5,18 +5,19 @@
...
+
+	// MARK: - Internal
+
 	var formattedValue: String? {
 		let decimals = max(0, Int(self.decimals))
-		let numberFormatter = NumberFormatter()
-		numberFormatter.numberStyle = .decimal
+		let numberFormatter = Self.decimalNumberFormatter
 		numberFormatter.minimumFractionDigits = Int(decimals)
 		numberFormatter.maximumFractionDigits = Int(decimals)
```
nickguendling, line 15: Moving this number formatter to a static let makes it not thread safe anymore if multiple KeyFigures are asked for their `formattedValue` in parallel, as the same shared object is reconfigured now. Unlikely that there are any practical implications here, but I want to mention it.

nickguendling (changes requested):

nickguendling (approved):

PR #3939, opened 2021-12-06
Fix/10925 key submission metadata

marcussc:
## Description
Key Submission data must only be submitted when we have a positive test result. 
This PR fixes this.

## Link to Jira
(https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10925)

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
+++ b/src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
@@ -255,18 +255,23 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 		submitted is true
 		OR
 		- differenceBetweenTestResultAndCurrentDateInHours >= hoursSinceTestResultToSubmitKeySubmissionMetadata
+		AND
+		Test result is positive
 		*/
 
 		let isSubmitted: Bool
 		let _testResultReceivedDate: Date?
+		let isTestResultPositive: Bool
 
 		switch type {
 		case .pcr:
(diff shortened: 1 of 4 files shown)


pascal-brause (approved):

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
```diff
@@ -281,7 +286,7 @@ final class PPAnalyticsSubmitter: PPAnalyticsSubmitting {
 		   differenceBetweenTestResultAndCurrentDateInHours >= hoursSinceTestResultToSubmitKeySubmissionMetadata {
 			timeDifferenceFulfillsCriteria = true
 		}
-		return isSubmitted || timeDifferenceFulfillsCriteria
+		return (isSubmitted || timeDifferenceFulfillsCriteria) && isTestResultPositive
```
KaiTeuber, line 289: why not return earlier  if `isTestResultPositive` is true?
I think all the stuff later isn't need for this criteria.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Services/PPAnalyticsSubmitter/PPAnalyticsSubmitter.swift
marcussc: Was think about this as well but this way we the code look exactly like the doc comment on top of the function :)

mlenkeit (approved):
LGTM 👍

KaiTeuber (approved):

PR #3940, opened 2021-12-06
Fix for crash at adding progress child 

pascal-brause:
## Description
Possibly fixes the well known Progress-Crash in our app. The crash is made when adding more child progresses then allowed. I added two guards to prevent adding a progress when it's already finished or cancelled in assumption that already finished or cancelled progresses crashed the app when you try to add them.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4368

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
@@ -65,7 +65,7 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
 		writtenPackages: WrittenPackages,
 		completion: @escaping (Result<[ENExposureWindow], Error>) -> Void
 	) -> Progress {
-		let progress = Progress(totalUnitCount: 2)
+		let progress = Progress(totalUnitCount: Int64.max)
 
 		let detectExposuresProgress = exposureDetector.detectExposures(
 			configuration: configuration,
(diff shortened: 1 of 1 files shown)


marcussc (approved):

30mar (approved):

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
```diff
@@ -65,7 +65,7 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
 		writtenPackages: WrittenPackages,
 		completion: @escaping (Result<[ENExposureWindow], Error>) -> Void
 	) -> Progress {
-		let progress = Progress(totalUnitCount: 2)
+		let progress = Progress(totalUnitCount: Int64.max)
```
ArturFriesen, line 68: Why is this change needed? totalUnitCount will never be reached with this number.

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
```diff
@@ -82,13 +82,27 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
 						completion(.failure(error))
 					}
 				}
+				
+				// Prevent adding a child which is perhaps already finish or cancelled because this would crash the app.
```
ArturFriesen, line 86:
```suggestion
				// Prevent adding a child which is perhaps already finished or cancelled because this would crash the app.
```

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
```diff
@@ -82,13 +82,27 @@ final class ExposureDetectionExecutor: ExposureDetectionDelegate {
...
+				
 				Log.info("2nd child added to progress", log: .riskDetection)
 				progress.addChild(exposureWindowsProgress, withPendingUnitCount: 1)
 			} else if let error = error {
 				self.clearCacheOnErrorBadParameter(error: error)
 				completion(.failure(error))
 			}
 		}
+		
+		// Prevent adding a child which is perhaps already finish or cancelled because this would crash the app.
```
ArturFriesen, line 101:
```suggestion
		// Prevent adding a child which is perhaps already finished or cancelled because this would crash the app.
```

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Exposure Transaction/ExposureDetectionExecutor.swift
pascal-brause: I forgot to mention it in the description: There are two assumptions to prevent the crash: on the one side we build in the guards and the other thing is: The crash occurs whenever we want to add another progress, but the max amount is already reached. So why not allow more progresses? I found nothing out, why there are only 2 totalUnitCount allowed, so i took the maximum amount.
Does it give you a stomach ache @ArturFriesen?

ArturFriesen: It looks wrong to me. I think at the moment we are not using the progress unit counts. We only use the progress to cancel it. As soon as we start using the progress unit counts 'Int64.max' will be wrong.

ArturFriesen (approved):

PR #3943, opened 2021-12-07
introduce some optional indentation for the bullet points

pascal-brause:
## Description
Fixes an UI problem that some legal texts were out off. The fix is not to manipulate the leading anchor of the stack view (which cut of labels randomly) but make it possible to enter some indentation before the bullet points so the text is set before the rendering of the cell is done.

This fixes also that when scanning a QR code, the label with "QR code wird gelesen" was cut of. This was fixes by giving the label 0 number of lines (Screenshot was hard to take because the label is only short time visible and this explains the transparency).

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10918

## Screenshots

10918-indenation
10918-label


KaiTeuber:
Is it really supposed to look like this with the transparency in the 'QR Code get read' box?

30mar (approved):

nickguendling (approved):

naveeddotio (approved):

PR #3947, opened 2021-12-07
Ticket Validation - Fix long error messages

naveeddotio:
## Description
The issue was the undesired long error messages, in order to fix it, we introduced a property on serviceError, so we could get the error message straight away in ticketValidationError. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10872

## Screenshots
WhatsApp Image 2021-12-07 at 16 59 14


src/xcode/ENA/ENA/Source/HTTPClient/Service/ServiceError.swift
```diff
@@ -8,6 +8,16 @@ import ENASecurity
 The errors that can occur while using the service and calling http methods.
 */
 enum ServiceError<RE>: Error, Equatable where RE: Error {
+	// we need this property for nested errors
+	var errorDescription: String {
```
ArturFriesen, line 12: Does it also work for the other nested errors?

naveeddotio: Currently, we need to exact error type for this specific case receivedResourceError, so it is currently confined to this case only. The implementation will remain untouched for other cases.

naveeddotio: Just to make sure if we all are on same page, it will cover all the messages for Ticket Validation.

ArturFriesen: What about trustEvaluationError?

naveeddotio: This is beyond the scope of this bug ticket.

nickguendling, line 12:
```suggestion
enum ServiceError<RE>: Error, CustomStringConvertible, Equatable where RE: Error {

		// MARK: - Protocol CustomStringConvertible

		var description: String {
```

I think it would be more fitting to use the `CustomStringConvertible` protocol as it is not really a full errorDescription but rather a nicer identifier.

And it would be nice to move the implementation below the cases, next to the `Equatable` implementation.

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidationError.swift
```diff
@@ -25,33 +25,59 @@ enum TicketValidationError: LocalizedError {
...
+			case .VD_ID_EMPTY_X5C, .VD_ID_NO_ATS_SIGN_KEY, .VD_ID_NO_ATS_SVC_KEY, .VD_ID_NO_ATS, .VD_ID_NO_VS_SVC_KEY, .VD_ID_NO_VS:
 				return "\(serviceProviderError) (\(error))"
-			default:
-				return "\(AppStrings.TicketValidation.Error.tryAgain) (\(error))"
+			case .REST_SERVICE_ERROR(let serviceError):
+				switch serviceError {
+				case .receivedResourceError(.VD_ID_CLIENT_ERR), .receivedResourceError(.VD_ID_PARSE_ERR):
+					return "\(serviceProviderError) (\(serviceError.errorDescription))"
+				default:
+					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(error))"
```
nickguendling, line 35:
```suggestion
					return "\(serviceProviderError) (\(serviceError))"
				default:
					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(serviceError))"
```

With `CustomStringConvertible` conformance you will get the description by just using the error like this again. And we should use the unpacked error for the `default` case as well to get the nicer error message for the other error cases as well.

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidationError.swift
```diff
@@ -25,33 +25,59 @@ enum TicketValidationError: LocalizedError {
...
 			switch error {
-			case .REST_SERVICE_ERROR(.receivedResourceError(.VS_ID_CERT_PIN_MISMATCH)), .REST_SERVICE_ERROR(.receivedResourceError(.VS_ID_CERT_PIN_HOST_MISMATCH)), .REST_SERVICE_ERROR(.receivedResourceError(.VS_ID_CLIENT_ERR)), .VS_ID_EMPTY_X5C, .VS_ID_NO_ENC_KEY, .VS_ID_NO_SIGN_KEY, .REST_SERVICE_ERROR(.receivedResourceError(.VS_ID_PARSE_ERR)):
+			case .VS_ID_EMPTY_X5C, .VS_ID_NO_ENC_KEY, .VS_ID_NO_SIGN_KEY:
 				return "\(serviceProviderError) (\(error))"
+			case .REST_SERVICE_ERROR(let serviceError):
+				switch serviceError {
+				case .receivedResourceError(.VS_ID_CERT_PIN_MISMATCH), .receivedResourceError(.VS_ID_CERT_PIN_HOST_MISMATCH), .receivedResourceError(.VS_ID_CLIENT_ERR), .receivedResourceError(.VS_ID_PARSE_ERR):
+					return "\(serviceProviderError) (\(serviceError.errorDescription))"
+				default:
+					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(error))"
```
nickguendling, line 47:
```suggestion
					return "\(serviceProviderError) (\(serviceError))"
				default:
					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(serviceError))"
```

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidationError.swift
```diff
@@ -25,33 +25,59 @@ enum TicketValidationError: LocalizedError {
...
 			switch error {
-			case .ATR_AUD_INVALID, .REST_SERVICE_ERROR(.receivedResourceError(.ATR_CERT_PIN_MISMATCH)), .REST_SERVICE_ERROR(.receivedResourceError(.ATR_CERT_PIN_NO_JWK_FOR_KID)), .REST_SERVICE_ERROR(.receivedResourceError(.ATR_CLIENT_ERR)), .ATR_JWT_VER_ALG_NOT_SUPPORTED, .ATR_JWT_VER_EMPTY_JWKS, .ATR_JWT_VER_NO_JWK_FOR_KID, .ATR_JWT_VER_NO_KID, .ATR_JWT_VER_SIG_INVALID, .ATR_PARSE_ERR, .REST_SERVICE_ERROR(.receivedResourceError(.ATR_PARSE_ERR)), .ATR_TYPE_INVALID:
+			case .ATR_AUD_INVALID, .ATR_JWT_VER_ALG_NOT_SUPPORTED, .ATR_JWT_VER_EMPTY_JWKS, .ATR_JWT_VER_NO_JWK_FOR_KID, .ATR_JWT_VER_NO_KID, .ATR_JWT_VER_SIG_INVALID, .ATR_PARSE_ERR, .ATR_TYPE_INVALID:
 				return "\(serviceProviderError) (\(error))"
+			case .REST_SERVICE_ERROR(let serviceError):
+				switch serviceError {
+				case .receivedResourceError(.ATR_CERT_PIN_MISMATCH), .receivedResourceError(.ATR_CERT_PIN_NO_JWK_FOR_KID), .receivedResourceError(.ATR_CLIENT_ERR), .receivedResourceError(.ATR_PARSE_ERR):
+					return "\(serviceProviderError) (\(serviceError.errorDescription))"
+				default:
+					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(error))"
```
nickguendling, line 63:
```suggestion
					return "\(serviceProviderError) (\(serviceError))"
				default:
					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(serviceError))"
```

src/xcode/ENA/ENA/Source/Services/TicketValidation/TicketValidationError.swift
```diff
@@ -25,33 +25,59 @@ enum TicketValidationError: LocalizedError {
...
 			switch error {
-			case .REST_SERVICE_ERROR(.receivedResourceError(.RTR_CERT_PIN_MISMATCH)), .REST_SERVICE_ERROR(.receivedResourceError(.RTR_CERT_PIN_HOST_MISMATCH)), .REST_SERVICE_ERROR(.receivedResourceError(.RTR_CLIENT_ERR)), .RTR_JWT_VER_ALG_NOT_SUPPORTED, .RTR_JWT_VER_EMPTY_JWKS, .RTR_JWT_VER_NO_JWK_FOR_KID, .RTR_JWT_VER_NO_KID, .RTR_JWT_VER_SIG_INVALID, .RTR_PARSE_ERR, .REST_SERVICE_ERROR(.receivedResourceError(.RTR_PARSE_ERR)):
+			case .RTR_JWT_VER_ALG_NOT_SUPPORTED, .RTR_JWT_VER_EMPTY_JWKS, .RTR_JWT_VER_NO_JWK_FOR_KID, .RTR_JWT_VER_NO_KID, .RTR_JWT_VER_SIG_INVALID, .RTR_PARSE_ERR:
 				return "\(serviceProviderError) (\(error))"
+			case .REST_SERVICE_ERROR(let serviceError):
+				switch serviceError {
+				case .receivedResourceError(.RTR_CERT_PIN_MISMATCH), .receivedResourceError(.RTR_CERT_PIN_HOST_MISMATCH), .receivedResourceError(.RTR_CLIENT_ERR), .receivedResourceError(.RTR_PARSE_ERR):
+					return "\(serviceProviderError) (\(serviceError.errorDescription))"
+				default:
+					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(error))"
```
nickguendling, line 79:
```suggestion
					return "\(serviceProviderError) (\(serviceError))"
				default:
					return "\(AppStrings.TicketValidation.Error.tryAgain) (\(serviceError))"
```

nickguendling (changes requested):
I tried the implementation, it works for most cases already, except for not using the unpacked error for the `default` case. Apart from that I just made a small suggestion to use a fitting protocol.

src/xcode/ENA/ENA/Source/HTTPClient/Service/ServiceError.swift
nickguendling: The trust evaluation errors are usually mapped to custom errors with the right prefix, so it should work fine for the TicketValidation scope at least, I think

ArturFriesen: Yes with this implementation, the short errors only work, if the error is mapped to receivedResourceError. It will not work with not mapped errors or other errors like invalidRequestError, resourceError.

naveeddotio: I have added the case for trust evaluation error.

nickguendling (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Service/ServiceError.swift
nickguendling: I think it makes sense to keep e.g. invalidRequestError and resourceError as it is so the outer error is visible so we can easily distinguish them (both have a ResourceError inside). There is no defined separate error code for them. From my perspective it should be okay to keep longer error description in these cases because they are not foreseen by the tech spec and we might need all the info we can get to debug.

ArturFriesen (approved):

PR #3968, opened 2021-12-10
Replace our logging with fallback system logging.

pascal-brause:
## Description
We have a crash at logging. This fixes the crash by preventing using our custom logger when we want to create a log that our custom logger could not be created 😌. But to have this information, we use the system logging in that case.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10672

### DIFF
--- a/src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
+++ b/src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
@@ -332,7 +332,9 @@ struct FileLogger {
 			let fileHandle = try? FileHandle(forWritingTo: url)
 			return fileHandle
 		} catch {
-			Log.error("File handle error", log: .localData, error: error)
+			// We must not use our Log here because it would produce a crash (we want to log in case we cannot create a log 🤪)
+			// swiftlint:disable:next no_direct_oslog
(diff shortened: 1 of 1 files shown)


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Workers/Logging/Logging.swift
```diff
@@ -332,7 +332,9 @@ struct FileLogger {
 			let fileHandle = try? FileHandle(forWritingTo: url)
 			return fileHandle
 		} catch {
-			Log.error("File handle error", log: .localData, error: error)
+			// We must not use our Log here because it would produce a crash (we want to log in case we cannot create a log 🤪)
+			// swiftlint:disable:next no_direct_oslog
+			os_log("%{public}@ %{public}@", log: .default, type: .error, "Error while creating log file handler. Fallback to system logging to log this error.")
```
marcussc, line 337:
```suggestion
			os_log("%{public}@", log: .default, type: .error, "Error while creating log file handler. Fallback to system logging to log this error.")
```
Should be enough right?

pascal-brause:
Changed target branch to 2.15 and created new PR: https://github.com/corona-warn-app/cwa-app-ios/pull/3975

PR #3976, opened 2021-12-13
Dismiss screen allow bypasing legal text

naveeddotio:
## Description
With the current implementation, if a user click on the dismiss button on the antigen test profile info screen, the screen is not shown again, which is not the expected behaviour. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10456

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
@@ -1160,7 +1160,6 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 				self?.showAntigenTestProfileInput(editMode: false)
 			},
 			dismiss: { [weak self] in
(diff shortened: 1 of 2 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -1159,7 +1159,10 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 			didTapContinue: { [weak self] in
 				self?.showAntigenTestProfileInput(editMode: false)
 			},
-			dismiss: { [weak self] in self?.dismiss() }
+			dismiss: { [weak self] in
```
nickguendling, line 1162: I'm wondering if it wouldn't be a simpler/easier to understand later solution to do it the other way around and move the `viewModel.markScreenSeen()` in `AntigenTestProfileInformationViewController` from `viewDidLoad()` to `didTapFooterViewButton()`, then we don't need to reset it here.

nickguendling (approved):

marcussc (approved):

PR #3979, opened 2021-12-13
Fix/10868 improve grouping standardized name

naveeddotio:
## Description
Improvement in grouping name to reflect change in tech spec:

https://github.com/corona-warn-app/cwa-app-tech-spec/pull/146/files

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10868


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
```diff
@@ -24,9 +24,9 @@ extension Name {
 	var standardizedName: String {
 		return [standardizedGivenName, standardizedFamilyName].formatted()
 	}
-
+	
 	var groupingStandardizedName: String {
-		return standardizedName
+		return trimmedStandardizedName
 			.replacingOccurrences(of: "\\s+", with: " ", options: .regularExpression)
 			.replacingOccurrences(of: "<+", with: "<", options: .regularExpression)
```
ArturFriesen, line 31: Is this still needed? After trimming "<" this line should never be executed?

naveeddotio: Trimming is done from leading and trailing, this regular expression checks if we have more than one < and it replaces with a single <

30mar (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/__test__/Name+ExtensionTests.swift
```diff
@@ -17,6 +17,15 @@ class Name_ExtensionTests: XCTestCase {
 		XCTAssertEqual(name.groupingStandardizedName, name.standardizedName)
 		XCTAssertEqual(name.groupingStandardizedName, "THOMAS<ARMIN MEYER")
 	}
+	
+	func testGIVEN_ValidName_THEN_GroupingStandardizedNameIsTrimmed() {
+		// GIVEN
+		let name = Name(familyName: "Meyer ", givenName: "Thomas Arming ", standardizedFamilyName: "MEYER<", standardizedGivenName: "THOMAS<ARMIN<"
```
KaiTeuber, line 23:
```suggestion
		let name = Name(familyName: "Meyer ", givenName: "Thomas Armin ", standardizedFamilyName: "MEYER<", standardizedGivenName: "THOMAS<ARMIN<"
```
It the test working correct?

naveeddotio: Good catch. But the test is working correct because it it only takes into account standardizedGivenName.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
naveeddotio: Eg: in case when standardizedGivenName is THOMAS<<ARMIN<

After trimming, we shall get THOMAS<<ARMIN
and with this regex, we will get THOMAS<ARMIN while grouping

nickguendling (changes requested):
I think you need to increase `kCurrentHealthCertifiedPersonsVersion` in HealthCertificateService.swift to make the migration run again for existing persons.

nickguendling (approved):

ThiefMaster:
Just wondering.. why is the tech spec repo not public?

PR #3980, opened 2021-12-13
Fix/11002 no caching used by http client

pascal-brause:
## Description
This fixes the problem, that the new client did not returned the cached content in case of no network.
This PR introduces also default value for a resource, for the allowlist an empty set.
This PR changes and adds a lot of logs for the different services.

## Link to Jira
(https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11002)


KaiTeuber (approved):

ArturFriesen (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
```diff
@@ -69,18 +71,35 @@ extension Service {
 				}
 				
 				if let error = error {
-					completion(.failure(customError(in: resource, for: .transportationError(error))))
-					return
+					Log.info("No network connection (.transportationError)", log: .client)
+					// If we have no internet connection, we have to check first if we have a cached model for this resource.
+					if hasCachedData(resource) {
+						Log.info("Found some cached data", log: .client)
+						cached(resource, completion)
```
nickguendling, line 78: I think we need this handling not just for the no network error but also for the other error scenarios (trust evaluation and server error codes 4xx/5xx)

nickguendling (changes requested):

CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
KaiTeuber: Is that really useful? The server responses an error - not found (404) or an internal server error and the client should response with a cached result? I'm no sure if that is the wanted behavior of an eTag based caching.

nickguendling: Well, it is how the desired behaviour is defined in the [tech spec](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/ce23af4fb75ff1ef0f42ef2ee284c11450ad8fbb/docs/spec/dcc-validation-service.md#requesting-the-validation-service-allowlist):

> If the request succeeds with status code 304 Not Modified, the operation shall return the most recent Validation Service Allowlist.
> 
> 
> If the request fails because of a missing or poor network connection, the operation shall log a corresponding error and return the most recent Validation Service Allowlist.
> 
> If the request fails with a client error (i.e. status code 40x), the operation shall log a corresponding error and return the most recent Validation Service Allowlist.
> 
> If the request fails with a server error (i.e. status code 50x), the operation shall log a corresponding error and return the most recent Validation Service Allowlist.

pascal-brause: You are right. For this we refactored a lot: 
- There is now a default value you can define in the resource, which will be returned instead of a fail.
- You can now define caching use cases for .noNetwork and for statusCode ranges per resource.
- And we added several tests for this.

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
```diff
@@ -134,4 +149,115 @@ extension Service {
...
+			case .noNetwork:
+				defaultErrorHandling(error, resource, completion)
+			case .statusCode(let statusCodes):
+				Log.error("Unexpected server error: (\(statusCodes)", log: .client)
+				completion(.failure(customError(in: resource, for: .unexpectedServerError(statusCodes))))
+			}
+		}
+	}
+	
+	/// Handle default error for no network. If the error is nil, it is an invalid response.
```
ArturFriesen, line 243: Honestly I don't Unterstand the error handling only by looking at this function. It would be hard to maintain. 
Why is it an invalid response if the error is nil?
And why is it a transportationError in other cases?

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
```diff
@@ -134,4 +149,115 @@ extension Service {
...
+		}
+		// If not, we will handle now the cache use cases
+		else {
+			Log.info("Found nothing cached. Handling cache use cases")
+			switch cacheUseCase {
+			case .noNetwork:
+				defaultErrorHandling(error, resource, completion)
+			case .statusCode(let statusCodes):
+				Log.error("Unexpected server error: (\(statusCodes)", log: .client)
+				completion(.failure(customError(in: resource, for: .unexpectedServerError(statusCodes))))
```
ArturFriesen, line 238: Why is this not handled in the error handling, but sepereratly?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/__tests__/Resource+Fake.swift
```diff
@@ -0,0 +1,37 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+@testable import ENA
+
+class ResourceFake: Resource {
+	init(
+		locator: Locator = .fake(),
+		type: ServiceType = .caching([]),
```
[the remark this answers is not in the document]
pascal-brause: Good idea. I will add this.

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
pascal-brause: Gotcha, you found an small error. I'll fix that later and write a test for that (did not notice the error because the existing tests did not cover it.)

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
```diff
@@ -134,4 +149,115 @@ extension Service {
...
+		_ resource: R,
+		_ completion: @escaping (Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void
+	) where R: Resource {
+		
+		// Check if we can handle caching use case handling
+		guard case let .caching(usage) = resource.type,
+			  usage.contains(cacheUseCase) else {
+				  // Otherwise, fall back to the default
+				  Log.info("No cache use case handling defined. Fallback to default error handling")
+				  defaultErrorHandling(error, resource, completion)
```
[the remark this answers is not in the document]
pascal-brause: Because the cache policy will be handled only when it is contained in the resource's type. If not, we do not want handle the cache policy

ArturFriesen (approved):

naveeddotio (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Service/Services/_Service+Default.swift
```diff
@@ -157,11 +157,11 @@ extension Service {
 		return nil
 	}
 
-	func hasNoStatusCodeCachePolicy<R>(
+	func hasCachePolicyStatusCode<R>(
```
nickguendling, line 160: I think old word order made more sense: `hasStatusCodeCachePolicy` because it returns if it has a certain cache policy and not if it has a certain status code.

nickguendling (changes requested):

nickguendling (approved):

KaiTeuber:
@ArturFriesen @nickguendling Nice refatoring tipps 👍

PR #4003, opened 2021-12-16
Fix/11037 refactor file scanner errors to show missing errors

naveeddotio:
## Description
Fixing errors for fileScanner. For fileScanner we should use the same errors like in qrScanner for QRCodeParser. Also moved the error alerts from the controller to the coordinator. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11037


src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -59,6 +59,7 @@ protocol FileScannerProcessing {
 	var processingStarted: (() -> Void)? { get set }
 	var processingFinished: ((QRCodeResult) -> Void)? { get set }
 	var processingFailed: ((FileScannerError?) -> Void)? { get set }
+	var parsingFailed: ((QRCodeParserError) -> Void)? { get set }
```
ArturFriesen, line 62: I would propose to keep the API clean. processingStarted -> processingFinished or processingFailed. You could add the QRCodeParserError as an associated value to FileScannerError and call processingFailed with this error.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -8,13 +8,18 @@ import PDFKit
 import OpenCombine
 
 enum FileScannerError: CaseIterable {
+	static var allCases: [FileScannerError] {
+		return [.noQRCodeFound, .fileNotReadable, .invalidQRCode, .photoAccess, .passwordInput, .unlockPDF, .alreadyRegistered, .qrCodeParserError(.scanningError(.codeNotFound))]
+	}
```
KaiTeuber, line 13: if added manually you no longer need to fulfill `CaseIterable`

src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -32,6 +37,8 @@ enum FileScannerError: CaseIterable {
 			return AppStrings.FileScanner.PasswordError.title
 		case .alreadyRegistered:
 			return AppStrings.FileScanner.AlreadyRegistered.title
+		default:
+			return ""
```
KaiTeuber, line 41: Better add a `case . qrCodeParserError` - so we will get errors if the enum gets extended

src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -51,6 +58,8 @@ enum FileScannerError: CaseIterable {
 			return AppStrings.FileScanner.PasswordError.message
 		case .alreadyRegistered:
 			return AppStrings.FileScanner.AlreadyRegistered.message
+		default:
+			return ""
```
KaiTeuber, line 62: Same

src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
```diff
@@ -32,6 +37,8 @@ enum FileScannerError: CaseIterable {
 			return AppStrings.FileScanner.PasswordError.title
 		case .alreadyRegistered:
 			return AppStrings.FileScanner.AlreadyRegistered.title
+		default:
```
ArturFriesen, line 40: Instead of adding a default I would prefer to add the specific case.
With the default in place a future developer would not be warned about adding an string here for a new enum case.

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -718,6 +686,182 @@ class QRScannerCoordinator {
...
+			)
+		)
+		
+		if case .versionError = error {
+			alert.addAction(
+				UIAlertAction(
+					title: AppStrings.TicketValidation.Error.updateApp,
+					style: .default,
+					handler: { _ in
+						LinkHelper.open(urlString: "https://apps.apple.com/de/app/corona-warn-app/id1512595757?mt=8")
```
KaiTeuber, line 721: use links from the AppStrings please. See `AppStrings.Links`

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerCoordinator.swift
```diff
@@ -718,6 +686,182 @@ class QRScannerCoordinator {
...
+				)
+			)
+		}
+
+		DispatchQueue.main.async {
+			self.qrScannerViewController?.present(alert, animated: true)
+		}
+	}
+
+	// swiftlint:disable cyclomatic_complexity
```
KaiTeuber, line 732: opt out that check isn't a good solution.
Better refactor that func so it isn't needed.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/FileScanner/FileScannerCoordinatorViewModel.swift
nickguendling: I think you could also just remove the whole CaseIterable conformance and this manual implementation because I think this is just for a unit that we don't really need

nickguendling (changes requested):

CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/QRScanner/QRScannerViewController.swift
```diff
@@ -47,10 +48,11 @@ class QRScannerViewController: UIViewController {
 						// Don't show an error in simulator to enable debugging/UI-Tests
 						return
 						#else
-						self?.showCameraPermissionErrorAlert()
+						self?.viewModel?.deactivateScanning()
```
nickguendling, line 51: This leads to an infinite loop and crashes the app if the camera permission was denied
```suggestion
```

nickguendling (changes requested):

nickguendling (approved):

KaiTeuber (approved):

PR #4017, opened 2021-12-23
Handling error on key packages download

naveeddotio:
## Description
An attempt to fix the issue:

When key packages could not be downloaded due to any reason (for example: unavailable internet connection), the error is not handled properly and we have to call self.failOnTargetQueue on failure and if the user presses the retry button, nothing happens because the state is not changed and it's still "idle". 

Also added a missing state for .riskManuallyRequested to message (data is being downloaded) even the risk fails instantly, just to tell the user that something has happened. Also added a delay so user can see the message. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10742
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10998


src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
```diff
@@ -43,7 +43,7 @@ class ExposureDetectionViewModel: CountdownTimerDelegate {
 				switch $0 {
 				case .downloading:
 					self?.setupForDownloadingState()
-				case .detecting:
+				case .detecting, .riskRequested:
```
ArturFriesen, line 46: In my opinion, this change will lead to a wrong behavior at the activity state UI. It will lead to showing the detection state immediately. But the correct process is to show the downloading state and the detecting state only if the detection has started.

ArturFriesen (changes requested):
Please see comment.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
nickguendling: I see the same issue as Artur here, you can see the risk card jumping every time the home tab is opened even though we don't actually calculate the risk every time.

naveeddotio: Update: Me and Arthur had a small discussion together with Anna-Maria and I have incorporated those changes.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/MockRiskProvider.swift
```diff
@@ -46,14 +46,13 @@ class MockRiskProvider: RiskProviding {
 
 	func requestRisk(userInitiated: Bool, timeoutInterval: TimeInterval) {
 		for consumer in consumers {
-			consumer.didChangeActivityState?(.riskRequested)
+			userInitiated ? consumer.didChangeActivityState?(.riskManuallyRequested) : consumer.didChangeActivityState?(.riskRequested)
```
ArturFriesen, line 49:
```suggestion
			consumer.didChangeActivityState?(userInitiated ? .riskManuallyRequested : .riskRequested)
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -99,44 +99,50 @@ final class RiskProvider: RiskProviding {
...
+					self.failOnTargetQueue(error: error)
+				}
 			}
 
 			return
 		}
 
 		queue.async {
-			self.updateActivityState(.riskRequested)
+			userInitiated ? self.updateActivityState(.riskManuallyRequested) : self.updateActivityState(.riskRequested)
```
ArturFriesen, line 145:
```suggestion
			self.updateActivityState(userInitiated ? .riskManuallyRequested : .riskRequested)
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -262,7 +268,14 @@ final class RiskProvider: RiskProviding {
 			case .success:
 				completion(.success(()))
 			case .failure(let error):
-				completion(.failure(.failedKeyPackageDownload(error)))
+				if case .noNetworkConnection = error {
+					// we need this delay so the user can see something happened
```
ArturFriesen, line 272: Why is this delay not needed for other errors, which might also occur very fast?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -262,7 +268,14 @@ final class RiskProvider: RiskProviding {
...
+					DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
```
ArturFriesen, line 273: Please dispatch on targetQueue instead of hardcoded main.

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
```diff
@@ -355,8 +356,12 @@ class KeyPackageDownload: KeyPackageDownloadProtocol {
 				case let .success(days):
 					Log.info("KeyPackageDownload: Server data is available for day packages.", log: .riskDetection)
 					completion(.success(days))
-				case .failure:
-					completion(.failure(.uncompletedPackages))
+				case .failure(let error):
+					if case .noResponse = error {
```
ArturFriesen, line 360: Is this the correct case? There is an error .noNetworkConnection. Isn't this the correct one?

ArturFriesen (changes requested):
Please see comment about error case.

src/xcode/ENA/ENA/Source/Services/Risk/KeyPackageDownload/KeyPackageDownload.swift
naveeddotio: Yes, the case .noResponse is correct. It is returned when there is no network in the failure and then we pass .noNetworkConnection which we have introduced here in KeyPackageDownload. I have tested it and it works as expected.

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
```diff
@@ -41,7 +41,7 @@ class ExposureDetectionViewModel: CountdownTimerDelegate {
 				self?.riskProviderActivityState = $0
 
 				switch $0 {
-				case .downloading:
+				case .downloading, .riskManuallyRequested:
 					self?.setupForDownloadingState()
```
ArturFriesen, line 45: I still see a very little, negligible problem with this. It will go from state downloading -> downloading in case of manual reload. Which will lead to a flickering in the UI. But I think its acceptable because the whole scenario is an edge case.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -257,12 +263,15 @@ final class RiskProvider: RiskProviding {
...
 	private func downloadDayPackages(completion: @escaping (Result<Void, RiskProviderError>) -> Void) {
-		keyPackageDownload.startDayPackagesDownload(completion: { result in
+		keyPackageDownload.startDayPackagesDownload(completion: { [weak self] result in
 			switch result {
 			case .success:
 				completion(.success(()))
 			case .failure(let error):
-				completion(.failure(.failedKeyPackageDownload(error)))
+				// we need this delay so the user can see something happened
+				self?.targetQueue.asyncAfter(deadline: .now() + 1.0) {
```
ArturFriesen, line 272: Don't we need this delay only in case of manual reloading?

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
naveeddotio: Since the content is same in both states, I don't think there will be any flickering.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
naveeddotio: Delay of a second in case of an error is negligible, I don't think we should spend effort on tracking if the error happened during manual risk.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/__tests__/RiskProviderTests.swift
```diff
@@ -654,7 +654,7 @@ final class RiskProviderTests: CWATestCase {
 		
 		let didFailCalculateRiskExpectation = expectation(description: "expect didFailCalculateRisk not to be called")
 		
-		let expectedActivityStates: [RiskProviderActivityState] = [.riskRequested, .downloading, .detecting, .idle]
+		let expectedActivityStates: [RiskProviderActivityState] = [.riskManuallyRequested, .downloading, .detecting, .idle]
```
nickguendling, line 657: I'm wondering if we could keep one test that is `userInitiated` and add one that is not so we can test that if it is not initiated by the user the activity state is set to `. riskRequested`?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureDetection/ExposureDetectionViewModel.swift
```diff
@@ -41,7 +41,7 @@ class ExposureDetectionViewModel: CountdownTimerDelegate {
 				self?.riskProviderActivityState = $0
 
 				switch $0 {
-				case .downloading:
+				case .downloading, .riskManuallyRequested:
```
nickguendling, line 44: I'm wondering if need to make the same change in 522 as well and I think we need to extend the `isActive` property of the `RiskProviderActivityState` as well to have the screen correctly configured

nickguendling (approved):

PR #4022, opened 2021-12-31
Error Submitting exposure keys

30mar:
## Description
The community Bug about the exposure submission error is hard to reproduce, there are multiple causes for this generic message, which is why i added these logs so we can locate the problem next time it occurs.
I used static text as printing the description of the enum will include the URLs which we dont want to de

Edit: As requested the ticket comments, the PR now includes the actual message inside the PR.

## Link to Jira
(https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11056)

## Screenshots
Simulator Screen Shot - iPhone 12 14 4 - 2022-01-10 at 18 26 25


ArturFriesen (approved):

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestServiceError.swift
```diff
@@ -34,7 +34,14 @@ enum CoronaTestServiceError: LocalizedError, Equatable {
 				return AppStrings.ExposureSubmissionError.noResponse
 			case .resourceError, .invalidResponse:
 				return AppStrings.ExposureSubmissionError.invalidResponse
-			case .invalidRequestError, .trustEvaluationError, .fakeResponse:
+			case .invalidRequestError:
+				Log.error("ExposureSubmissionError: serviceError invalidRequestError", log: .api)
```
nickguendling, line 38: Wouldn't it make more sense/be cleaner to log where the error is occurring instead of when the localized description is accessed?

nickguendling (changes requested):
See comment

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestServiceError.swift
30mar: mmm but how is that?, i think the bigger issue is that we are not sure what is causing the error and after the point of getting the localized generic message there is no way of knowing the exact cause, or am i missing something here ?

nickguendling (approved):

PR #4035, opened 2022-01-04
Update ExposureDetection_Guide_Home string

krichly:
## Description
There is an unnecessary second dot at the end of the ExposureDetection_Guide_Home string in the English localization.


naveeddotio (approved):

marcussc:
@krichly Thanks for you contribution, sadly this change will not work due to our translation process. I will notify the translation team about this error, so the can be fixed with the next translation delivery.

PR #4041, opened 2022-01-05
Show QR code in pdf generation in correct orientation

pascal-brause:
## Description
This fixes the problem that when creating the pdf of a DCC, the QR code inside the pdf is upside-down.

Does NOT fix the problem with the Spain Travel Health Formular. 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11074

## Screenshots
IMG_4E429D8FF879-1


naveeddotio (approved):

marcussc (approved):

30mar (approved):

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificate+PDF/HealthCertificate+PDF.swift
```diff
@@ -30,9 +30,12 @@ extension HealthCertificate {
 			throw PDFGenerationError.qrCodeCreationFailed
 		}
 
+		// We need to adjust the orientation of the qrCode image because the coordinate system is upside-down. So flip it horizontally.
+		let flippedImage = image.transformed(by: CGAffineTransform(scaleX: 1, y: -1))
```
naveeddotio, line 34: While I being thought Affine Transformation in Computer Graphics course in the university, I thought I would we use it practically :D

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificate+PDF/HealthCertificate+PDF.swift
nickguendling: Wouldn't it make sense to do the flipping inside the `pdfDocumentEmbeddingImageAndText` func where we also flip the whole coordinate system?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificate+PDF/HealthCertificate+PDF.swift
pascal-brause: Not for me, because `pdfDocumentEmbeddingImageAndText` is used multiple times. In all cases, we pass the QRCode upside-down and in `pdfDocumentEmbeddingImageAndText` we flip the coordinate system to match the UIKit's one.  But only in the DCC-pdf-case we need it vice-versa. So for me it makes sense to pass it already to `pdfDocumentEmbeddingImageAndText` so that it is flipped then again but is displayed correctly. @nickguendling what do you think?

nickguendling: Not sure I follow 😅 I think we do not pass it in upside-down anywhere else. We use it for the poster generation for events (and a configurable version of that in the dev menu), and the QR code is upside down in these as well, so I think we should fix it for those instances, too. And from an API standpoint I would not expect to have to flip the image before passing it in, so I think we should flip it inside the shared function.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificate+PDF/HealthCertificate+PDF.swift
pascal-brause: Okay, I moved the code so it makes more sense now. Inside `pdfDocumentEmbeddingImageAndText` we flip the coordinate system, but the image has be "reverted" so it appears correctly. 
I also updated the comment.
Additionally, i checked the checkin qrcode (which i did not checked before) and it appears now correctly, too. And in the developer menu (see additional screen shots). And the unit test is running green, too.

pascal-brause:
IMG_0114
IMG_4A9F99DFC4BF-1

nickguendling (approved):

PR #4044, opened 2022-01-05
Fix/8567 performance improvement

nickguendling:
## Description
Determines the validity states and schedules the notifications on a background thread to improve UI performance when scanning and deleting certificates

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8567

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
@@ -537,7 +537,9 @@ class HealthCertificateService {
 	func updateValidityStatesAndNotifications(shouldScheduleTimer: Bool = true) {
 		Log.info("Update validity state and notifications.")
 
-		DispatchQueue.global(qos: .default).async {
+		DispatchQueue.global(qos: .default).async { [weak self] in
+			guard let self = self else { return }
+
 			let currentAppConfiguration = self.appConfiguration.currentAppConfig.value
 			self.healthCertifiedPersons.forEach { healthCertifiedPerson in
 				healthCertifiedPerson.healthCertificates.forEach { healthCertificate in
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
@@ -537,56 +537,58 @@ class HealthCertificateService {
 	func updateValidityStatesAndNotifications(shouldScheduleTimer: Bool = true) {
 		Log.info("Update validity state and notifications.")
 
-		let currentAppConfiguration = appConfiguration.currentAppConfig.value
-		healthCertifiedPersons.forEach { healthCertifiedPerson in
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -537,56 +537,58 @@ class HealthCertificateService {
...
-					.dgcParameters.blockListParameters.blockedUvciChunks
-				if healthCertificate.isBlocked(by: blockedIdentifierChunks) {
-					healthCertificate.validityState = .blocked
-				} else {
-					let signatureVerificationResult = self.dccSignatureVerifier.verify(
-						certificate: healthCertificate.base45,
-						with: self.dscListProvider.signingCertificates.value,
-						and: Date()
-					)
+					let blockedIdentifierChunks = self.appConfiguration.currentAppConfig.value
```
KaiTeuber, line 553: this might leak memory

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -537,56 +537,58 @@ class HealthCertificateService {
...
-		let currentAppConfiguration = appConfiguration.currentAppConfig.value
-		healthCertifiedPersons.forEach { healthCertifiedPerson in
-			healthCertifiedPerson.healthCertificates.forEach { healthCertificate in
-				let expirationThresholdInDays = currentAppConfiguration.dgcParameters.expirationThresholdInDays
-				let expiringSoonDate = Calendar.current.date(
-					byAdding: .day,
-					value: -Int(expirationThresholdInDays),
-					to: healthCertificate.expirationDate
-				)
+		DispatchQueue.global(qos: .default).async {
```
KaiTeuber, line 540:
```suggestion
		DispatchQueue.global(qos: .default).async { [weak self] in
```
otherwise we capture self in the closure as a strong reference

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
nickguendling: In this case that shouldn't be a big downside, as the closure will be executed and will not be stored afterwards, so the strong reference will only be kept alive until the executions has finished. But just to make sure and not unnecessarily execute it I have added the weak reference now :)

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
nickguendling: This comment is also in regard to the strong self reference, right?

KaiTeuber: 👍

KaiTeuber (approved):

naveeddotio (approved):

PR #4047, opened 2022-01-05
fix/Crash getAppConfig

ArturFriesen:
## Description
This PR is an attempt to fix a crash while accessing the app configuration.
The fix is to dispatch the callback of the http client to the app config queue.
Also it changes the queue to a serial queue. Before that it was a concurrent queue called with a barrier, which is actually a serial queue.
I have also added a stress test, which concurrently fires the app config API.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11166?filter=-1


src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -157,72 +157,74 @@ final class CachedAppConfiguration: AppConfigurationProviding {
 
 	private func getAppConfig(with etag: String? = nil) -> Future<AppConfigResponse, Never> {
 		return Future { promise in
-			Self.queue.sync(flags: .barrier) {
+			Self.queue.sync {
```
KaiTeuber, line 160: Will this 'Return immediately...' case still happen if a serial queue is used now?
Is the property `requestIsRunning` still used, I can't see any change?

ArturFriesen: `requestIsRunning` is still used and will return early and save the promise for later fulfillment.
I think `requestIsRunning` is still needed with the serial queue and even if its called in sync. Because the call to `fetchAppConfiguration` is escaping. The serial queue can be empty while `fetchAppConfiguration` is still executing.

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Client/__tests__/Mocks/CachingHTTPClientMock.swift
```diff
@@ -68,12 +68,21 @@ final class CachingHTTPClientMock: CachingHTTPClient {
 	var onFetchAppConfiguration: ((String?, @escaping CachingHTTPClient.AppConfigResultHandler) -> Void)?
 
 	override func fetchAppConfiguration(etag: String? = nil, completion: @escaping CachingHTTPClient.AppConfigResultHandler) {
+		let clientQueue = DispatchQueue(label: "ClientQueue", attributes: .concurrent)
+
 		guard let handler = self.onFetchAppConfiguration else {
 			let response = AppConfigurationFetchingResponse(CachingHTTPClientMock.staticAppConfig, "fake")
-			completion((.success(response), nil))
+			// Dispatch the completion call to simulate URLSession calling back on another thread.
+			clientQueue.async {
```
marcussc, line 76: Will the `clientQueue` still be in scope at this point? 🤔

marcussc (approved):

PR #4050, opened 2022-01-06
Fix/7265 app shows alert

KaiTeuber:
## Description
I was not able to reproduce the error as described in the ticket. After investigating why that alert appears, I found a situation that might lead to the problem. `Fake` and `noResponse` server answers got handled as an error - but in this case we might ignore them. By use of the network link conditioner with profile `100% Loss` I saw the alert happened regularly.

My theorie about it: 
If the app goes into background mode, while the request is still running the probability increases that no response might happen. Same might be an issue in background mode. The alert gets trigged and the user will see it the next time the app opens from background.

I skip the alter on `noRespone` and `Fake` by now and can reproduce with the `100% Loss` profile that it won't show up.

## Notice
This is draft for the moment until I discussed this with a PO

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7265


src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -128,35 +128,27 @@ class HomeTableViewModel {
...
-					// Errors because of no registered corona tests or expired tests are ignored
-					break
-				case .responseFailure, .serviceError, .registrationTokenError, .unknownTestResult, .malformedDateOfBirthKey:
-					// Only show errors for corona tests that are still expecting their final test result
-					if self.coronaTestService.antigenTest != nil && self.coronaTestService.antigenTest?.finalTestResultReceivedDate == nil {
-						self.testResultLoadingError = error
+		CoronaTestType.allCases.forEach { coronaTestType in
+			Log.info("Updating result for test of type: \(coronaTestType)")
+			coronaTestService.updateTestResult(for: coronaTestType, force: true) { [weak self] result in
+				guard let self = self else { return }
```
pascal-brause, line 134:
```suggestion
				guard let self = self else {
				Log.error("Could not create strong self")
				 return 
				 }
```

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -128,35 +128,27 @@ class HomeTableViewModel {
...
+
+				if case .failure(let error) = result {
+					switch error {
+					case .noCoronaTestOfRequestedType, .noRegistrationToken, .testExpired:
+						// Errors because of no registered corona tests or expired tests are ignored
+						break
+					case .responseFailure(let responseFailure):
+						switch responseFailure {
+						case .fakeResponse:
+							Log.info("fake response - skip it's not an error")
```
pascal-brause, line 144: Perhaps increase from `info` to `warn`?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -128,35 +128,27 @@ class HomeTableViewModel {
...
+						case .noResponse:
+							Log.info("tried to get test result but no response was given")
```
pascal-brause, line 146: Perhaps increase from `info` to `warn`?

pascal-brause (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
KaiTeuber: basically we only log this for info purpose - I don't think this is a warning.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
KaiTeuber: we only log this for info purpose - I don't think this is a warning.

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -128,35 +128,30 @@ class HomeTableViewModel {
...
```
naveeddotio, line 149:
```suggestion
							Log.info("Tried to get test result but no response was received")
```

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -128,35 +128,30 @@ class HomeTableViewModel {
...
+
+				if case .failure(let error) = result {
+					switch error {
+					case .noCoronaTestOfRequestedType, .noRegistrationToken, .testExpired:
+						// Errors because of no registered corona tests or expired tests are ignored
+						break
+					case .responseFailure(let responseFailure):
+						switch responseFailure {
+						case .fakeResponse:
+							Log.info("fake response - skip it's not an error")
```
naveeddotio, line 147:
```suggestion
							Log.info("Fake response - skip it as it's not an error")
```

naveeddotio (approved):

Ein-Tim:
I think this PR broke localization of the messages that are shown if a test result couldn't be loaded, see: https://github.com/corona-warn-app/cwa-app-ios/issues/4180#issuecomment-1031795978

PR #4058, opened 2022-01-09
small localization fix

matthiasdiener:


Ein-Tim:
Same as #4035 which was closed.

marcussc:
Thanks, a translation delivery which will fix this Issue is expected to arrive today :)

PR #4062, opened 2022-01-10
Fix/11112 notification opens cwa home panel instead of test result screen

30mar:
## Description
Currently when the the User gets a notification that his test result has arrived it opens the homeViewController While it should have opened the testResultViewController.

The reasons and solutions:

1. The HomeCoordinator gets deallocated -> so we need to re-instantiate it.
2. the ParentViewController from which we are trying to present is **not** in the view Hierarchy -> so we to get the current visible ViewController from the UIWindow and do the presentation from it if it exists.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11112

## Video
https://user-images.githubusercontent.com/15270737/148738134-6e54d8bd-a287-44e2-b08a-b9ba8d478169.MP4


src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
```diff
@@ -159,13 +159,13 @@ class HomeCoordinator: RequiresAppDependencies {
 		})
 	}
 
-	func showTestResultFromNotification(with testType: CoronaTestType) {
+	func showTestResultFromNotification(with testType: CoronaTestType, visibleViewController: UIViewController?) {
 		if let presentedViewController = rootViewController.presentedViewController {
 			presentedViewController.dismiss(animated: true) {
-				self.showExposureSubmission(testType: testType)
+				self.showExposureSubmission(testType: testType, visibleViewController: visibleViewController)
```
nickguendling, line 165: Not sure if I agree with this approach to just present it on some other view controller, because this could even be the exposure submission flow already, couldn't it? So we would present it twice? And in line 164 we are already trying to dismiss any view controllers that are already presented, so the former "visible view controller" should actually be dismissed right before you are trying to present from it. So I'm wondering why this works at all and why it didn't work as expected before 😅

nickguendling (changes requested):

nickguendling (approved):

KaiTeuber (approved):

PR #4072, opened 2022-01-12
Extend decoding error with extra error to pass underlying error

pascal-brause:
## Description
Extends the .decoding error to pass the underlying error.

### DIFF
--- a/src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
@@ -43,7 +43,7 @@ extension Resource {
 /**
  The errors that can occur while handling resources
  */
-enum ResourceError: Error, Equatable {
+enum ResourceError: Error {
 	case missingData
 	case decoding(Error?)
 	case encoding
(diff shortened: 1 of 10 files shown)


src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
```diff
@@ -3,21 +3,21 @@
...
-A Resource is a composition of locator (where a resources can be found), service type to be used, data to send (sendResource) and data to receive (receiveResource).
-*/
+ A Resource is a composition of locator (where a resources can be found), service type to be used, data to send (sendResource) and data to receive (receiveResource).
+ */
 protocol Resource {
 	associatedtype Send: SendResource
 	associatedtype Receive: ReceiveResource
 	associatedtype CustomError: Error
-
+
```
KaiTeuber, line 12: why all those spaces?

KaiTeuber (approved):

naveeddotio:
@pascal-brause Unit tests are failing

marcussc (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
pascal-brause: 🤷‍♂️

src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
```diff
@@ -41,17 +41,64 @@ extension Resource {
...
+			return true
+		case (.undefined, _):
+			return false
+			
+		case (.missingEtag, .missingEtag):
+			return true
+		case (.missingEtag, _):
+			return false
+		}
+	}
```
nickguendling, line 101:
```suggestion
```

The manual implementation for enums with associated values should not be needed anymore since Swift 4.1, I think you can just remove this

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/_Resource.swift
pascal-brause: Yes, but only if i add the optionality and `.localizedDescription` in the `Equatable` implementation of `ServiceError`.

pascal-brause: But makes the code less complicated, so I can change it.

PR #4074, opened 2022-01-12
Fix A11y order in HealthCertificateQRCodeCell

marcussc:
## Description
Fix A11y order in HealthCertificateQRCodeCell

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9976

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Cells/HealthCertificateQRCodeCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Cells/HealthCertificateQRCodeCell.swift
@@ -44,7 +44,7 @@ class HealthCertificateQRCodeCell: UITableViewCell, ReuseIdentifierProviding {
 
 		titleLabel.text = cellViewModel.title
 		titleLabel.isHidden = cellViewModel.title == nil
-		
+
 		subtitleLabel.text = cellViewModel.subtitle
 		subtitleLabel.isHidden = cellViewModel.subtitle == nil
 		subtitleLabel.accessibilityLabel = cellViewModel.titleAccessibilityText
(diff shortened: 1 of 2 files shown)


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Cells/HealthCertificateQRCodeCell.swift
```diff
@@ -14,7 +14,7 @@ class HealthCertificateQRCodeCell: UITableViewCell, ReuseIdentifierProviding {
 		isAccessibilityElement = false
 
 		qrCodeView.isAccessibilityElement = true
-		titleLabel.isAccessibilityElement = true
+		titleLabel.isAccessibilityElement = false // https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9976
```
nickguendling, line 17: Another technique that I often use when the elements are out of order is to set the `accessibilityElements` array in the correct order. Then the correct elements would be still highlighted when read out loud

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Cells/HealthCertificateQRCodeCell.swift
marcussc: Yes I tried that but this is utterly complicated since this View as so many subviews for different use cases, that it seemed better to just take out this element completely. Since the recommendation from the a11y expert was to have it read it out as one element and not two.

nickguendling (approved):

PR #4078, opened 2022-01-13
Story/10125 recycle bin show deletion date

naveeddotio:
## Description
- Adding deletion date on items in recycle bin
- Change in text
- Change in font colour (as per figma) 
- Changes to unit tests

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10125

## Screenshots
WhatsApp Image 2022-01-13 at 14 24 53


KathrinLuecke (approved):
UA approved

naveeddotio:
> When I compare the screenshot with Figma it seems like the distance between new text and the old info is a little to big.

@marcussc That is fixed already in the code.

naveeddotio:
> When I compare the screenshot with Figma it seems like the distance between new text and the old info is a little to big.

@marcussc Fixed and updated the screenshot.

src/xcode/ENA/ENA/Source/Scenes/RecycleBin/Cells/RecycleBinItemCellModel.swift
```diff
@@ -123,6 +123,18 @@ class RecycleBinItemCellModel {
 		}
 	}()
 
+	lazy var quaternaryInfo: String? = {
+		guard let expirationDate = Calendar.current.date(byAdding: .day, value: 30, to: recycleBinItem.recycledAt) else {
```
nickguendling, line 127: Could we make the `30` a static let on `RecycleBin` or something like that so we have a single place to change that value if we ever decide to change it?

nickguendling (changes requested):

marcussc (approved):

nickguendling (approved):

PR #4086, opened 2022-01-14
Support new booster notation

30mar:
## Description
Support new booster notation

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11276

## Screenshots
IMG_0166
IMG_0167


Ein-Tim:
Will you really show "Vaccination 3 of 2"?
Wouldn't it make more sense vaccination "3 of 2 (Booster)"? 
Or respective "2 of 1 (Booster)"?

Or can't you do this because of some EU requirements?

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificateArray+MostRelevant.swift
```diff
@@ -42,18 +42,34 @@ extension Array where Element == HealthCertificate {
...
 				return false
 			}
-
+			
 			return vaccinationEntry.doseNumber >= vaccinationEntry.totalSeriesOfDoses && (
 				$0.ageInDays ?? 0 > 14 ||
 				vaccinationEntry.isBoosterVaccination ||
-				vaccinationEntry.isRecoveredVaccination)
+				vaccinationEntry.isRecoveredVaccination ||
+				vaccinationEntry.doseNumber > vaccinationEntry.totalSeriesOfDoses
```
nickguendling, line 55: I think this logic should be included in the `isBoosterVaccination` property, because other places have to be able to rely on it detecting booster vaccinations as well.

And it would be great if you can add the unit tests to the target again for this PR already and update them, they should definitely be included in the target. It should be enough to extend the `vaccinationCertificate(type: .booster, ...)` helper to cover boosters with dn > sd.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/HealthCertificateArray+MostRelevant.swift
```diff
@@ -42,18 +42,34 @@ extension Array where Element == HealthCertificate {
...
+		
+		// 2- check if there is multiple complete vaccination certificates that have the same ageInDays
+		let completeVaccinationCertificatesWithTheSameAgeInDays = completeVaccinationCertificates.filter {
+			$0.ageInDays == latestCompleteVaccinationCertificateAgeInDays
 		}
+		
+		// 3- return the certificate with the latest issuedAt date
+		return completeVaccinationCertificatesWithTheSameAgeInDays.sorted(by: {
+			$0.cborWebTokenHeader.issuedAt < $1.cborWebTokenHeader.issuedAt
+		}).max()
```
nickguendling, line 70: Instead of checking for same-day vaccinations I would propose to enhance the `Comparable` implementation of the `HealthCertificate` class. You can introduce a secondary order by using tuples like this: `(lhsDate, lhs.cborWebTokenHeader.issuedAt) < (rhsDate, lhs.cborWebTokenHeader.issuedAt)`. We are already sorting certificates when the are added to a person, so with this enhanced implementation the should already be in the right order when the most relevant certificate is determined in this extension.

With this approach it would also be preferable to have unit tests in the tests for the `HealthCertificate` class. I think there should already be some that test the sorting that you can maybe adapt to check for same-day or even same-time certificates.

nickguendling (changes requested):

mlenkeit:
> Will you really show "Vaccination 3 of 2"? Wouldn't it make more sense vaccination "3 of 2 (Booster)"? Or respective "2 of 1 (Booster)"?
> 
> Or can't you do this because of some EU requirements?

@Ein-Tim  _2 of 1_ is the result of an agreement on EU level for boosters after vaccinations that require a single dose (such as Janssen). Some countries may choose _3 of 2_ for boosters after vaccinations that require two doses. The PR is adding support for these notations.

Ein-Tim:
@mlenkeit Okay, agreed.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertificate.swift
```diff
@@ -90,9 +90,12 @@ final class HealthCertificate: Codable, Equatable, Comparable, RecycleBinIdentif
 
 	static func < (lhs: HealthCertificate, rhs: HealthCertificate) -> Bool {
 		if let lhsDate = lhs.sortDate, let rhsDate = rhs.sortDate {
-			return lhsDate < rhsDate
+			if lhsDate == rhsDate {
+				return lhs.cborWebTokenHeader.issuedAt < rhs.cborWebTokenHeader.issuedAt
+			} else {
+				return lhsDate < rhsDate
+			}
```
nickguendling, line 97:
```suggestion
			return (lhsDate, lhs.cborWebTokenHeader.issuedAt) < (rhsDate, rhs.cborWebTokenHeader.issuedAt)
```

This should work and looks a bit more elegant. Sorry, had a typo in it in my earlier comment (`lhs` in the right hand tuple), that's why it probably didn't work if you tried it out already 😅

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/HealthCertificate.swift
30mar: just tried it, it woks, nice!

nickguendling (approved):

marcussc (approved):

PR #4105, opened 2022-01-20
Feature/11355 business configuration dcc prioritization/11466 faq anchor url

naveeddotio:
## Description
Resolve an FAQ Anchor to a URL string

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11466


CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/Workers/__tests__/LinkHelperTests.swift
```diff
@@ -0,0 +1,33 @@
...
+	}
+	
+	func testFAQAnchorOtherLanguages() throws {
+		let supportedLocale = ["tr", "ro", "pl"]
+		if supportedLocale.contains(NSLocale.current.languageCode ?? "") {
+			let url = LinkHelper.urlString(suffix: "test", type: .faq)
+
+			XCTAssertEqual(url, "https://www.coronawarn.app/en/faq/#test")
+		}
+	}
```
nickguendling, line 32: I think this structure will not work, as the first test would fail if the current `languageCode` is not "de". I think it will not fail for us because we only test unit tests in German. But with that in mind we will never test anything in the other two tests as the current language will always be German.

Idea one: Figure out a way to override the language settings in single tests.
Idea two: Don't use the `Localizable.links.strings` files, instead hardcode the URL in `urlString(suffix: String, type: LinkType)`, but change the signature to `urlString(suffix: String, type: LinkType, languageCode: String? = Locale.current.languageCode)` so you can pass it in for the tests and don't have to pass it in anywhere else.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Workers/__tests__/LinkHelperTests.swift
naveeddotio: I already search if we have any mechanism to override locale for a particular unit test but unfortunately we don't have one, the unit tests are run with one locale. Have refactored it to option 2 which you have suggested.

nickguendling (approved):

KaiTeuber (approved):

Issue #4076, opened 2022-01-13
App crashes when scanning a certificate


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone XR
- iOS version: 15.2
- App version: 2.15.1

## Describe the bug

The Corona-Warn-App for iOS crashes when scanning a specific certificate. Under Android, the app does not crash.

## Steps to reproduce the issue

1. Open the universal QR code scanner
2. Scan the certificate
3. The app crashes

## Expected behaviour

The app should import the certificate and not crash.

## Additional context

I can provide the certificate via mail to you (@thomasaugsten).

---
Internal Tracking ID: [EXPOSUREAPP-11450](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11450)

Ein-Tim:
The crash is still reproducible on version 2.16.2.

Under Android the certificate is still accepted.

I would appreciate if someone could take a look at this, to be sure that this is an edge case and does not happen to more users (@thomasaugsten)

@dsarkar I guess it would be good to get this mirrored to JIRA.

thomasaugsten:
Yes please provide the certificate

Ein-Tim:
@thomasaugsten Your e-mail is thomasaugsten@sap.com, right?

larswmh:
Thanks for your report @Ein-Tim. Internal Tracking ID: [EXPOSUREAPP-11450](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11450)

---
Corona-Warn-App Open Source Team

thomasaugsten:
thomas.augsten@sap.com

Ein-Tim:
@thomasaugsten I send it to thomas.augsten@sap.com.

larswmh:
@Ein-Tim FYI PR #4107

Ein-Tim:
Thanks @larswmh!

Seems like this will be included in a hot fix! 

I was able to confirm that this PR fixes the issue.

dsarkar:
@Ein-Tim Thanks for the feedback!

Ein-Tim:
[...]

PR #4107, opened 2022-01-20
Use custom uncompression logic.

ArturFriesen:
## Description
Use custom decompression logic to fix crash.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11450


marcussc (approved):

marcussc:
@Ein-Tim just out curiosity from where did you get a QR-Code that would crash the old implementation?

Ein-Tim:
@marcussc An user on twitter send me the QR code via DM (reporting that this QR code would crash the app), I then created https://github.com/corona-warn-app/cwa-app-ios/issues/4076, today I send this QR code to @thomasaugsten via mail and now we're here 😁

naveeddotio (approved):

marcussc:
@Ein-Tim Thanks! Was hoping you had an other source :D One that we could use in our Uni-Tests. Due to Privacy Protection we can not use the Code in our Unit-Tests since everybody could "see" it then.

PR #4110, opened 2022-01-21
Feature/11471 allow sync access to cache o resource

KaiTeuber:
## Description
Added a possibility to RestServiceProvider to access ReceiveModel from Cache - if available

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11471

### DIFF
--- a/src/xcode/ENA/ENA/Source/HTTPClient/Service/RestServiceProvider.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/Service/RestServiceProvider.swift
@@ -69,6 +69,19 @@ class RestServiceProvider: RestServiceProviding {
 		}
 	}
 
+	func cached<R>(
+		_ resource: R,
+		_ completion: @escaping (Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void
+	) where R: Resource {
+		switch resource.type {
+		case .caching:
+			cachedRestService.cached(resource, completion)
+		default:
+			Log.error("Cache is not supported by that type of restService")
+			completion(.failure(.resourceError(.missingCache)))
+		}
+	}
+
 	// update evaluation trust - only possible for dynamic pinning at the moment
 	func update(_ evaluateTrust: EvaluateTrust) {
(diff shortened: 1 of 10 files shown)


pascal-brause (approved):

marcussc (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Service/RestServiceProvider.swift
```diff
@@ -69,6 +69,19 @@ class RestServiceProvider: RestServiceProviding {
 		}
 	}
 
+	func cached<R>(
+		_ resource: R,
+		_ completion: @escaping (Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void
+	) where R: Resource {
```
nickguendling, line 75: I'm wondering why this has an asynchronous API and if it needs to have that

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Service/RestServiceProvider.swift
KaiTeuber: This is because we use it for the async load function to pass in the completion closure. I'll have a look if we can change that a bit.

src/xcode/ENA/ENA/Source/HTTPClient/Service/__tests__/CachedRestServiceTests.swift
```diff
@@ -58,7 +58,7 @@ class CachedRestServiceTests: XCTestCase {
 		let cache = KeyValueCacheFake()
 		cache[locator.hashValue] = CacheData(data: dummyData, eTag: eTag, date: Date())
 
-		// Return nil and http code 304. In this case the service should load the reault from the cache.
+		// Return nil and http code 304. In this case the service should (load) the reault from the cache.
```
nickguendling, line 61: Not sure what this change is about 😅 

But found a typo
```suggestion
		// Return nil and http code 304. In this case the service should load the result from the cache.
```

nickguendling (approved):

PR #4113, opened 2022-01-22
Use multi-line comments as multi-line comments in Strings files

Jeehut:
## Description
This solves future potential issues with translation management tools that will only keep a single "file header comment". Currently, the multi-line style `/* comment */` was used, but instead of putting all header comments into one, a new comment was opened on each line. This PR merges these into one comment to be in line with Xcode default generated "header comment" style format.

## Link to Jira
N/A (this is such a small change, don't worth the overhead of reporting to me)

## Screenshots
Nothing changed in the apps UI.


marcussc:
@Jeehut Thanks for this fix, sadly the currently used translation tool will override this changes with the next translation delivery.

PR #4118, opened 2022-01-24
Feature/11464 information about caching or not

pascal-brause:
## Description
Introduces a new protocol `MetaDataProviding` which can be implemented by a model. This adds then to the model metadata like `isLoadedFromCache` or `headers`.
With this, i removed the `ModelWithHeaders` and `WithHeaderReceiveResource`, added a new receive model `TicketValidationAccessTokenReceiveModel` and adopted the refactoring to the TicketValidation processings and their tests.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11464


src/xcode/ENA/ENA/Source/HTTPClient/Service/Models/Receive/ModelWithCache.swift
```diff
@@ -0,0 +1,50 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+protocol ModelWithCaching {
```
ArturFriesen, line 5:
```suggestion
protocol ModelWithMetadata {
```

src/xcode/ENA/ENA/Source/HTTPClient/Service/Models/Receive/ModelWithCache.swift
```diff
@@ -0,0 +1,50 @@
...
+	var isCached: Bool { get set }
```
ArturFriesen, line 6:
```suggestion
	var loadedFromCache: Bool { get set }
```

src/xcode/ENA/ENA/Source/HTTPClient/Service/Models/Receive/ModelWithCache.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ Generel model to encapsulate the real model, which has to fullfill CBORDecoding. The main purpose of this model is to store a property if the real model returned from the service is from the cache or freshly fetched. Normally you cannot differentiate here.
+ */
+struct ModelWithCache<WrappedModel>: CBORDecoding, ModelWithCaching where WrappedModel: CBORDecoding {
```
ArturFriesen, line 12: This solution would only works for CBORResources? Would it be possible to have a general approach?

ArturFriesen, line 12: Why are there different protocols 'CBORDecoding' and 'ReceiveResource' ? They look quite the same. Could this be refactored to be simpler? Why isn't the 'ReceiveResource' also returning 'ModelDecodingError'? Would be more precise IMHO.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Service/Models/Receive/ModelWithCache.swift
pascal-brause: I renamed the `CBORDecoding` to `CBORDecodable` and it's func to `make(with data: Data)` to clarify that `CBORDecodable` is for models and `ReceiveResource` for resources. This is the generic approach to map the generic resource onto the specific one (JSON, ProtoBuf, CBOR etc.)
`CBORDecodable` does not only decode the CBOR but also creates the model based on the CBOR.

At the end the protocols seems very similar but are for different things.

pascal-brause: Yeah, good idea, but this would lead into an huge refactoring because the resource's decoding work not all the same way (JSON and Protobuf offer an decoding API, CBOR not). We would need to change our architecture here.
So i would like to let it as it is.

ArturFriesen: And what if we would resolve this at a higher level. For example, would it be possible to have a func 'loadWithMetadata(...)' Which returns a touple, metadata and the model?

pascal-brause: I will have a look.

pascal-brause: Okay, on the level as new API call it is not as easy as it seems. But I refactored on the model layer some things.

ArturFriesen (approved):
💫

KaiTeuber (approved):

PR #4119, opened 2022-01-24
Show 2G+ label for booster vaccinations.

CollinAlpert:
## Description
Show the "2G+" label for booster vaccinations.

Referenced issues: 
https://github.com/corona-warn-app/cwa-documentation/issues/808

## JIRA
[EXPOSUREAPP-11401](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11401)

## Screenshots

Screen Shot 2022-01-24 at 19 19 38
Screen Shot 2022-01-24 at 19 19 56
Screen Shot 2022-01-24 at 19 20 28
Screen Shot 2022-01-24 at 19 20 16


thomasaugsten:
Thanks for your effort here but at the moment it is not possible to determin "2G+ + B" for all certificate combination. 
We need here an approach which is working for all certificates and both OS as well the possibility to check the information in the CovPassCheck.

CollinAlpert:
This is not a solution for _all_ certificate combinations. I only added a case for booster vaccinations, so those would be displayed as 2G+. It is a step in the right direction, I believe. I can imagine that it will be quite difficult to add cases for all combination.

thomasaugsten:
We working in parallel on approach for all certificate there is no effort needed for a partial solution.
Only a complete solution over all certificates and apps will help here to prevent confusion of the user.

Thanks for you contribution but I will close the the PR in favor of a complete solution.

Ein-Tim:
@thomasaugsten 

But wouldn't be better to at least show 2G+ when we  an detect it instead of not showing it at all?

However, as the regulations differ from federal state to federal state, we should make sure to nowhere show 2G+ although the person does not fulfill the 2G+ rules.

thomasaugsten:
This had the Disadvantaged for boostered user where we cannot detect this.
If the status is not checkable we open the possibility for fraud.
As you mentioned the simple booster case differs also in federal states which again make the status incorrect

Ein-Tim:
> This had the disadvantaged for boostered user which we cannot detect.

Yeah but at the moment **all** boostered use have this disadvantage.

CollinAlpert:
I agree. When your certificate is scanned with the CovPassCheck app, it does not show the number of vaccinations, so if you are boostered, you are forced to show details of your vaccination.

xrad:
Thanks for referencing my PR @Ein-Tim, which is intended as a suggestion. My reasoning was simply: if I have a valid 3/3 certificate, it must be a booster on top of a 2/2 series. Is this not a correct assumption?

Ein-Tim:
@xrad 

It is a booster, but not in all federal states a booster always guarantees you access to 2G+ events without a test (e.g. in some federal states a booster guarantees you access to 2G+ events only if the vaccination was less than 90 days ago, in others less than 180 days, etc.). 

If we implement something which just checks if a 3/3 certificate is stored and if yes we show 2G+, there is the risk that we show a wrong status in some federal states and users are thinking they are allowed to visit 2G+ events, however, at the event they are turned away because e.g. their vaccination was more than 90 days ago.

Also, being boostered is not sufficent to visit e.g. a nursing home, there even boostered people need a test.

This is a very complex topic and the team is working on a rule engine which covers all of this: https://github.com/corona-warn-app/cwa-app-ccl

xrad:
I see. It seems the term "2G+" is a bit overloaded and a simple badge probably won't cut it. 
Definitely a good idea to factor out the logic to a common module.

PR #4120, opened 2022-01-25
Feature/reordering new http client

pascal-brause:
## Description
Brings new order to the hole new HTTPClient folder structure. Moving and renaming a lot of stuff. 
To test, check out the branch and have look in the project navigator.


KaiTeuber (approved):

src/xcode/ENA/ENA.xcodeproj/project.pbxproj
```diff
@@ -30,7 +30,6 @@
 		010A47562754C1BC000AA80E /* VersionError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 010A47552754C1BC000AA80E /* VersionError.swift */; };
 		010A47662754C74F000AA80E /* TicketValidationResultTokenResource.swift in Sources */ = {isa = PBXBuildFile; fileRef = 010A47652754C74F000AA80E /* TicketValidationResultTokenResource.swift */; };
 		010A47682754C7E7000AA80E /* TicketValidationResultTokenSendModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 010A47672754C7E7000AA80E /* TicketValidationResultTokenSendModel.swift */; };
-		010A476A2754D5F6000AA80E /* TicketValidationResultTokenLocator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 010A47692754D5F6000AA80E /* TicketValidationResultTokenLocator.swift */; };
```
ArturFriesen, line 33: The comment is not regarding this line, but in general, I think the folders HTTPClient/Resources and HTTPClient/Service/LocatorsNotYetUsed and parts of HTTPClient/Security shouldn't be part of the HTTPClient folder. These files are containing business data and are not belonging to the core component implementation.

src/xcode/ENA/ENA/Source/HTTPClient/Resources/TeleTan/Locator+TeleTan.swift
```diff
@@ -10,7 +10,7 @@ extension Locator {
 	// receive:	JSON
 	// type:	default
 	// comment:
-	static func registrationToken(
+	static func teleTan(
```
ArturFriesen, line 13: How was the decision about that naming? The path says "registrationToken"...

src/xcode/ENA/ENA/Source/HTTPClient/Resources/TeleTan/TeleTanSendModel.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-struct KeyModel: PaddingResource {
+struct TeleTanSendModel: PaddingResource {
```
ArturFriesen, line 7: Same question here, why is it named TeleTan? The key type can also be qrCode or guid

src/xcode/ENA/ENA/Source/HTTPClient/Resources/RegistrationToken/RegistrationReceiveModel.swift
```diff
@@ -2,7 +2,7 @@
 // 🦠 Corona-Warn-App
 //
 
-struct SubmissionTANModel: Codable {
+struct RegistrationReceiveModel: Codable {
```
ArturFriesen, line 5: Why is it called 'Registration' when its 'submissionTAN' inside?

ArturFriesen (changes requested):

src/xcode/ENA/ENA.xcodeproj/project.pbxproj
pascal-brause: So you propose to move the resources from `HTTPClient/Resources` to the location where they are used?
And what about the `LocatorsNotYetUsed`? Where to move them?

What about renaming `HTTPClient` to `HTTPClientCore` and then creating on the same level a folder called `HTTPClientResources`?

So at the end we can create a own package for the `HTTPClientCore`....

pascal-brause: Okay an own package is not as easy because we have too many dependencies. This would better be an separate ticket because it takes too much time.
But what I did is rename the `HTTPClient` folder to `HTTPClientCore` and create a new folder `HTTPClient` where all the custom resources and the LocatorsNotYetUsed are contained in.
The security stays in `HTTPClientCore`, because it is needed for the functionality of the `RestServiceProvider` itself.

src/xcode/ENA/ENA/Source/HTTPClient/Resources/TeleTan/Locator+TeleTan.swift
pascal-brause: Good catch! There was a renaming failure. registrationToken and teleTan got swapped. I corrected this.

src/xcode/ENA/ENA/Source/HTTPClient/Resources/TeleTan/TeleTanSendModel.swift
pascal-brause: Good catch! There was a renaming failure. registrationToken and teleTan got swapped. I corrected this.

src/xcode/ENA/ENA/Source/HTTPClient/Resources/RegistrationToken/RegistrationReceiveModel.swift
pascal-brause: Good catch! There was a renaming failure. registrationToken and teleTan got swapped. I corrected this.

src/xcode/ENA/ENA/Source/HTTPClient/Resources/TeleTan/TeleTanSendModel.swift
pascal-brause: That is right, but the KeyModel was only used in combination with a TeleTanResource. So I thought i change the name to be consistent.
Otherwise, i could revert the naming here and move the model out of the folder TeleTan.
@ArturFriesen what do you think?

ArturFriesen (approved):

PR #4122, opened 2022-01-25
Feature/integrate funktions package

ArturFriesen:
## Description
Integrates the json functions package

### DIFF
--- a/src/xcode/ENA.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/src/xcode/ENA.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -95,9 +95,9 @@
         "package": "jsonfunctions",
         "repositoryURL": "https://github.com/corona-warn-app/json-functions-swift",
         "state": {
-          "branch": "main",
-          "revision": "c7e764f670212cda736dccdb8c22463baf95428b",
-          "version": null
+          "branch": null,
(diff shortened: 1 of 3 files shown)


marcussc (approved):

src/xcode/ENA.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,6 +1,15 @@
...
     "pins": [
+      {
+        "package": "AnyCodable",
+        "repositoryURL": "https://github.com/Flight-School/AnyCodable",
+        "state": {
+          "branch": null,
+          "revision": "b1a7a8a6186f2fcb28f7bda67cfc545de48b3c80",
+          "version": "0.6.2"
+        }
+      },
```
nickguendling, line 12: I also added this on my branch, is this required to import our package and automatically added?

And I think we need to add the legal reference in the `licenses.json` file, don't we? And to generate the model for the licenses screen?

ArturFriesen: I think this is added because jsonfunctions is using AnyCodable. At that point we don't need to add the license.

ArturFriesen: In other words, we don't need to add licenses for implicit dependencies.

nickguendling (approved):

PR #4128, opened 2022-01-26
fix/11579-missing-error-when-flightmode

pascal-brause:
## Description
Error was not called due to missing completion and additionally the error code got lost due to double custom error casting.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11579

## Screenshots
11579-light


marcussc (approved):

KaiTeuber (changes requested):
I don't see why not return a customError any more.
The missing completion call before return I did unterstood.

pascal-brause:
> I don't see why not return a customError any more.
> The missing completion call before return I did unterstood.

Because we try to cast the error here a second time to a custom error. It is already done inside the service and we receive already the customError packed into a .receiveResourceError

KaiTeuber (approved):

PR #4145, opened 2022-01-28
Feature/11580 dev menu ccl config

pascal-brause:
## Description
Add a new dev menu toggle to force CCLConfiguration update.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11595

## Screenshots
Simulator Screen Shot - iPhone 12 - 2022-01-31 at 16 47 54


src/xcode/ENA/ENA/Source/Developer Menu/Features/DMCCLConfiguration/DMCCLConfigurationViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+
+		switch section {
+			
+		case .getCall:
+			return DMButtonCellViewModel(
+				text: "Call GET CLL Configuration",
+				textColor: .white,
+				backgroundColor: .enaColor(for: .buttonPrimary),
+				action: { [weak self] in
+					self?.restService.load(CCLConfigurationResource()) { result in
```
nickguendling, line 51: Don't we need to do this through the CCLService so the update would also have an effect on the data in the app?

pascal-brause: if the CCLService holds the data, yeah, you are right. I will left this PR as draft so you can extend it as soon the CCLService is done.

marcussc (approved):

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMCCLConfiguration/DMCCLConfigurationViewModel.swift
ArturFriesen: @nickguendling I think this topic can be closed, because setting the flag does not load any data. Instead the tester needs to restart the app.

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMCCLConfiguration/DMCCLConfigurationViewModel.swift
```diff
@@ -0,0 +1,76 @@
...
+import CertLogic
+
+final class DMCCLConfigurationViewModel {
+
+	// MARK: - Init
+
+	init(
+		restServiceProvider: RestServiceProviding
+	) {
+		self.restService = restServiceProvider
```
nickguendling, line 18: I think the rest service is not needed anymore since we only set a flag now

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Developer Menu/Features/DMCCLConfiguration/DMCCLConfigurationViewModel.swift
ArturFriesen: Removed.

nickguendling (approved):

ArturFriesen (approved):

PR #4147, opened 2022-01-28
Call evaluate on instance property.

ArturFriesen:
## Description
Call json functions API.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11467


src/xcode/ENA/ENA/Source/Services/CCLService/CCLService.swift
```diff
@@ -61,7 +61,7 @@ class CLLService: CCLServable {
 	}
 	
 	func evaluateFunction<T>(name: String, parameters: [String: AnyDecodable]) throws -> T where T: Decodable {
-		return try JsonFunctions().evaluateFunction(name: name, parameters: parameters)
+		return try jsonFunctions.evaluateFunction(name: name, parameters: parameters)
```
nickguendling, line 64: I just realized this is specific for the UI texts, and maybe we should have a separate func for that, but according to the [tech spec](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/logic-engine/docs/spec/dcc-ccl.gen.md#formatting-a-ui-text---system-time-dependent) we need to pass in the default parameters (os, language, now) for these in addition to the parameters that are passed in. 

Maybe we can just extend this implementation to always pass in the default parameters as well, because currently we never need a function call without them.

ArturFriesen: Good point. I have changed the func name and added the default parameters.

src/xcode/ENA/ENA/Source/Services/CCLService/CCLService.swift
```diff
@@ -60,8 +60,10 @@ class CLLService: CCLServable {
 		}
 	}
 	
-	func evaluateFunction<T>(name: String, parameters: [String: AnyDecodable]) throws -> T where T: Decodable {
-		return try JsonFunctions().evaluateFunction(name: name, parameters: parameters)
+	func evaluateFunctionWithDefaultValues<T>(name: String, parameters: [String: AnyDecodable]) throws -> T where T: Decodable {
+		let language = Locale.current.languageCode ?? "en"
```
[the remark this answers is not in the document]
ArturFriesen: Added default

nickguendling (approved):

naveeddotio (approved):

PR #4148, opened 2022-01-28
Feature/11457 localized strings wallet info

naveeddotio:
## Description
Adding support for localised strings based on DCCUIText

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11457
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11588


CV113 (approved):
UA approved

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,237 @@
...
+					}
+				}
+			}
+		}
+		
+		return nil
+	}
+
+	private func localizedSystemTimeDependentFormatText(languageCode: String?) -> String? {
+		let service = CCLService()
```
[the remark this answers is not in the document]
ArturFriesen: I think it is better to pass it into every model. I cannot see another solution without introducing tight coupling.

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+import AudioToolbox
```
ArturFriesen, line 6: What is this import for? 🎸

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
...
+	static let systemTimeDependent = "system-time-dependent"
+}
+
+struct DCCUITextParameter {
+	let type: String
+	let value: Any
+}
+
+public struct DCCUIText: Codable, Equatable {
+	let type: String
```
ArturFriesen, line 31: These properties should be unter mark internal.

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
...
+		if let parameters = parameters.value as? [[String: Any]] {
+			if parameters.isEmpty {
+				// text without parameters
+				if let textDescriptorQuantity = quantity {
+					return quantityBasedFormatText(formatText: formatText, quantity: textDescriptorQuantity)
+				}
+			} else {
+				var mappedParameters: [DCCUITextParameter] = []
+				
+				// we could get multiple parameters
```
ArturFriesen, line 113: This part is the same like in line 84. Could be extracted to make it better readable and maintainable.

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
...
+				for parameter in parameters {
+					guard let type = parameter["type"] as? String, let value = parameter["value"] else {
+						return nil
+					}
+					mappedParameters.append(DCCUITextParameter(type: type, value: value))
+				}
+				
+				// quantity shall be set to the value of textDescriptor.quantity
+				if let textDescriptorQuantity = quantity {
+					// text shall be determined by passing formatTexts and formatParameters to a quantity-depending printf-compatible format function by using quantity
```
ArturFriesen, line 123: Comments are misleading because of using naming "formatParameters" without having the same name in code.

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
...
+
+	static let dateFormatter: ISO8601DateFormatter = .iso8601DateFormatter()
+	static let localDateFormatter: DateFormatter = .localDateFormatter()
+	static let localDateTimeFormatter: DateFormatter = .localDateTimeFormatter()
+	static let outputDateFormatter: DateFormatter = .outputDateFormatter()
+	static let outputDateTimeFormatter: DateFormatter = .outputDateTimeFormatter()
+	
+	// MARK: - Internal
+
+	func localized(languageCode: String? = Locale.current.languageCode) -> String? {
```
ArturFriesen, line 46: Returning optional means, that there is no detailed error handling in place? What happens on the other side if nil is returned?

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
```diff
@@ -0,0 +1,273 @@
...
+
+public struct DCCUIText: Codable, Equatable {
+	let type: String
+	let quantity: Int?
+	let quantityParameterIndex: Int?
+	let functionName: String?
+	let localizedText: [String: AnyCodable]?
+	let parameters: AnyCodable
+
+	static let dateFormatter: ISO8601DateFormatter = .iso8601DateFormatter()
```
ArturFriesen, line 38: I guess you are using these date formatters for similar strings like I was using in the other task. We should consider to merge them.

ArturFriesen (changes requested):

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
naveeddotio: Output of all four formatters are different as that of your formatters, so I would like to stick with the ones used here. 

Unit tests:

XCTAssertEqual failed: ("Optional("Gültig ab 2022-01-01")") is not equal to ("Optional("Gültig ab 01.01.22")") - Failing Test: placeholder of type utcDate

XCTAssertEqual failed: ("Optional("Gültig ab 2022-01-01T23:30:00Z")") is not equal to ("Optional("Gültig ab 01.01.22, 23:30")") - Failing Test: placeholder of type utcDateTime

XCTAssertEqual failed: ("Optional("Gültig ab 2022-01-02")") is not equal to ("Optional("Gültig ab 02.01.22")") - Failing Test: placeholder of type localDate

XCTAssertEqual failed: ("Optional("Gültig ab 2022-01-02T00:30:00+01:00")") is not equal to ("Optional("Gültig ab 02.01.22, 00:30")") - Failing Test: placeholder of type localDateTime

nickguendling (approved):

src/xcode/ENA/ENA/Source/Services/CCLService/DCCUIText.swift
ArturFriesen: Would it be possible to return an empty string instead of an optional?
This way we don't need to unwrap at every place we are using this API.

ArturFriesen (approved):

PR #4160, opened 2022-02-01
set the badge if the risk level changes regardless of the level

30mar:
## Description
set the badge if the risk level changes regardless of the level

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11545


KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
```diff
@@ -503,10 +503,9 @@ class HomeCoordinator: RequiresAppDependencies {
...
 				guard case let .risk(risk) = riskState,
 					  risk.riskLevelHasChanged,
-					  risk.level == .high,
 					  self?.rootViewController.tabBarController?.selectedViewController != self?.rootViewController
 				else {
-					Log.info("wrong risk level or home screen tab is active - skipped to set tab bar badge")
+					Log.info("home screen tab is active - skipped to set tab bar badge")
 					return
 				}
 				self?.badgeWrapper.update(.riskStateIncreased, value: 1)
```
nickguendling, line 511:
```suggestion
				self?.badgeWrapper.update(.riskStateChanged, value: 1)
```

nickguendling (changes requested):
I think we should reflect the change in behaviour in the name and comment

nickguendling (approved):

PR #4170, opened 2022-02-03
Fix/11625 fix test complete vaccination admission state()

30mar:
## Description
New fix for the test without fakes

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11625

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/AdmissionState/AdmissionStateTableViewCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/AdmissionState/AdmissionStateTableViewCell.swift
@@ -35,7 +35,6 @@ class AdmissionStateTableViewCell: UITableViewCell, UITextViewDelegate, ReuseIde
 	func configure(with cellModel: AdmissionStateCellModel) {
 		titleLabel.text = cellModel.title
 		titleLabel.isHidden = (cellModel.title ?? "").isEmpty
-		titleLabel.accessibilityIdentifier = AccessibilityIdentifiers.HealthCertificate.AdmissionState.title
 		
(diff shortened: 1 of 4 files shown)


src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/AdmissionState/AdmissionStateTableViewCell.swift
```diff
@@ -47,6 +47,7 @@ class AdmissionStateTableViewCell: UITableViewCell, UITextViewDelegate, ReuseIde
 
 		faqLinkTextView.attributedText = cellModel.faqLink
 		faqLinkTextView.isHidden = (cellModel.faqLink?.string ?? "").isEmpty
+		faqLinkTextView.accessibilityTraits = .button
 		faqLinkTextView.accessibilityIdentifier = AccessibilityIdentifiers.HealthCertificate.AdmissionState.faq
```
nickguendling, line 51: Totally makes sense to add the button trait 👍 Could you move it to the `private let faqLinkTextView` configuration as it is not depending on any information from the cellModel? Actually would also make sense for all the accessibility identifiers that we have here, now that I see this. 

And can you also add the trait on the `VaccinationState` cell? I think it would make sense that it is announced in VoiceOver as a button as well.

nickguendling (changes requested):

ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/AdmissionState/AdmissionStateTableViewCell.swift
30mar: it is fixed now

nickguendling (approved):

PR #4200, opened 2022-02-10
Fix flaky test

naveeddotio:
## Description
Fix flaky test - test_AddStatisticsButton_flow


marcussc (approved):

pascal-brause (approved):
So we do not test any longer if we can remove a statistic tile or do we check that in another test?

naveeddotio:
> So we do not test any longer if we can remove a statistic tile or do we check that in another test?

> So we do not test any longer if we can remove a statistic tile or do we check that in another test?

@pascal-brause @marcussc added another test to check the removal in another test using the launch argument which I created earlier. Please re-check

PR #4202, opened 2022-02-10
Fix/11772 priority fallback string

naveeddotio:
## Description
Fixing fallback for localised string and adding some unit test to verify

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11772


marcussc (approved):

mlenkeit:
I'll try to put the test case together...

naveeddotio:
> @naveeddotio I'm not sure the test case matches the bug. The test case is rather cover that if the language if e.g. `tr` and there's no text for `tr`, that `en` is used as a fallback instead of `de`.

@mlenkeit this is the exact test case which match what you have said:

{
 	  "description": "simple hello world - fallback en",
 	  "textDescriptor": {
 		"type": "string",
 		"localizedText": {
 		  "de": "Hallo Welt",
 		  "en": "Hello World"
 		},
 		"parameters": []
 	  },
 	  "assertions": [
 		{
 		  "languageCode": "en",
 		  "text": "Hello World"
 		}
 	  ]
 	},

naveeddotio:
> @naveeddotio I'm not sure the test case matches the bug. The test case is rather cover that if the language if e.g. `tr` and there's no text for `tr`, that `en` is used as a fallback instead of `de`.

Update: Discussed with Max and updated the test cases.

src/xcode/ENA/ENA/Source/Services/CCLService/__tests__/ccl-text-descriptor-test-cases.json
```diff
@@ -533,6 +533,39 @@
 		  "text": "vor 2 Minuten"
 		}
 	  ]
+	},
```
mlenkeit, line 536: Can you please update the entire file with the content from https://github.com/corona-warn-app/cwa-app-ccl/releases/download/latest/ccl-text-descriptor-test-cases.gen.json so that the date and the treeish in the header are updated, too?

mlenkeit: see https://github.com/corona-warn-app/cwa-app-ios/pull/4202/commits/9dd3dc32531080667839ff5dcc98f42c49065c7a

naveeddotio: 👍

30mar (approved):

PR #4214, opened 2022-02-14
unify days of installation calculation

KaiTeuber:
## Description
Unify calculation of days installed

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10329

## Screenshots
Screen Shot 2022-02-14 at 14 36 18
Screen Shot 2022-02-14 at 14 36 20
Screen Shot 2022-02-14 at 14 36 26

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
@@ -322,8 +322,8 @@ final class ExposureNotificationSettingViewController: UITableViewController, Ac
 			fatalError("Cell is not registered")
 		}
 
-		let appInstallationDate = store.appInstallationDate.map { Calendar.autoupdatingCurrent.startOfDay(for: $0).ageInDays ?? 0 } ?? 0
-		daysSinceInstallationCell.configure(daysSinceInstall: max(appInstallationDate, 0))
+		let daysSinceInstallation = store.appInstallationDate.map { Calendar.autoupdatingCurrent.startOfDay(for: $0).ageInDays ?? 0 } ?? 0
+		daysSinceInstallationCell.configure(daysSinceInstall: max(daysSinceInstallation, 0))
 		return daysSinceInstallationCell
 	}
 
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
```diff
@@ -318,12 +318,12 @@ final class ExposureNotificationSettingViewController: UITableViewController, Ac
 	}
 
 	private func daysSinceInstallationCell(for indexPath: IndexPath, in tableView: UITableView) -> UITableViewCell {
-		guard let daysSinceInstallationCell = tableView.dequeueReusableCell(withIdentifier: ReusableCellIdentifier.daysSinceInstallationCell.rawValue, for: indexPath) as? DaysSinceInstallTableViewCell else {
+		guard let daysSinceInstallationCell = tableView.dequeueReusableCell(withIdentifier: ReusableCellIdentifier.daysSinceInstallationCell.rawValue, for: indexPath) as? DaysSinceInstallTableViewCell,
+			  let appInstallationDate = store.appInstallationDate else {
 			fatalError("Cell is not registered")
```
marcussc, line 323:
```suggestion
			fatalError("Cell is not registered or no appInstallationDate")
```

nickguendling, line 323: I think we should not use a fatalError if the `appInstallationDate` is not set, maybe use a separate constant that contains something like `let installationAgeInDays = store.appInstallationDate.map { Calendar.autoupdatingCurrent.startOfDay(for: $0).ageInDays } ?? 0` and then pass this into the `max()` call below

marcussc (approved):

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ENSetting/ExposureNotificationSettingViewController.swift
KaiTeuber: I''ll try the suggested .map solution.

nickguendling (approved):

PR #4216, opened 2022-02-14
Error Message "notResponding" in Onboarding Process

KaiTeuber:
## Description
The error happens if navigation (forward & backward) is done inside the onboarding. This leads to multiple requests to activate ENF. To avoid the error we need to store the current activation state, so we can bypass it.

I added a new property to the store and even disabled navigation (back) until we get a result.
Otherwise we might miss the response.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11655

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
@@ -8,13 +8,6 @@ import UserNotifications
 import ExposureNotification
 import WebKit
 
-enum OnboardingActivationState: Codable {
-	case notActivated
-	case pending
-	case failed
-	case activated
-}
-
 // swiftlint:disable:next type_body_length
 final class OnboardingInfoViewController: UIViewController {
 
@@ -97,6 +90,7 @@ final class OnboardingInfoViewController: UIViewController {
 	private var client: Client
 	private var pageSetupDone = false
 	private var onboardingInfos = OnboardingInfo.testData()
+	private var exposureManagerActivated = false
 	private var subscriptions = [AnyCancellable]()
 
 	@IBAction private func didTapNextButton(_: Any) {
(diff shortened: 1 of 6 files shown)


src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -346,13 +353,27 @@ final class OnboardingInfoViewController: UIViewController {
...
-		guard !exposureManagerActivated else {
+		guard store.exposureActivationOnboardingState == .notActivated else {
 			completion?()
 			return
 		}
 
+		// we need to disable back navigation until we have an
+		// response for activation. Otherwise we can navigate back,
+		// kill the closure and never receive a result. When next
+		// activation request gets triggered an alter will appear
```
Ein-Tim, line 364: Do you mean "alter" or is this a typo and should be "altert"? :D

KaiTeuber: Thanks for the information, I fixed the typo

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -90,7 +97,7 @@ final class OnboardingInfoViewController: UIViewController {
 	private var client: Client
 	private var pageSetupDone = false
 	private var onboardingInfos = OnboardingInfo.testData()
-	private var exposureManagerActivated = false
+//	private var exposureManagerActivated = false
```
marcussc, line 100:
```suggestion
```

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
```diff
@@ -361,7 +382,7 @@ final class OnboardingInfoViewController: UIViewController {
 				persistForDPP(accepted: false)
 				completion?()
 			} else {
-				self.exposureManagerActivated = true
+				self.store.exposureActivationOnboardingState = .activated
```
marcussc, line 385: Hmm why not keep it as an instance variable but store it the store? 🤔
Keeping it as an instance variable would reset the flag after a restart of the app which would be desirable if ENF is acting up.

marcussc (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Onboarding/OnboardingInfoViewController.swift
KaiTeuber: ENF gets activated while onboarding gets show. But if you navigate back the view controller gets freed from memory. Navigating forward again will create a new instance, not knowing that ENF was activated already.
Maybe we need to exclude the flag from reset logic? But the behavior isn't different than before.

KaiTeuber: As discussed, I found a better solution. Please have a 2nd look.

PR #4223, opened 2022-02-15
Added texts for QR Code Scanner RAT Errors

KaiTeuber:
## Description
We missed some error texts that Android has.
Added a result type to CoronaTestQRCodeParser to communicate error up to the caller, so it can show.
Texts added as a fixed sting, they are not localized in Android either.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11735

## Screenshots
Screen Shot 2022-02-15 at 17 38 50


Ein-Tim (commented):
Remember you are showing this pop up to a user who is not a developer. A text what to do would certainly make sense.

@UA-Team

thomasaugsten:
This is not an error for the user. This an information/error code for the test provider without receiving the original qr code

Ein-Tim:
@thomasaugsten But in the end the user will see it, or?

thomasaugsten:
We don't want this we only want a error code for the test provider.

Ein-Tim:
Ok so this is mainly for testing with the Test Flight App and should **never** occur in PROD?

thomasaugsten:
This should never be visible productive because all test provider going through an onboarding process

marcussc (approved):

pascal-brause (approved):

naveeddotio (approved):

PR #4230, opened 2022-02-16
Fix/11802 make failed certificates removable

nickguendling:
## Description
Introduces a (on purpose very hidden) delete function on certificates for which the decoding failed and which cannot be recovered, because we e.g. enforce stricter rules for the "iat" field in the meantime. This delete alert is reached by tapping on the certificate 7 times.
We want this delete functionality to be hidden so it can't be triggered accidentally and to encourage users to contact us when they find themselves with certificates that cannot be decoded, so we can detect issues with our decoding process.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11802

## Screenshots
IMG_144B486948F8-1

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
@@ -604,6 +604,17 @@ class HealthCertificateService {
 		}
 	}
 
+	func remove(decodingFailedHealthCertificate: DecodingFailedHealthCertificate) {
+		healthCertifiedPersons
+			.first {
+				$0.decodingFailedHealthCertificates.contains(decodingFailedHealthCertificate)
+			}?
+			.decodingFailedHealthCertificates
+			.removeAll {
+				$0 == decodingFailedHealthCertificate
+			}
+	}
+
 	// MARK: - Private
 
 	private let store: HealthCertificateStoring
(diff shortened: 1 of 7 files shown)


KaiTeuber (approved):

KathrinLuecke (approved):
UA approved

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -604,6 +604,17 @@ class HealthCertificateService {
...
 
+	func remove(decodingFailedHealthCertificate: DecodingFailedHealthCertificate) {
+		healthCertifiedPersons
+			.first {
+				$0.decodingFailedHealthCertificates.contains(decodingFailedHealthCertificate)
+			}?
+			.decodingFailedHealthCertificates
+			.removeAll {
+				$0 == decodingFailedHealthCertificate
+			}
```
pascal-brause, line 615: should we check not better for the uniqueIdentifier of the certificate?

nickguendling: Equatable for `DecodingFailedHealthCertificate` compares the base45 string which is also unique among all certificates, as it literally is the whole certificate. And we can't compare the uniqueIdentifier in this case, as we would need to be able to decode the certificate first to get the identifier

pascal-brause (approved):

PR #4256, opened 2022-02-23
Feature/11897 dcc reissuance resource

ArturFriesen:
## Description
This PR introduces the rest service components needed for dcc reissuance.
It also contains a refactoring of the certificate pinning.
Instead of using different clients for pinning, we use different implementations which are bound to the Ressource.
This way we gain a lot more flexibility and, for example, a combination of caching and pinning would be possible.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11897


KaiTeuber (approved):
Very nice work 👍

pascal-brause (approved):

PR #4258, opened 2022-02-23
Feature/11900 update models and schedule notification

nickguendling:
## Description
Extends the data model and schedules notification for certificate reissuances.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11900


src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
```diff
@@ -112,6 +112,38 @@ class HealthCertificateNotificationService {
...
+		let name = person.name?.standardizedName
+		guard let newCertificateReissuance = person.dccWalletInfo?.certificateReissuance else {
+			Log.info("No certificate reissuance found for person \(private: String(describing: name))", log: .vaccination)
+			completion?()
+
+			return
+		}
+
+		if newCertificateReissuance != previousCertificateReissuance {
+			// we need to have an ID for the notification and since the certified person doesn't have this property "unlike the certificates" we will compute it as the hash of the string of the standardizedName + dateOfBirth
```
ArturFriesen, line 129: Would it be possible to introduce that identifier to the person? (could be the same, hash of name and dateOfBirth)

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
```diff
@@ -213,6 +245,46 @@ class HealthCertificateNotificationService {
...
+	}
+
+	private func scheduleCertificateReissuanceNotification(
+		personIdentifier: String,
+		completion: (() -> Void)? = nil
+	) {
+		Log.info("Schedule certificate reissuance notification for person with id: \(private: personIdentifier)", log: .vaccination)
+
+		let content = UNMutableNotificationContent()
+		content.title = AppStrings.LocalNotifications.certificateGenericTitle
```
ArturFriesen, line 276: Are the notifications looking the same or is it a mistake?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
nickguendling: Yes. I found four places where we calculated the hash independently, so I've added the identifier to the person now

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
nickguendling: Very good question! The tile was correct, but I got the message wrong, fixed it now :)

ArturFriesen (approved):

KaiTeuber (approved):

PR #4282, opened 2022-03-02
Feature/11911 merge configurations

nickguendling:
## Description
Merges CCL Configurations in case the default (bundled) configuration contains a newer version than the cached one.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11911


src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
```diff
@@ -15,10 +15,10 @@ protocol Resource {
 	var sendResource: Send { get }
 	var receiveResource: Receive { get }
 	
-	// Defines a default value for no network cases as the specific receive model (for resources like e.g. AppConfig, AllowList)
-	var defaultModel: Receive.ReceiveModel? { get }
-	
 	var trustEvaluation: TrustEvaluating { get }
+
+	// Defines a default value for no network cases as the specific receive model (for resources like e.g. AppConfig, AllowList)
+	func defaultModel() -> Receive.ReceiveModel?
```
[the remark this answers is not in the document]
nickguendling: Because I can't override it in `MockCCLConfigurationResource` if it is a computed property or a lazy var in the `CCLConfigurationResource`

src/xcode/ENA/ENA/Source/Services/CCLService/__tests__/CCLConfiguationTests.swift
```diff
@@ -171,7 +188,16 @@ class CCLServiceConfigurationTests: CCLServiceBaseTests {
 		)
 		let cache = try cache(with: Locator.CCLConfiguration(isFake: false), eTag: eTag, date: yesterday, responseData: oldCCLConfigurationData)
 		let restServiceProvider = RestServiceProvider(session: stack.urlSession, cache: cache)
-		let cclService = CCLService(restServiceProvider, appConfiguration: appConfiguration, cclServiceMode: [.configuration], signatureVerifier: MockVerifier())
+		let resource = MockCCLConfigurationResource(defaultModel: nil)
```
ArturFriesen, line 191: Why not use "CCLConfigurationResource" directly and overwriting the default model (property) after initialization?

src/xcode/ENA/ENA/Source/Services/CCLService/CCLService.swift
```diff
@@ -335,14 +325,38 @@ class CCLService: CCLServable {
...
+		jsonFunctions = JsonFunctions()
+
+		for configuration in newCCLConfigurations {
+			registerJsonFunctions(from: configuration)
+		}
+
+		var registeredConfigurations = newCCLConfigurations
+
+		/// Register functions from the default configurations as well, in case the default configurations contain (new) configurations not contained in the cached/fetched configurations
+		if let defaultConfigurations = cclConfigurationResource.defaultModel()?.cclConfigurations {
```
ArturFriesen, line 341: I think the CCLService should not have a dependency to CCLConfigurationResource.
Instead it should load the default on its own.

src/xcode/ENA/ENA/Source/Services/CCLService/__tests__/CCLConfiguationTests.swift
```diff
@@ -171,7 +188,16 @@ class CCLServiceConfigurationTests: CCLServiceBaseTests {
...
+
+		let cclService = CCLService(
+			restServiceProvider,
+			appConfiguration: appConfiguration,
+			cclServiceMode: [.configuration],
+			signatureVerifier: MockVerifier(),
+			cclConfigurationResource: resource
```
ArturFriesen, line 198: I think it would be good if we would have a component which loads the default. This component could be used in the resource and also be injected here.
This way we would not touch the service implementation. And avoid the cross dependency between the CCLService (CCL processing) and the CCLConfigurationResource (HTTP).

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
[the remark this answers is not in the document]
ArturFriesen: Didn't know that its not possible to override computed properties. Also didn't used inheritance for a long time 😅

src/xcode/ENA/ENA/Source/Services/CCLService/__tests__/CCLConfiguationTests.swift
nickguendling: Done that now 👍

ArturFriesen (approved):

naveeddotio (approved):

PR #4287, opened 2022-03-03
Add UI test for certificate replacement.

ArturFriesen:
## Description
Add UI-Test for certificate replacement.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12053


src/xcode/ENA/ENAUITests/HealthCertificate/ENAUITests_CertificateReplacement.swift
```diff
@@ -0,0 +1,51 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import XCTest
+
+class ENAUITests_CertificateReplacement: CWATestCase {
```
nickguendling, line 7: Maybe we should use the same naming here as well and the test plan could also be called "HealthCertificateReiisuance" or "ReissueHealthCertificate"
```suggestion
class ENAUITests_CertificateReissuance: CWATestCase {
```

nickguendling (approved):

marcussc (approved):

PR #4310, opened 2022-03-09
Task/12149 unit and UI tests for g status update

30mar:
## Description
Add Unit and UI tests for the G status update

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12149

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/__tests__/HealthCertificateServiceTests.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/__tests__/HealthCertificateServiceTests.swift
@@ -1750,12 +1750,13 @@ class HealthCertificateServiceTests: CWATestCase {
 		XCTAssertEqual(store.recycleBinItems.count, 1)
 	}
 	
-	func testDCCAdmissionStateChangge_Then_flagIsSetInHealthCertifiedPerson() throws {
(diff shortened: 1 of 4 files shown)


src/xcode/ENA/ENA/Source/Services/HealthCertificate/__tests__/HealthCertificateServiceTests.swift
```diff
@@ -1749,4 +1749,53 @@ class HealthCertificateServiceTests: CWATestCase {
...
+		var cclService = FakeCCLService()
+		cclService.dccWalletInfoResult = .success(newDCCWalletInfo)
+		cclService.didChange = true
+
+		let expectation = expectation(description: "dccWalletInfo updated")
+
+		let subscription = healthCertifiedPerson.$dccWalletInfo
+			.dropFirst()
+			.sink { _ in
+				if healthCertifiedPerson.isAdmissionStateChanged {
```
pascal-brause, line 1782: I think I would add here an XCTAssertTrue to better show that this is our test core condition. Also for the future, it makes the test more stable.

30mar: when the HealthCertficate service is initiated we can update wallet info twice the fist time the change is recorded but the second time there is no change so the flag is false and this causes the test to fail so in my implementation we can check that it was changed once at least

ArturFriesen (approved):

PR #4320, opened 2022-03-09
Feature/12100 notification if already showing increased risk

KaiTeuber:
## Description
Trigger default notification if another high exposure comes in.
Also show an alert if this happens while processing happened in background task.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12100


CV113 (approved):
UA approved

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
```diff
@@ -1381,6 +1381,14 @@ Bei ausgeschalteter Hintergrundaktualisierung müssen Sie die App täglich aufru
 
 "Home_Alert_RiskStatusLowered_PrimaryButtonTitle" = "OK";
 
+/* Home Alert another high exposure */
+
+"Home_Alert_RiskStatusAnotherHighExposure_Title" = "Neue Risikobegegnung";
+
+"Home_Alert_RiskStatusAnotherHighExposure_Message" = "Sie hatten eine erneute Begegnung mit erhöhtem Risiko. Alle Risikobegegnungen der vergangenen 10 Tage finden Sie in Ihrem Kontakttagebuch.";
```
KathrinLuecke, line 1388: "...alle Risikobegegnungen der vergangenen 10 Tage..." > shouldn't the number of days be a variable maybe that is configurable?  Or is it always 10 days regardless of what the relevant exposure time is (see https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12065)?

KathrinLuecke (commented):
"...alle Risikobegegnungen der vergangenen 10 Tage..." > shouldn't the number of days be a variable maybe that is configurable? Or is it always 10 days regardless of what the relevant exposure time is (see https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12065)?

Ein-Tim:
Wouldn't it make sense to not only have the button "OK" but also "Zum Tagebuch"?

KaiTeuber:
We have another story that will reduce days to 10 - equal to the Android on you pointed out.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -434,15 +434,17 @@ final class RiskProvider: RiskProviding {
 		store.enfRiskCalculationResult = enfRiskCalculationResult
 		store.checkinRiskCalculationResult = checkinRiskCalculationResult
 
+		// first check if a notification might get triggered then write mostRecentDateWithRiskLevel to the store
 		checkIfRiskLevelHasChangedForNotifications(risk)
+		store.mostRecentDateWithRiskLevel = risk.details.mostRecentDateWithRiskLevel
```
ArturFriesen, line 439: Maybe it would be better to save the date inside of `checkIfRiskLevelHasChangedForNotifications`.
This way its clearer, that it belongs to the algorithm of `checkIfRiskLevelHasChangedForNotifications` and its unlikely that the order will be changed here.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -457,16 +459,44 @@ final class RiskProvider: RiskProviding {
...
+		case let .unchanged(riskLevel):
+			guard riskLevel == .high,
+				  let mostRecentDateWithRiskLevel = risk.details.mostRecentDateWithRiskLevel,
+				  let previousMostRecentDateWithRiskLevel = store.mostRecentDateWithRiskLevel,
+				  mostRecentDateWithRiskLevel > previousMostRecentDateWithRiskLevel
+			else {
+					  Log.info("Missing mostRecentDateWithRiskLevel - do not trigger anything")
+					  return
+				  }
+			triggerHighRiskNotification()
```
ArturFriesen, line 476: If the app is processing this in the background, a notification will trigger here.
And then if the user starts the app, another notification will show up because of `store.showAnotherHighExposureAlert = true`. Is this correct?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -457,16 +459,44 @@ final class RiskProvider: RiskProviding {
 	
 	private func checkIfRiskLevelHasChangedForNotifications(_ risk: Risk) {
 		/// Triggers a notification for every risk level change.
-		if risk.riskLevelHasChanged {
```
ArturFriesen, line 460: Could the code be cleaned to not use `risk.riskLevelHasChanged` anymore, but instead check if `risk.riskLevelChange` is not .unchanged? Asking, because I'm a little bit confused about the 2 properties now.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Resources/Localization/de.lproj/Localizable.strings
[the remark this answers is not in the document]
ArturFriesen: I will do it in the context of my story.

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
KaiTeuber: I will have a look if we can refactor it to make `riskLevelHasChanged` a computed property

ArturFriesen (approved):

src/xcode/ENA/ENAUITests/Utils/LaunchArguments.swift
```diff
@@ -64,6 +64,8 @@ enum LaunchArguments {
 		static let numberOfDaysWithRiskLevel = LaunchArgument(name: "numberOfDaysWithRiskLevel")
 		/// Set the level of risk for checkin
 		static let checkinRiskLevel = LaunchArgument(name: "checkinRiskLevel")
+		/// set bool for anotherHighEncounterAlert
+		static let anotherHightEncounter = LaunchArgument(name: "anotherHightEncounter")
```
nickguendling, line 68:
```suggestion
		static let anotherHighEncounter = LaunchArgument(name: "anotherHighEncounter")
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -457,14 +456,42 @@ final class RiskProvider: RiskProviding {
...
-			)
+		switch risk.riskLevelChange {
+		case .decreased:
+			Log.info("decrease risk change state won't trigger a high risk notification")
+		case .increased:
+			triggerHighRiskNotification()
+		case let .unchanged(riskLevel):
+			guard riskLevel == .high,
+				  let mostRecentDateWithRiskLevel = risk.details.mostRecentDateWithRiskLevel,
+				  let previousMostRecentDateWithRiskLevel = store.mostRecentDateWithRiskLevel,
```
nickguendling, line 467: I think we could get rid of the new stored `mostRecentDateWithRiskLevel` if we just passed in the previous risk here. This func is called right after we initialized the new risk and we would still have the previous calculation result available there to also get the previous risk.

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -8,12 +8,23 @@ import OpenCombine
 #if !RELEASE
 
 final class MockTestStore: Store, PPAnalyticsData {
+
+	init() {
+#if DEBUG
+		if isUITesting {
+			self.showAnotherHighExposureAlert = LaunchArguments.risk.anotherHightEncounter.boolValue
+		}
+#endif
```
nickguendling, line 17: For `userNeedsToBeInformedAboutHowRiskDetectionWorks` we do this inside the `HomeTableViewController`. But not sure which version I prefer, only think maybe we can consolidate this, maybe even makes sense to move it to the store as well 🤔

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
KaiTeuber: I think it makes more sense to move it into the mock, this will keep the production code cleaner

nickguendling (approved):

naveeddotio (approved):

PR #4321, opened 2022-03-10
RapidPCR Pending without DGC shows wrong details in testresult

30mar:
## Description
fix the bug using certificateSupportedByPointOfCare

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12007

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/PCRTest+Mock.swift
+++ b/src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/PCRTest+Mock.swift
@@ -31,6 +31,7 @@ extension PCRTest {
 			submissionTAN: submissionTAN,
 			keysSubmitted: keysSubmitted,
 			journalEntryCreated: journalEntryCreated,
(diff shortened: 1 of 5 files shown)


ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/CoronaTestService/__tests__/PCRTest+Mock.swift
```diff
@@ -31,6 +31,7 @@ extension PCRTest {
 			submissionTAN: submissionTAN,
 			keysSubmitted: keysSubmitted,
 			journalEntryCreated: journalEntryCreated,
+			certificateSupportedByPointOfCare: true,
```
nickguendling, line 34: should be added as a parameter so it can be configured by anyone using this mock. And usually we go with false as a default, unless there is a specific reason not to do so

nickguendling (changes requested):

nickguendling (approved):

PR #4322, opened 2022-03-10
Feature/10751 risk calculation max days/12122 filter exposure windows

ArturFriesen:
## Description
This PR filters ExposureWindow based on 'maxEncounterAgeInDays'.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10751


30mar (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/Models/ExposureWindow.swift
```diff
@@ -91,5 +91,18 @@ struct ExposureWindow: Codable, Equatable {
 	let reportType: ENDiagnosisReportType
 	let infectiousness: ENInfectiousness
 	let scanInstances: [ScanInstance]
+	
+	// For each Exposure Window, its `age` shall be calculated as the full as the number of days between its date and the current device time.
+	func age(from now: Date = Date()) -> Int {
```
nickguendling, line 96:
```suggestion
	// For each Exposure Window, its `age` shall be calculated as the full as the number of days between its date and the current device time in UTC.
	func ageInDays(from now: Date = Date()) -> Int {
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/RiskProvider.swift
```diff
@@ -418,7 +418,8 @@ final class RiskProvider: RiskProviding {
 		let configuration = RiskCalculationConfiguration(from: appConfiguration.riskCalculationParameters)
 
 		let enfRiskCalculationResult = enfRiskCalculation.calculateRisk(exposureWindows: exposureWindows, configuration: configuration)
```
nickguendling, line 420: Would it make sense to pass in the filtered windows already instead of filtering twice? In the tech spec filtering seems to be part of `Obtaining Exposure Windows`, not the Risk Calculation itself. So we could even think about returning a filtered result from `executeExposureDetection(...)` already.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Calculation/Models/ExposureWindow.swift
ArturFriesen: done

nickguendling (approved):

naveeddotio (approved):

PR #4324, opened 2022-03-11
use maxEncounterAgeInDays in alert

KaiTeuber:
## Description
Use maxEncounterAgeInDays from the AppConfig inside the alter.
As a fallback add 10 days.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12207

## Screenshots
Simulator Screen Shot - Clone 1 of iPhone 13 Pro - 2022-03-11 at 12 06 12

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
@@ -822,9 +822,12 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 			return
 		}
 
+		let currentAppConfig = appConfigurationProvider.currentAppConfig.value
+		let numberOfDays = currentAppConfig.hasRiskCalculationParameters ? Int(currentAppConfig.riskCalculationParameters.maxEncounterAgeInDays) : 10
+
 		let alert = UIAlertController(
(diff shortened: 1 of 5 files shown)


CV113 (approved):
UA approved

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -822,9 +822,12 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 			return
 		}
 
+		let currentAppConfig = appConfigurationProvider.currentAppConfig.value
+		let numberOfDays = currentAppConfig.hasRiskCalculationParameters ? Int(currentAppConfig.riskCalculationParameters.maxEncounterAgeInDays) : 10
```
nickguendling, line 826: #4325 seems to use a different approach here and defaults to 14, maybe the property from that PR can be used here as well?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
KaiTeuber: It's not available on the release branch yet - I can try to cherry pick it.

nickguendling (approved):

PR #4341, opened 2022-03-16
Fix failing test_screenshot_local_statistics_card()

naveeddotio:
## Description
Removing unnecessary swipe

### DIFF
--- a/src/xcode/ENA/ENAUITests/Home/ENAUITests_01b_Statistics.swift
+++ b/src/xcode/ENA/ENAUITests/Home/ENAUITests_01b_Statistics.swift
@@ -258,7 +258,6 @@ class ENAUITests_01b_Statistics: CWATestCase {
 	
 	func test_screenshot_local_statistics_card() throws {
 		// GIVEN
-		let incidenceTitle = AccessibilityIdentifiers.Statistics.Combined7DayIncidence.title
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENAUITests/Home/ENAUITests_01b_Statistics.swift
```diff
@@ -258,18 +258,13 @@ class ENAUITests_01b_Statistics: CWATestCase {
...
 		let addStatisticsButtonTitle = AccessibilityIdentifiers.LocalStatistics.addLocalIncidencesButton
 		let localStatisticsTitle = AccessibilityIdentifiers.LocalStatistics.localStatisticsCardTitle
 
 		// WHEN
 		app.setPreferredContentSizeCategory(accessibility: .normal, size: .S)
 		app.launch()
 		app.swipeUp(velocity: .slow)
-
-
-		XCTAssert(self.app.staticTexts[incidenceTitle].waitForExistence(timeout: .medium))
```
naveeddotio, line 271: This is not really not needed, the screen is already initialised to what we need. It is checking for something which is only on 10-20% on the screen.

marcussc (approved):

ArturFriesen (approved):

PR #4349, opened 2022-03-18
feature/10297 httpclient key package download/10298 retry

pascal-brause:
## Description
Add the retry functionality to any resource for the new RestService.
For this, any resource has a mandatory property. If the resource should not retry downloading things, the property should be 0. Otherwise, we set an Int on the property with the amount of retries.

## Link to Jira
Mainstory: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10297
Subtask: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10298


src/xcode/ENA/ENA/Source/HTTPClient/Resources/RegistrationToken/RegistrationTokenResource.swift
```diff
@@ -46,7 +46,8 @@ struct RegistrationTokenResource: Resource {
 	var type: ServiceType
 	var sendResource: PaddingJSONSendResource<RegistrationTokenSendModel>
 	var receiveResource: JSONReceiveResource<RegistrationTokenReceiveModel>
-	
+	var retryingCount: Int?
```
marcussc, line 49: Can we have a protocol default implementation that makes adding this to all the resources that don't want to take part in retrying unnecessary?

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
```diff
@@ -14,12 +15,15 @@ protocol Resource {
...
-	
+
 	// Defines a default value for no network cases as the specific receive model (for resources like e.g. AppConfig, AllowList)
 	var defaultModel: Receive.ReceiveModel? { get }
-	
+	// Defines the trust evaluation for the certificate pinning.
 	var trustEvaluation: TrustEvaluating { get }
-	
+	// Indicates if the loading of the resource should be retried. Counts descending.
+	// NOTE: Compete logically with defaultModel: When setting both, we will never execute another retry but instead we will return the defaultModel.
```
ArturFriesen, line 24: Why is that? Wouldn't it make sense to execute every retry, and if all failed, return the default model?

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
```diff
@@ -57,12 +57,33 @@ extension Service {
...
-		
+
+		// To check if we want to retry the load call, we hook in the completion call.
+		let retryingCompletion: ((Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void) = { result in
+			switch result {
+
+			case .success(let model):
+				completion(.success(model))
+			case .failure(let error):
+				// For any failures we check if the resource has a retry flag and has some retries left and if so, we call load() recursivly.
```
ArturFriesen, line 68: I guess there is no retry flag anymore?

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
```diff
@@ -57,12 +57,33 @@ extension Service {
...
+				if let retryingCount = resource.retryingCount,
+				   retryingCount > 0 {
+					Log.debug("Retry for resource discovered. Retry counter at: \(retryingCount)", log: .client)
+					var mutatableResource = resource
+					mutatableResource.retryingCount? -= 1
```
ArturFriesen, line 73: Would it be possible, that multiple threads are reading and writing retryingCount?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/RegistrationToken/RegistrationTokenResource.swift
pascal-brause: No, we can't because we want to change the property during service usage what is forbidden ("Extensions must not contain stored properties").

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
pascal-brause: After short discussion, i change the logic so we can define both: default model and retry. After the retries exhaust, we will return the default model.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
```diff
@@ -57,12 +57,33 @@ extension Service {
 			completion(.success(receiveModel))
 			return
 		}
-		
+
+		// To check if we want to retry the load call, we hook in the completion call.
+		let retryingCompletion: ((Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void) = { result in
```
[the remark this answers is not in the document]
pascal-brause: Converted to private func.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
pascal-brause: Modified the comment to point better out, what we expect here.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
```diff
@@ -14,12 +15,15 @@ protocol Resource {
...
+
 	// Defines a default value for no network cases as the specific receive model (for resources like e.g. AppConfig, AllowList)
 	var defaultModel: Receive.ReceiveModel? { get }
-	
+	// Defines the trust evaluation for the certificate pinning.
 	var trustEvaluation: TrustEvaluating { get }
-	
+	// Indicates if the loading of the resource should be retried. Counts descending.
+	// NOTE: Compete logically with defaultModel: When setting both, we will first exhaust all retries and then when still failing we will return the defaultModel.
+	var retryingCount: Int? { get set }
```
KaiTeuber, line 25: why make the count optional, zero as a default value might be good to indicate a no retrying?

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
pascal-brause: yeah, I will change this which comes with a change to thread-safety of this property.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
pascal-brause: By ensuring that every resource is a struct and we pass a copy in the recursive load call, it should be thread safe now.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
```diff
@@ -231,66 +161,222 @@ extension Service {
...
-			Log.info("Found a defaultModel which conforms to MetaDataProviding. Adding metadata now.", log: .client)
-			modelWithMetadata.metaData.loadedFromDefault = true
-			guard let originalModelWithMetadata = modelWithMetadata as? R.Receive.ReceiveModel else {
-				Log.warning("Cast back to R.Receive.ReceiveModel failed. Returning the model without metadata.", log: .client)
-				return .success(defaultModel)
+		_ responseData: Data? = nil,
+		_ completion: @escaping (Result<R.Receive.ReceiveModel, ServiceError<R.CustomError>>) -> Void
+	) where R: Resource {
+
+		if resource.retryingCount > 1 {
```
ArturFriesen, line 171: Shouldn't it be > 0 ?

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
```diff
@@ -231,66 +161,222 @@ extension Service {
...
+			guard Mirror(reflecting: resource).displayStyle == .struct else {
+				Log.error("Your resource has the wrong type. It should be a struct, not a class!", log: .client)
+				completion(.failure(.wrongResourceType))
```
ArturFriesen, line 174: Asking myself if this should be a fatal error. Because its clearly a programming error.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
pascal-brause: No, otherwise we would count the 0 -loop as a retry too. Then we would change all retry counts from 3 to 2. But here I wanted to keep the old logic to not confuse everyone, where we had 3 retries and it was declared as a "3".

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Services/_Service+Default.swift
pascal-brause: I can change it, makes the code easier because we can remove one more error case.

pascal-brause:
I removed the story branch, so I had to create a new PR: https://github.com/corona-warn-app/cwa-app-ios/pull/4502

PR #4359, opened 2022-03-21
Feature/12350 update old logic to check the blocklist

30mar:
## Description
update old logic to check the blocklist based on the list of blocked certificate refrences that we get from the CCL

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12350


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Services/CCLService/__tests__/DCCWalletInfo+Fake.swift
```diff
@@ -16,7 +16,8 @@ extension DCCWalletInfo {
 		mostRelevantCertificate: DCCCertificateContainer = .fake(),
 		verification: DCCVerification = .fake(),
 		validUntil: Date = Date(),
-		certificateReissuance: DCCCertificateReissuance? = nil
+		certificateReissuance: DCCCertificateReissuance? = nil,
+		certificatesRevokedByInvalidationRules: [DCCCertificateContainer] = [.fake()]
```
nickguendling, line 20:
```suggestion
		certificatesRevokedByInvalidationRules: [DCCCertificateContainer] = []
```
For fakes the default should usually be empty. The caller should not assume anything about the default parameters and set them if they are required for the test case.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -687,12 +687,10 @@ class HealthCertificateService: HealthCertificateServiceServable {
...
 	}
 
-	private func updateValidityState(for healthCertificate: HealthCertificate) {
+	private func updateValidityState(for healthCertificate: HealthCertificate, person: HealthCertifiedPerson) {
 		let previousValidityState = healthCertificate.validityState
 
-		let blockedIdentifierChunks = appConfiguration.currentAppConfig.value
-			.dgcParameters.blockListParameters.blockedUvciChunks
-		if healthCertificate.isBlocked(by: blockedIdentifierChunks) {
+		if let invalidationRules = person.dccWalletInfo?.certificatesRevokedByInvalidationRules, healthCertificate.isBlocked(by: invalidationRules) {
```
nickguendling, line 693: If I would keep it in a separate function, I would probably move it to the person now, but I don't think it makes much sense to have this separate function anymore as it is just a `contains` call.

```suggestion
		if let revokedCertificates = person.dccWalletInfo?.certificatesRevokedByInvalidationRules, 
			revokedCertificates.contains(where: { $0.certificateRef.barcodeData == healthCertificate.base45 }) {
```

Unit tests (which still seem to be to do for this PR) will then have to be done with the service unit tests.

nickguendling (changes requested):

nickguendling (approved):

PR #4364, opened 2022-03-21
wait for the screen to appear before tapping

30mar:
## Description
another attempt to fix the test, the test start and finishes while the launch screen is still loading so the tab bar button cant be tapped, now we wait for the button to appear before tapping

### DIFF
--- a/src/xcode/ENA/ENAUITests/HealthCertificate/ENAUITests_13_CreateHealthCertificate.swift
+++ b/src/xcode/ENA/ENAUITests/HealthCertificate/ENAUITests_13_CreateHealthCertificate.swift
@@ -575,9 +575,7 @@ class ENAUITests_13_CreateHealthCertificate: CWATestCase {
 		app.launch()
 
 		// Navigate to Certificates Tab.
-		let certificateTab = app.buttons[AccessibilityIdentifiers.TabBar.certificates]
-		XCTAssertTrue(certificateTab.waitForExistence(timeout: .medium))
-		certificateTab.waitAndTap()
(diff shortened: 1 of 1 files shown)


naveeddotio (approved):

src/xcode/ENA/ENAUITests/HealthCertificate/ENAUITests_13_CreateHealthCertificate.swift
```diff
@@ -575,7 +575,9 @@ class ENAUITests_13_CreateHealthCertificate: CWATestCase {
 		app.launch()
 
 		// Navigate to Certificates Tab.
-		app.buttons[AccessibilityIdentifiers.TabBar.certificates].waitAndTap()
+		let certificateTab = app.buttons[AccessibilityIdentifiers.TabBar.certificates]
+		XCTAssertTrue(certificateTab.waitForExistence(timeout: .medium))
+		certificateTab.waitAndTap()
```
ArturFriesen, line 580: waitAndTap() is already doing what's in line 579. You can just pass .medium as timeout.

ArturFriesen (changes requested):

src/xcode/ENA/ENAUITests/HealthCertificate/ENAUITests_13_CreateHealthCertificate.swift
30mar: i updated the code and it works locally, thanks

ArturFriesen (approved):

PR #4372, opened 2022-03-23
Set homestate to non nil value for DiaryCoordinator.

ArturFriesen:
## Description
Before that fix the HomeState, which was used in the DiaryCoordinator, was nil.
Hence the registration for risk changes was called against nil.
Because of that the UI was not reloaded then the risk changed from high to low and the high risk entries were not removed.
Now with the fix the HomeState is properly injected into the DiaryCoordinator.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12394


src/xcode/ENA/ENA/Source/RootCoordinator.swift
```diff
@@ -20,6 +20,7 @@ protocol CoordinatorDelegate: AnyObject {
 	Should be used as a delegate in view controllers that need to communicate with other view controllers, either for navigation, or something else (e.g. transferring state).
 	Helps to decouple different view controllers from each other and to remove navigation responsibility from view controllers.
 */
+// swiftlint:disable:next type_body_length
```
marcussc, line 23: Is there no other way?

ArturFriesen: Was shortly thinking about splitting. But its the RootCoordinator 🤷‍♂️

marcussc (approved):

30mar (approved):

KaiTeuber (approved):

PR #4383, opened 2022-03-24
fix saving tests into contact diary

30mar:
## Description
fix saving Tests to contact diary

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12401

## Screenshots
Simulator Screen Shot - iPhone 13 - 2022-03-24 at 15 37 34


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -103,7 +103,12 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 		})
 	}
 
-	func showTestResultScreen(triggeredFromTeletan: Bool = false) {
+	func showTestResultScreen(triggeredFromTeletan: Bool = false, saveToContactDiary: Bool = false) {
+		if saveToContactDiary {
+			model.coronaTestService.createCoronaTestEntryInContactDiary(
```
nickguendling, line 108: Shouldn't the diary entry for negative tests already be there without the need for the user to even open the test result screen? I think the original bug report was only meant for positive tests as we don't directly show the result

ArturFriesen (changes requested):
Could you please describe what was wrong and how you fixed it?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
30mar: Yes then we can save directly as soon as we get the test result, i will update the code

30mar:
> Could you please describe what was wrong and how you fixed it?

The original bgd was that if the user has positive test result and he entered the "Test result is available" screen and then dismissed the screen without actually seeing the test result then opened the diary he would find the test result there.
The solution was to move the saving logic AFTER the user saw the test result, this is why you see some code in the coordinator.
When i tested this i didn't test for negative tests which appears that i broke while fixing the previous bug. that is why we need a second fix

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
ArturFriesen: @30mar  Would a negative test result now be safe twice?

30mar: @ArturFriesen  no i removed the logic for the negative test from the coordinator, so it will be saved once in the service

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -266,10 +271,7 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 				guard let self = self, let coronaTest = self.model.coronaTest else { return }
 
 				guard coronaTest.isSubmissionConsentGiven else {
-					self.model.coronaTestService.createCoronaTestEntryInContactDiary(
-						coronaTestType: self.model.coronaTestType
-					)
-					self.showTestResultScreen()
+					self.showTestResultScreen(saveToContactDiary: true)
```
ArturFriesen, line 274: I think it was better the other way around.
Because just by reading "self.showTestResultScreen(saveToContactDiary: true)"
Its not clear to me what's happening.
Is the test result screen saved to the contact diary?

30mar: @ArturFriesen  you mean moving the logic from this function to where ever it was called?

ArturFriesen: I mean that, self.showTestResultScreen should only show the screen (single responsibility). And the call for creating the test entry should not be contained there. But called separately. Like it was before (the deleted lines 269-272).

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -907,7 +907,9 @@ class CoronaTestService: CoronaTestServiceProviding {
 					if case .positive = testResult, let coronaTest = self.coronaTest(ofType: coronaTestType), !coronaTest.keysSubmitted {
 						self.createKeySubmissionMetadataDefaultValues(for: coronaTest)
 					}
-
+					if case .negative = testResult {
```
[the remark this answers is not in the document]
30mar: @ArturFriesen done and pushed

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
30mar: @ArturFriesen  done and pushed

ArturFriesen (approved):

nickguendling (approved):

PR #4406, opened 2022-03-31
Fix/12367 bg task expiration

ArturFriesen:
## Description
On background task, call completion to fix background task timeout.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12570


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -445,6 +448,7 @@ class HealthCertificateService: HealthCertificateServiceServable {
 			.first()
 			.sink { [weak self] _ in
 				self?.updateValidityStatesAndNotifications()
+				completion()
```
nickguendling, line 451: I took a look at the `updateValidityStatesAndNotifications()` implementation and it seems like the scheduling of notifications is an async operation. Some parts already have completion closures, but not all of them. I think it would make sense to go all the way through with completion closures to make sure the notifications are scheduled before we finish the background task.

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
ArturFriesen: Good point. This escalated quickly. I have added completions to the roots.

marcussc (approved):

nickguendling (approved):

PR #4413, opened 2022-04-01
Move Screenshots into own Workflow

marcussc:
## Description
To not block the CI to much Screenshot will now only be created if the new Screenshot workflow is triggered manually. 
In the future the Screenshots will also be automatically created with a nightly as well.

Also the retrying of Screenshots had to be disabled since this leads to an endless loop :/
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12568

### DIFF
--- a/src/xcode/fastlane/Fastfile
+++ b/src/xcode/fastlane/Fastfile
@@ -141,7 +141,6 @@ platform :ios do
         output_directory: output_dir,
         result_bundle: true,
         skip_open_summary: true,
-        number_of_retries: 2,
         derived_data_path: "./DerivedData",
         test_without_building: options[:test_without_building] == true,
         skip_package_dependencies_resolution: true,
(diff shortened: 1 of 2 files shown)


pascal-brause (approved):

src/xcode/fastlane/Fastfile
```diff
@@ -141,7 +141,6 @@ platform :ios do
         output_directory: output_dir,
         result_bundle: true,
         skip_open_summary: true,
-        number_of_retries: 2,
```
naveeddotio, line 144: so we assume that all the tests will part in first try 😄

naveeddotio (approved):

src/xcode/fastlane/Fastfile
marcussc: Would have been nice if it worked but this way its at least not worse then our current CI xD

PR #4427, opened 2022-04-05
Feature/12406 restore tests from recycle bin

nickguendling:
## Description
So far restoring tests that already were in the recycle bin only worked from the universal QR code scanner for user tests. This PR adds support for family member tests from the universal QR code scanner as well as support for the external camera for both kinds of tests.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12406

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
@@ -25,16 +25,16 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 		antigenTestProfileStore: AntigenTestProfileStoring,
 		vaccinationValueSetsProvider: VaccinationValueSetsProviding,
 		healthCertificateValidationOnboardedCountriesProvider: HealthCertificateValidationOnboardedCountriesProviding,
-		qrScannerCoordinator: QRScannerCoordinator
+		qrScannerCoordinator: QRScannerCoordinator,
+		recycleBin: RecycleBin
 	) {
 		self.parentViewController = parentViewController
+		self.healthCertificateService = healthCertificateService
+		self.healthCertificateValidationService = healthCertificateValidationService
 		self.antigenTestProfileStore = antigenTestProfileStore
 		self.vaccinationValueSetsProvider = vaccinationValueSetsProvider
 		self.healthCertificateValidationOnboardedCountriesProvider = healthCertificateValidationOnboardedCountriesProvider
 		self.qrScannerCoordinator = qrScannerCoordinator
-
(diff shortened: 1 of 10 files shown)


src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -624,6 +634,14 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
 				}
 
 				DispatchQueue.main.async {
+					if let recycleBinItemToRestore = self.model.recycleBinItemToRestore(for: testRegistrationInformation) {
+						self.dismiss {
+							self.showTestRestoredFromBinAlert(recycleBinItem: recycleBinItemToRestore)
```
pascal-brause, line 639: do we need here a [weak self]?

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
```diff
@@ -1493,6 +1511,83 @@ class ExposureSubmissionCoordinator: NSObject, RequiresAppDependencies {
...
+					case .success:
+						self.restoreAndStartWithTest(from: recycleBinItem)
+					case .failure(.testError(.testTypeAlreadyRegistered)):
+						self.showTestOverwriteNotice(recycleBinItem: recycleBinItem)
+					}
+				}
+			)
+		)
+
+		parentViewController?.present(alert, animated: true)
```
pascal-brause, line 1553: Dispatch on main queue needed?

30mar (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: Since the exposure submission coordinator is a bit special in the way that the navigation controller holds the strong reference to the coordinator, it doesn't work with a weak self here. The closure needs to hold a strong self reference so it continues to exist until the alert comes up. But in general weak self is never required for dismiss as the dismiss closure is not stored anywhere (at least not long-term) and therefore no reference cycle can occur

KaiTeuber (approved):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/ExposureSubmissionCoordinator.swift
nickguendling: I tried it out, there seems to be no queue change for the dismiss closure, so no dispatch seems to be needed

PR #4438, opened 2022-04-06
receive on main

marcussc:
## Description
Not only manipulation of ui needs to be done on main-thread, also getting info about an ui object like parentviewcontroller needs to be done on main.

This Issue was revealed through a flaky UITest: `test_RegisterCoronaTestFromSubmitCardButton`


pascal-brause (approved):
Nice finding!

ArturFriesen (approved):
🤩

PR #4441, opened 2022-04-08
Fix/12676 tap on notification when app is terminated opens home

30mar:
## Description

- With the new setup function inside the healthCertificatesService the service is required to call this to handle migration and fetch persons from the store, this operation is async

- The notification manager needs to be updated with a call this setup function in the health service so we can compare the incoming notification identifier with the available certificates, so the structure of the functions was adapted to handle this async call

- Repleced the if/else syntax with a switch statment for better performance and readability

- Update the unit tests 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12676

## Screenshots
https://user-images.githubusercontent.com/15270737/162391012-6e3b8127-49bd-4d82-a83b-00cfc531ab9a.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
@@ -143,7 +143,9 @@ final class NotificationManager: NSObject, UNUserNotificationCenterDelegate {
 						healthCertificate: healthCertificate
 					)
 					Log.debug("Received \(certificateIdentifier.rawValue) notification")
-					self?.showHealthCertificate(route)
+					DispatchQueue.main.async { [weak self] in
+						self?.showHealthCertificate(route)
+					}
 				}
 			})
 		case .boosterVaccination, .certificateReissuance, .admissionStateChange:
@@ -151,7 +153,9 @@ final class NotificationManager: NSObject, UNUserNotificationCenterDelegate {
 				if let certifiedPerson = result {
 					let route = Route(healthCertifiedPerson: certifiedPerson)
 					Log.debug("Received \(certificateIdentifier.rawValue) notification")
-					self?.showHealthCertifiedPerson(route)
+					DispatchQueue.main.async { [weak self] in
+						self?.showHealthCertifiedPerson(route)
+					}
 				}
 			})
 		case .checkout:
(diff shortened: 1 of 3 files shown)


marcussc (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
```diff
@@ -91,18 +91,16 @@ final class NotificationManager: NSObject, UNUserNotificationCenterDelegate {
 	// MARK: - Internal
 	
 	// Internal for testing
-	func extract(_ prefix: String, from: String) -> (HealthCertifiedPerson, HealthCertificate)? {
-		guard from.hasPrefix(prefix) else {
```
ArturFriesen, line 95: Why is this not needed anymore?

src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
```diff
@@ -128,65 +126,68 @@ final class NotificationManager: NSObject, UNUserNotificationCenterDelegate {
...
 		}
 	}
 	
-	private func findHealthCertificate(_ identifier: String) -> (HealthCertifiedPerson, HealthCertificate)? {
-		for person in healthCertificateService.healthCertifiedPersons {
-			if let certificate = person.$healthCertificates.value
-				.first(where: { $0.uniqueCertificateIdentifier == identifier }) {
-				return (person, certificate)
+	private func findHealthCertificate(_ identifier: String, completion: @escaping((HealthCertifiedPerson, HealthCertificate)?) -> Void) {
+		healthCertificateService.setup(updatingWalletInfos: true) { [weak self] in
```
ArturFriesen, line 163: Can it happen, that we now call healthCertificateService.setup twice? Because setup is also called on this instance from the app delegate.

ArturFriesen (changes requested):
Added some questions

src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
30mar: we check the prefix inside the checkForLocalNotificationsActions function as we need to know the enum of the notification for the switch cases

src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
30mar: @ArturFriesen  I just checked if the app is in either foreground, background or terminated mode and we dont call the healthCertificateService.setup() in appdelegate and all the 3 cases so it appears no.
I also added a dispatch statement to show the UI from the main thread

ArturFriesen (approved):

PR #4443, opened 2022-04-08
Fix/12663 app in dead mode after tap on notification on lock screen

30mar:
## Description
The bug had **2 issues**:
1- the first is being locked inside the certificate "which i couldn't reproduce so far"
2- the wrong routing, tapping the notification should show the risk statue on the home screen "**solved**"
since the NEW routing will open the home and not a certificate modal view then the locked out problem is solved anyway but we need to see if the locked problem is reproduced then it can be solved in a separate ticket

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12663

## Screenshots
https://user-images.githubusercontent.com/15270737/162438749-6ca4581e-6fe5-4282-91e3-e364eeb297e5.MP4
https://user-images.githubusercontent.com/15270737/162438847-cfba8bac-67ba-4e41-abab-42edb6b147b0.MP4


src/xcode/ENA/ENA/Source/AppDelegate & Globals/NotificationManager.swift
```diff
@@ -44,9 +44,12 @@ final class NotificationManager: NSObject, UNUserNotificationCenterDelegate {
 		}
 	}
 
+	// swiftlint:disable:next cyclomatic_complexity
```
marcussc, line 47: Is there really no other way?

30mar: @marcussc  not really, it is a delegate function and this is a new case, another way is to it to checkForLocalNotificationsActions() function which is called in the default statement of the switch but this function only handles health Certificate notifications so far, what do you think?

marcussc (approved):

src/xcode/ENA/ENA/Source/Scenes/Home/HomeCoordinator.swift
```diff
@@ -68,6 +68,7 @@ class HomeCoordinator: RequiresAppDependencies {
 				 .none:
 				rootViewController.dismiss(animated: false)
 				rootViewController.popToRootViewController(animated: false)
+				rootViewController.tabBarController?.selectedIndex = 0
```
ArturFriesen, line 71: Seems strange to me, that we show the home tab in case of .checkIn and .healthCertificateFromNotification. Don't we have any routing in this case?

30mar: both .checkIn and .healthCertificateFromNotification are checked in the showHome function in line 118 in the RootCoordinator inside the defer block so basiccly we show the home THEN excute the defer and show the rest

ArturFriesen: Oh wow, its like "GOTO line 42"
Could you add a comment here?

ArturFriesen (approved):

PR #4450, opened 2022-04-12
Fix/12768 revoked certificate cant be changed back to valid

30mar:
## Description
both blocked and invalid states are skipped when we recheck validity state and that keeps the certificate always blocked or always with invalid signature even if it becomes unblocked later.
**The solution**: re calculate if the state is **STILL** blocked instead of checking the old calculated value

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12768


marcussc (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -788,18 +794,15 @@ class HealthCertificateService: HealthCertificateServiceServable {
...
+				updateValidityState(for: healthCertificate, person: healthCertifiedPerson)
 			}
 		}
 	}
 
-	private func updateTimeBasedValidityState(for healthCertificate: HealthCertificate) {
-		guard healthCertificate.validityState != .invalid && healthCertificate.validityState != .blocked else {
-			return
-		}
+	private func updateTimeBasedValidityState(for healthCertificate: HealthCertificate, person: HealthCertifiedPerson) {
```
nickguendling, line 805: Looks like the additional person parameter is not used

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
30mar: done

30mar: @nickguendling

nickguendling (approved):

PR #4453, opened 2022-04-13
Feature/12372 family members rat profile/12648 antigen profile overview

naveeddotio:
## Description
PR to add feature to support antigen test profile for family members

P.S. This PR only fixes the current unit tests, new ones to be done in another task

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12372


KathrinLuecke (approved):
UA approved

30mar (approved):

src/xcode/ENA/ENA/Source/Scenes/AntigenTestProfile/AntigenTestProfileOverview/AntigenTestProfileOverviewViewModel.swift
```diff
@@ -0,0 +1,84 @@
...
+	// MARK: - Init
+
+	init(
+		store: AntigenTestProfileStoring,
+		onEntryCellTap: @escaping (AntigenTestProfile) -> Void
+	) {
+		self.store = store
+		self.onEntryCellTap = onEntryCellTap
+		
+		if let existingProfile = store.antigenTestProfile {
```
ArturFriesen, line 19: Can we remove the property `antigenTestProfile` at store level?
I can think of a solution to remove the property `antigenTestProfile` and add some logic to the getter of `antigenTestProfiles` which tries to read and add `antigenTestProfile` to the list.

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -219,16 +219,29 @@ final class SecureStore: SecureKeyValueStoring, Store, AntigenTestProfileStoring
...
+		get {
+			var antigenTestProfiles = kvStore["antigenTestProfiles"] as [AntigenTestProfile]? ?? []
+			
+			if let existingProfile = kvStore["antigenTestProfile"] as AntigenTestProfile? {
+				// add existing profile to the list of profiles and update store
+				antigenTestProfiles.append(existingProfile)
+				kvStore["antigenTestProfiles"] = antigenTestProfiles
+				
+				// clear the existing profile from the store
+				kvStore["antigenTestProfile"] = nil
```
ArturFriesen, line 234: Not sure if it will work every time.
Szenario: 'antigenTestProfiles' was get, but never set. So only a read operation happened. Then the old profile would get lost?

naveeddotio: If the antigenTestProfiles was never set then antigenTestProfiles will be empty array and if an existing profile is there, will add that profile into the antigenTestProfiles and then set the store in line 231. So, the old profile is retained.

ArturFriesen: Correct, somehow I missed line 231

ArturFriesen (approved):

PR #4459, opened 2022-04-19
Implement sorting

30mar:
## Description
sort the profiles as per the tech specs.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12802


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Models/Profile/AntigenTestProfile.swift
```diff
@@ -91,5 +91,11 @@ struct AntigenTestProfile: Codable {
 	var city: String?
 	var phoneNumber: String?
 	var email: String?
-
+	
+	var fullName: String? {
+		guard let first = firstName, let last = lastName else {
```
ArturFriesen, line 96: Shouldn't the possibility be covered where only firstName or lastName is set?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Models/Profile/AntigenTestProfile.swift
```diff
@@ -91,5 +91,8 @@ struct AntigenTestProfile: Codable {
 	var city: String?
 	var phoneNumber: String?
 	var email: String?
-
+	
+	var fullName: String {
+		return (firstName ?? "") + " " + (lastName ?? "")
```
nickguendling, line 96: Just a suggestion:
```suggestion
		[firstName, lastName]
			.compactMap { $0 }
			.joined(separator: " ")
```
(With this if none is set the result is "" instead of " ")

src/xcode/ENA/ENA/Source/Models/Profile/AntigenTestProfile.swift
30mar: @ArturFriesen this is updated now 👍

ArturFriesen (approved):

PR #4460, opened 2022-04-19
RevocationList Provider update list

KaiTeuber:
## Description
Implemented logic to update revocation list / cache give by tech spec

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12580


nickguendling (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/NotificationManagerTests.swift
```diff
@@ -17,7 +17,7 @@ class NotificationManagerTests: XCTestCase {
 		let expectedName = Name.fake(standardizedFamilyName: "BRAUSE", standardizedGivenName: "PASCAL")
 		
 		let vaccinationCertificate1Base45 = try base45Fake(
-			from: DigitalCovidCertificate.fake(
+			digitalCovidCertificate: DigitalCovidCertificate.fake(
```
ArturFriesen, line 20:
```suggestion
			digitalCovidCertificate: .fake(
```

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService+UITestingMethods.swift
```diff
@@ -27,11 +27,11 @@ extension HealthCertificateService {
 			registerHealthCertificate(base45: HealthCertificateMocks.firstBase45Mock, checkSignatureUpfront: shouldCheckSignatureUpfront, completedNotificationRegistration: { })
 		} else if LaunchArguments.healthCertificate.secondHealthCertificate.boolValue {
 			let secondDose = DigitalCovidCertificateFake.makeBase45Fake(
-				from: DigitalCovidCertificate.fake(
+				certificate: DigitalCovidCertificate.fake(
 					name: .fake(familyName: "Schneider", givenName: "Andrea", standardizedFamilyName: "SCHNEIDER", standardizedGivenName: "ANDREA"),
 					vaccinationEntries: [VaccinationEntry.fake(doseNumber: 2, uniqueCertificateIdentifier: "01DE/84503/1119349007/DXSGWLWL40SU8ZFKIYIBK39A3#E")]
 				),
-				and: CBORWebTokenHeader.fake(issuer: "DE", expirationTime: expirationTime)
+				header: CBORWebTokenHeader.fake(issuer: "DE", expirationTime: expirationTime)
```
ArturFriesen, line 34:
```suggestion
				header: .fake(issuer: "DE", expirationTime: expirationTime)
```

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Revocation/RevocationLocation.swift
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	// MARK: Internal
+
+	let keyIdentifier: String
+	let type: String
+	var certificates: [RevocationCoordinate: [HealthCertificate]]
+
+}
+
```
ArturFriesen, line 23: Lost in space 👾

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Revocation/RevocationProvider.swift
```diff
@@ -0,0 +1,266 @@
...
+								certificate: certificate
+							)
+						}
+					}
+				}
+				certificateByRLC.sort { lhs, rhs in
+					lhs.type < rhs.type
+				}
+
+				DispatchQueue.global().async {
```
ArturFriesen, line 97: Wondering why this dispatch is needed? Aren't we on a background thread already at this point?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Revocation/__tests__/RevocationProviderTests.swift
```diff
@@ -0,0 +1,239 @@
...
+					}
+				)
+			]
+		)
+
+		let revocationProvider = RevocationProvider(restService, signatureVerifier: MockVerifier())
+
+		let expectation = expectation(description: "success expectation")
+
+		revocationProvider.updateCache(with: certificates) {
```
ArturFriesen, line 99: Is it not needed to check if the correct data is in the cache after updating it?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Revocation/__tests__/RevocationProviderTests.swift
nickguendling: The goal of this test from our point of view was to check that the correct requests are done in the correct order. Assuming the RestServiceProvider works correctly (which I guess is tested somewhere else) we then know that the cache will be put-to-date. (And as the `RestServiceProviderStub` has no working cache implementation we cannot easily test this and would only test the stub implementation anyway.)

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Revocation/RevocationProvider.swift
nickguendling: We moved the dispatch inside the function now and added a comment to explain that this is just a safety measure to ensure that the dispatch group will not block the main thread, regardless of the thread on which it is called.

CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/NotificationManagerTests.swift
```diff
@@ -17,7 +17,7 @@ class NotificationManagerTests: XCTestCase {
 		let expectedName = Name.fake(standardizedFamilyName: "BRAUSE", standardizedGivenName: "PASCAL")
 		
 		let vaccinationCertificate1Base45 = try base45Fake(
-			from: DigitalCovidCertificate.fake(
+			digitalCovidCertificate: .fake(
```
ArturFriesen, line 20: What about all the other `DigitalCovidCertificate.fake(` ?

ArturFriesen (approved):

naveeddotio (approved):

PR #4494, opened 2022-04-28
Feature/10307 risk calculation http client/10309 request available hours

KaiTeuber:
## Description
Replaced request availableHours to use in new client.
To realize a special behavior a new feature was added to the client:
* 404 will result in an empty success result (default model)
* all other error will lead to error

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10309

## Tech Spec
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/main/docs/spec/key-download.md


src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient+Configuration.swift
```diff
@@ -88,6 +88,7 @@ extension HTTPClient {
 			)
 		}
 
+		@available(*, deprecated)
```
marcussc, line 91: This will be completley removed once the story is done?

src/xcode/ENA/ENA/Source/HTTPClient/Resources/AvailableHours/AvailableHoursResource.swift
```diff
@@ -0,0 +1,50 @@
...
+	typealias CustomError = Error
+
+	var locator: Locator
+	var type: ServiceType
+	var sendResource: EmptySendResource
+	var receiveResource: JSONReceiveResource<[Int]>
+	var trustEvaluation: TrustEvaluating
+
+	var defaultModelRange: [Int] {
+		[404]
```
marcussc, line 38: Do we have a tech spec we can link here to explain why?

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient+Configuration.swift
KaiTeuber: It gets no longer used, I remived it

src/xcode/ENA/ENA/Source/HTTPClient/Resources/AvailableHours/AvailableHoursResource.swift
KaiTeuber: I copied over the old comment - I'll search for a tech spec

marcussc (approved):

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
```diff
@@ -32,11 +37,25 @@ protocol Resource {
...
+	var defaultModelRange: [Int] {
+		[]
+	}
+
+	// empty default model
 	var defaultModel: Receive.ReceiveModel? {
 		nil
 	}
 
+	// if no default model range is give we always is the default model
```
ArturFriesen, line 51:
```suggestion
	// if no default model range is give we always use the default model
```

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
```diff
@@ -32,11 +37,25 @@ protocol Resource {
...
+	func useFallBack(_ statusCode: Int) -> Bool {
+		if defaultModelRange.isEmpty {
+			return true
+		}
+		return defaultModelRange.contains(statusCode)
```
ArturFriesen, line 56: Is it correct that the available hours call does not have a default model handling for all other errors that are != 404?

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/Resource/_Resource.swift
KaiTeuber: I double check it, yes it is correct.

ArturFriesen (approved):

Issue #4152, opened 2022-01-30
DOB input field scrolls out of view during entering the DOB when requesting a test certificate


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 6s
- iOS version: 15.3
- App version: 2.16.2

## Describe the bug

During entering the DOB for the test certificate, the input field scrolls out of view:

https://user-images.githubusercontent.com/67682506/151712434-8f85f67a-d774-4799-926b-2287e32e1359.mp4

## Steps to reproduce the issue

1. Open the CWA
2. Scroll down to the "Sie lassen sich testen?"
3. Click on "Weiter"
4. Click on "QR-Code scannen"
5. Scan the QR code
6. Click on "Einverstanden"
7. Click on "Einmal teilen"
8. Click in the input field
9. See that the view scrolls to far down

## Expected behaviour

The input field should be visible.

---
Internal Tracking ID: [EXPOSUREAPP-11624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11624)

larswmh:
Thanks for your report @Ein-Tim. We have created an internal ticket for it and will raise this topic internally.
Internal Tracking ID: [EXPOSUREAPP-11624](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11624)

---
Corona-Warn-App Open Source Team

Ein-Tim:
@larswmh 

What's the status of this ticket? This is quite an annoying and easy to fix issue.

Ein-Tim:
PR: https://github.com/corona-warn-app/cwa-app-ios/pull/4495

Ein-Tim:
Confirm fixed with version 2.22. Closing.

PR #4495, opened 2022-04-29
Fix/12784 name textfield moves out of visible area

30mar:
## Description
We re-adjust the layout based on keyboard notification **twice** once in the dynamicTableViewController and another in the TopBottomContainerViewController which causes the bug.
**The current solution**: introduces a flag to prevent keyboard subscription in the dynamicTableViewController if we know that the Viewcontroller will be part of a TopBottomContainerViewController

## Link to Jira
1. fix layout issue the textfield for the family member name
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12784

2. fix layout issue of the date picker for the date of birth
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11624

## Screenshots
https://user-images.githubusercontent.com/15270737/165901369-5b853de6-1294-4805-be1b-efcb0db5a1a0.MP4


Ein-Tim:
Does this also fix #4152?

KaiTeuber:
> Does this also fix #4152?

Yes it should

30mar:
@Ein-Tim thanks for the heads up, the same approach should fix all similar issues, i just pushed a commit to fix your issue as well. 👍

KaiTeuber (changes requested):
changing the target branch only will submit changes from 2.23 to 2.22
I guess you have start work from 2.22, cherry pick commits and reopen a new PR

PR #4496, opened 2022-04-29
Fix/12784 name textfield moves out of visible areas

30mar:
## Description
We re-adjust the layout based on keyboard notification **twice** once in the dynamicTableViewController and another in the TopBottomContainerViewController which causes the bug.
**The current solution**: introduces a flag to prevent keyboard subscription in the dynamicTableViewController if we know that the Viewcontroller will be part of a TopBottomContainerViewController

## Link to Jira
1. fix layout issue the textfield for the family member name
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12784

2. fix layout issue of the date picker for the date of birth
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11624

## Screenshots
Uploading RPReplay_Final1651216078.MP4…


pascal-brause:
Closed for refactoring the solution.

PR #4497, opened 2022-04-29
Feature/10321 refactor test dcc requests

nickguendling:
## Description
Refactor the test certificate request to bring them to the new rest service.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10321


src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -403,89 +352,6 @@ final class HTTPClient: Client {
...
-					}
-					do {
-						let decodedResponse = try JSONDecoder().decode(
-							DCCResponse.self,
-							from: responseBody
-						)
-						completion(.success(decodedResponse))
-					} catch {
-						Log.error("Failed to decode response json", log: .api, error: error)
-						completion(.failure(.jsonError))
```
pascal-brause, line 439: For this error you take now the default one of the RestService? Because I could not find this one in customError() in the Resource

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
```diff
@@ -5,7 +5,16 @@
 #if !RELEASE
 
 struct LoadResource {
-	let result: Result<Any, Error>
+
+	init(
+		result: @escaping @autoclosure () throws -> Result<Any, Error>,
```
pascal-brause, line 10: 😳

pascal-brause (commented):
So much code removed 🚀

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
nickguendling: Exactly, just using the default one, and the error message is the same as both cases had the `.tryAgain` error text

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -305,57 +305,6 @@ final class HTTPClient: Client {
...
-						Log.error("Error in response with status code: \(String(response.statusCode))", log: .api)
-						completion(.failure(.unhandledResponse(response.statusCode)))
-					}
-				case let .failure(error):
-					if case .noNetworkConnection = error {
-						Log.error("No network connection", log: .api)
-						completion(.failure(.noNetworkConnection))
-					} else {
-						Log.error("Error in response body", log: .api, error: error)
-						completion(.failure(.defaultServerError(error)))
```
ArturFriesen, line 352: I can't see this error handled in the new resource. Did you consider it?

src/xcode/ENA/ENA/Source/HTTPClient/Resources/DigitalCovid19Certificate/DigitalCovid19CertificateResource.swift
```diff
@@ -0,0 +1,94 @@
...
+	// MARK: - Init
+
+	init(
+		isFake: Bool = false,
+		sendModel: DigitalCovid19CertificateSendModel,
+		trustEvaluation: TrustEvaluating = DefaultTrustEvaluation(
+			publicKeyHash: Environments().currentEnvironment().pinningKeyHashData
+		)
+	) {
+		self.locator = .digitalCovid19Certificate(isFake: false)
```
ArturFriesen, line 18:
```suggestion
		self.locator = .digitalCovid19Certificate(isFake: isFake)
```

src/xcode/ENA/ENA/Source/HTTPClient/Resources/DCCPublicKeyRegistration/DCCPublicKeyRegistrationResource.swift
```diff
@@ -0,0 +1,72 @@
...
+	// MARK: - Init
+
+	init(
+		isFake: Bool = false,
+		sendModel: DCCPublicKeyRegistrationSendModel,
+		trustEvaluation: TrustEvaluating = DefaultTrustEvaluation(
+			publicKeyHash: Environments().currentEnvironment().pinningKeyHashData
+		)
+	) {
+		self.locator = .dccPublicKeyRegistration(isFake: false)
```
ArturFriesen, line 18:
```suggestion
		self.locator = .dccPublicKeyRegistration(isFake: isFake)
```

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
```diff
@@ -403,89 +352,6 @@ final class HTTPClient: Client {
...
-		registrationToken token: String,
-		isFake: Bool,
-		completion: @escaping DigitalCovid19CertificateCompletionHandler
-	) {
-		guard let request = try? URLRequest.dccRequest(
-				configuration: configuration,
-				registrationToken: token,
-				headerValue: isFake ? 1 : 0
-		) else {
-			completion(.failure(.urlCreationFailed))
```
ArturFriesen, line 417: This error is not handled in the new code. Did you consider it?

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateServiceError.swift
```diff
@@ -106,45 +106,39 @@ enum HealthCertificateServiceError: Error {
 			switch self {
 			case .publicKeyRegistrationFailed(let registrationError):
 				switch registrationError {
-				case .badRequest:
+				case .receivedResourceError(.badRequest):
 					return String(format: AppStrings.HealthCertificate.Overview.TestCertificateRequest.Error.clientErrorCallHotline, "PKR_400")
```
ArturFriesen, line 110: Would it be possible to move the strings to DCCPublicKeyRegistrationError (also for the other error types? It would cleanup this switch.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
ArturFriesen: Why is it needed? Tried to find out where it is used, but compiler still compiles with other LoadResource 🤷‍♂️

ArturFriesen (changes requested):

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
nickguendling: It's handled by the usual service errors now, no need for a separate case to wrap it anymore

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
nickguendling: Yes, our locators are not failable, so no need to handle this anymore

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
nickguendling: This doesn't change the call side. It just packs whatever is at the call side inside a closure and we can execute it later, only when needed. I needed this in my unit tests because the result of the second request depends on the first already being executed. In the old client implementation I was able to use a closure to define the result of a request, so I went with the `@autoclosure` here, as I was able to keep the same structure for the tests that needed to be refactored and I didn't need to change much on the ones that were already using the `LoadResource`.

src/xcode/ENA/ENA/Source/Client/HTTP Client/HTTPClient.swift
pascal-brause: And we have a default `invalidRequestError` which is returned by the service when the URL creation fails.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateServiceError.swift
nickguendling: I tried to move the `LocalizedError` conformances to the separate errors now where possible. For some it was not possible because the are defined in the HealthCertificateToolkit and sometimes reused in other places.

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
ArturFriesen: Could you point me to one of these tests?

pascal-brause (approved):

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
nickguendling: One example is `testTestCertificateRegistrationAndExecution_Success()` where the `keyPair` for the second LoadResource is only available after the first LoadResource went through. Most of the tests in the same test class do the same thing

naveeddotio (approved):

src/xcode/ENA/ENA/Source/HTTPClientCore/Service/__tests__/RestServiceProviderStub.swift
ArturFriesen: Ok got it now. Nice solution! Didn't know this trick 🧙‍♂️

ArturFriesen (approved):

Issue #4454, opened 2022-04-13
Certificates tab overscrolls if "Certificates" tab icon is tapped after deleting a certificate


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: iPhone XR 
- iOS version: 15.4.1
- App version: 2.20.3

## Describe the bug

The certificates tab overscrolls if the "Certificates" tab icon is tapped to scroll to the top.
"Overscroll" means that there is empty area at the top of the screen, as shown in this screenshot:
<img src="https://user-images.githubusercontent.com/67682506/163226444-acb4569b-f495-4b39-a39f-b85168f2adc2.PNG" width="300">

## Steps to reproduce the issue

1. Open the CWA
2. Open the Certificates tab
3. Add a certificate
4. Delete the certificate
5. Tab on the "Certificates" tab icon
6. See that the view overscrolls

## Expected behaviour

View should not overscroll.

## Possible Fix

Probably similar to https://github.com/corona-warn-app/cwa-app-ios/pull/4344.

## Additional context

Maybe something @KaiTeuber, as he also authored the PR https://github.com/corona-warn-app/cwa-app-ios/pull/4344, would like to fix this. 

---
Internal Tracking ID: [EXPOSUREAPP-12880](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12880)

larswmh:
@Ein-Tim is this still an issue in the latest version? I was not able to reproduce this on iPhone 11 Pro Max, CWA 2.21.2, iOS 15.5 (beta version)

larswmh:
Thanks for your report @Ein-Tim. We have created an internal ticket for it and will raise this topic internally.
Internal Tracking ID: [EXPOSUREAPP-12880](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12880)

---
Corona-Warn-App Open Source Team

Ein-Tim:
PR https://github.com/corona-warn-app/cwa-app-ios/pull/4517 is the fix for this.

Ein-Tim:
[...]

PR #4517, opened 2022-05-06
Remove obsolete reload data that interferes with scrolling to top

nickguendling:
## Description
Removes the reload data call [originally introduced to update the permission cell](https://github.com/corona-warn-app/cwa-app-ios/pull/3124). It is now obsolete as the permission cell is not shown on this screen anymore. The reload interfered with the scroll-to-top feature on the second tab bar item tap (despite the previously attempted fix). Now the screen scrolls up smoothly again and doesn't over scroll.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12972


marcussc (approved):

nickguendling:
> Does this fix #4454 (asking because the JIRA ticket linked in the issue is not the same as the ticket referred to here)?

@Ein-Tim Yes, looks like that is the same issue. Seems like we have multiple JIRA tickets for that one. Should be fixed with this as well.

KaiTeuber (approved):

PR #4534, opened 2022-05-13
Feature/13096 support old wallet versions + update notifications

30mar:
## Description

- add support to the old wallet certificates from old versions
- update notifications logic

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13096

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
@@ -188,7 +188,7 @@ class HealthCertificateNotificationService {
 			return
 		}
 		
-		 if newIdentifier != previousCertificateReissuance?.reissuanceDivision.identifier {
+		 if newIdentifier != previousCertificateReissuance?.reissuanceDivision.identifier ?? "renew" {
 			guard let personIdentifier = person.identifier else {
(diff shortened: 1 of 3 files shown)


ArturFriesen (approved):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
```diff
@@ -180,14 +180,15 @@ class HealthCertificateNotificationService {
...
+			  let newIdentifier = newCertificateReissuance.reissuanceDivision.identifier else {
 			Log.info("No certificate reissuance found for person \(private: String(describing: name))", log: .vaccination)
 			completion()
 
 			return
 		}
-
-		if newCertificateReissuance != previousCertificateReissuance {
+		
+		 if newIdentifier != previousCertificateReissuance?.reissuanceDivision.identifier ?? "renew" {
```
nickguendling, line 191: Just to document our discussion: The fallback should only be for the case that the `identifier` is nil, not if the `previousCertificateReissuance` itself is nil

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateNotificationService.swift
30mar: @nickguendling  updated 👍

nickguendling (approved):

PR #4535, opened 2022-05-13
Fix/12957 risk level change

ArturFriesen:
## Description
Fixes the calculation of risk level change.
Before that fix it was possible to have an increased risk change without having an actual increased risk.

For example, without the fix, this set of risks returned risk change: .increased (the correct change is .unchanged(.high))
currentENFRisk: .high, previousENFRisk: .high, currentCheckinRisk: .low, previousCheckinRisk: .high,

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12627


src/xcode/ENA/ENA/Source/Services/Risk/Risk.swift
```diff
@@ -100,25 +157,24 @@ extension Risk {
...
-		self.init(
-			level: totalRiskLevel,
-			details: details,
-			riskLevelChange: riskLevelChange
-		)
+		
+		return details
+	}
+	
+	// Merge the results from both risk calculation. For each date, the higher risk level is used.
```
marcussc, line 164:
```suggestion
	// Merge the results from both risk calculation. For each date, the highest risk level is used.
```

marcussc (approved):

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/RiskTests.swift
```diff
@@ -114,6 +114,73 @@ class RiskTests: CWATestCase {
 		XCTAssertEqual(risk.details.mostRecentDateWithRiskLevel, today)
 		XCTAssertEqual(risk.details.numberOfDaysWithRiskLevel, 1)
 	}
+	
+	func test_Given_PreviousENFRiskLow_And_PreviousCheckinRiskLow_When_ENFRiskHigh_And_CheckinRiskLow_Then_ChangeIsIncreased() {
```
nickguendling, line 118: I think the name is outdated

src/xcode/ENA/ENA/Source/Services/Risk/Risk.swift
```diff
@@ -74,21 +119,33 @@ extension Risk {
...
-		Log.debug("[Risk] totalRiskLevel: \(totalRiskLevel)", log: .riskDetection)
-
+		
+		return totalRiskLevel
+	}
+	
+	private static func riskDetails(
+		enfRiskCalculationResult: ENFRiskCalculationResult,
+		checkinCalculationResult: CheckinRiskCalculationResult,
+		risk: RiskLevel
```
nickguendling, line 129:
```suggestion
		riskLevel: RiskLevel
```

src/xcode/ENA/ENA/Source/Services/Risk/Risk.swift
```diff
@@ -54,16 +54,61 @@ extension Risk {
...
+		if previousTotalRiskLevel == totalRiskLevel {
+			riskLevelChange = .unchanged(totalRiskLevel)
+		} else {
+			riskLevelChange = previousTotalRiskLevel > totalRiskLevel ? .decreased : .increased
+		}
+		
+		Log.debug("[Risk] riskLevelChange: \(riskLevelChange)", log: .riskDetection)
 
-		Log.debug("[Risk] tracingRiskLevelPerDate: \(tracingRiskLevelPerDate)", log: .riskDetection)
-		Log.debug("[Risk] checkinRiskLevelPerDate: \(checkinRiskLevelPerDate)", log: .riskDetection)
```
nickguendling, line 61: Would it make sense to still log these somewhere, in case we need to debug later?

nickguendling (changes requested):

src/xcode/ENA/ENA/Source/Services/Risk/__tests__/RiskTests.swift
ArturFriesen: Right. Renamed.

src/xcode/ENA/ENA/Source/Services/Risk/Risk.swift
ArturFriesen: Renamed

src/xcode/ENA/ENA/Source/Services/Risk/Risk.swift
ArturFriesen: Good point. Added logs.

nickguendling (approved):

PR #4575, opened 2022-05-23
Add new property to track test result change to invalid.

ArturFriesen:
## Description
With this fix, the test result change from pending, positive or negative to invalid is considered and the red dot is shown.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12667


src/xcode/ENA/ENA/Source/Services/CoronaTestService/FamilyMemberCoronaTestService.swift
```diff
@@ -788,7 +791,6 @@ class FamilyMemberCoronaTestService: FamilyMemberCoronaTestServiceProviding {
 					isNew: false,
 					testResult: testResult,
 					finalTestResultReceivedDate: testResult == .pending ? nil : Date(),
-					testResultWasShown: LaunchArguments.familyMemberTest.pcr.positiveTestResultWasShown.boolValue,
```
nickguendling, line 791: Do we use the launch arguments (this one and the one for antigen) anywhere? I think we should either replace them with new ones for the new property or remove them completely

nickguendling (changes requested):

nickguendling (approved):

30mar (approved):

PR #3144, opened 2021-07-13
Feature/#5612 Translucent Tab Bar (Implements #2159)

Ein-Tim:
## Description
This PR removes the definition of the tab bar color so that it's now translucent. 
This is the standard under iOS and will make CWA look even better integrated into the System as well as match native apps like the App Store, etc. too.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5612

## Screenshots
| Current Tab Bar | New Tab Bar |
|---|---|
|  |  |

## Additional information

This is a follow up to PR #2887, which merged in the `release/2.5.x` branch. For some reasons, this change was lost and thus isn't live.

PR #4584, opened 2022-05-24
Fix/9816 tabbar items low contrast

naveeddotio:
## Description
- Changing current assets 
- Adding tint to unselected tab bar items

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9816

## Screenshots
Low Contrast:
MicrosoftTeams-image (2)
High Contrast:
Screenshot 2022-05-24 at 17 46 15


src/xcode/ENA/ENA/Source/RootCoordinator.swift
```diff
@@ -277,6 +277,7 @@ class RootCoordinator: NSObject, RequiresAppDependencies, UITabBarControllerDele
 		diaryCoordinator.viewController.tabBarItem = diaryTabBarItem
 
 		tabBarController.tabBar.tintColor = .enaColor(for: .tint)
+		tabBarController.tabBar.unselectedItemTintColor = .enaColor(for: .textPrimary2)
```
Ein-Tim, line 280: Does this revert the change done in https://github.com/corona-warn-app/cwa-app-ios/pull/3144 (translucent tab bar to match iOS native apps)?

nickguendling, line 280: I think you've used two incompatible approaches here. You could either add more images (as you did) and set the rendering mode to original, or keep the template images and add high contrast colors in the color asset catalogue and set both tint colors as you did here. I think extending the colors in the asset catalogue would be preferable as it is more flexible (and it would add support for high contrast mode in other places where we use the colors)

nickguendling (changes requested):

nickguendling (approved):

ArturFriesen (approved):

KaiTeuber (approved):

PR #4587, opened 2022-05-25
removes the static text check

30mar:
## Description
remove the staticText check that was failing the UITests


naveeddotio (approved):

marcussc (approved):

src/xcode/ENA/ENAUITests/AntigenTestProfile/ENAUITests_12_AntigenTestProfile.swift
```diff
@@ -133,9 +133,6 @@ class ENAUITests_12_AntigenTestProfile: CWATestCase {
 		XCTAssertTrue(app.buttons[AccessibilityIdentifiers.ExposureSubmission.AntigenTest.Profile.editButton].waitForExistence(timeout: .long))
 
 		app.swipeUp(velocity: .slow)
```
nickguendling, line 135:
```suggestion
```
I guess it doesn't make sense to swipe up anymore to then just tap the close button

nickguendling (changes requested):

src/xcode/ENA/ENAUITests/AntigenTestProfile/ENAUITests_12_AntigenTestProfile.swift
30mar: @nickguendling  you are right, done

nickguendling (approved):

PR #4588, opened 2022-05-25
Fix/11498 request tan call button

naveeddotio:
## Description
Adding missing accessibility label

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11489


CV113 (approved):
UA approved

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionHotlineViewController.swift
```diff
@@ -105,7 +105,7 @@ class ExposureSubmissionHotlineViewController: DynamicTableViewController, ENANa
 							style: .headline,
 							color: .enaColor(for: .textTint),
 							title: AppStrings.ExposureSubmissionHotline.phoneNumberDomestic,
-							accessibilityLabel: AppStrings.ExposureSubmissionHotline.callButtonTitle,
+							accessibilityLabel: "\(AppStrings.ExposureSubmissionHotline.callButtonTitle):\n\n\(AppStrings.ExposureSubmissionHotline.phoneNumberDomestic)",
```
KaiTeuber, line 108: please move to `AccessibilityIdentifiers`

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionHotlineViewController.swift
```diff
@@ -122,7 +122,7 @@ class ExposureSubmissionHotlineViewController: DynamicTableViewController, ENANa
 							style: .headline,
 							color: .enaColor(for: .textTint),
 							title: AppStrings.ExposureSubmissionHotline.phoneNumberForeign,
-							accessibilityLabel: AppStrings.ExposureSubmissionHotline.callButtonTitle,
+							accessibilityLabel: "\(AppStrings.ExposureSubmissionHotline.callButtonTitle):\n\n\(AppStrings.ExposureSubmissionHotline.phoneNumberForeign)",
```
KaiTeuber, line 125: same here

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionHotlineViewController.swift
ArturFriesen, line 108: Why are the new lines needed?

KaiTeuber, line 108: I wonder about formatting here. Won't it make more sense to adde a new AppString?
And as @ArturFriesen  mentioned I also wonder about the newlines.

KaiTeuber (changes requested):

src/xcode/ENA/ENA/Source/Scenes/ExposureSubmission/View/Controller/ExposureSubmissionHotlineViewController.swift
naveeddotio: @KaiTeuber I removed the line breaks, I would suggest to use the current strings and not create a new string, reason: if we need to change the number, we have to do at two different places and if we forget to change then the voice over will be different to that of the exact number.

KaiTeuber (approved):

PR #4594, opened 2022-05-27
Fix/12215 app information faq link twice

KaiTeuber:
## Description
Remove FAQ link from the app information page because it on home screen's more section already

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12215

## Screenshots
https://user-images.githubusercontent.com/2675578/170706920-3ef1427b-7cdb-46a4-809c-7bb6379a89b1.mp4


naveeddotio (approved):

nickguendling (dismissed):
I think this is a sensible suggestion by the tester, but in Figma the link exists in both places. So it should be clarified with UX and Figma should be updated. Otherwise we will get a reverse bug ticket at some point about the missing FAQ link 😄

KaiTeuber:
> I think this is a sensible suggestion by the tester, but in Figma the link exists in both places. So it should be clarified with UX and Figma should be updated. Otherwise we will get a reverse bug ticket at some point about the missing FAQ link 😄

KaiTeuber:
i closed the PR until it's clarified by Anna-Maria

flxschmidt (approved):
LGTM 👍

CV113 (approved):
UA approved

PR #4608, opened 2022-06-03
Reproducible Status badge re-appear

30mar:
## Description
The issue was that if the device has a high risk and we go to the background then back to foreground a badge appears in the Status tab.
After investigation the reason was that inside the Risk.swift file, during calculation of the risk level we have a **default  previousRisk of .low.**

We call this function in 2 ways, one with providing the **previousENFRiskCalculationResult** then in "the old implementation" the previousTotalRiskLevel will remain low with is wrong and with current high calculation we will always get the red badge on the home tab.

if the previous risk is nil then the risk level should be unchanged and not nil.

And since we dont always pass previous risks to compare with, this make the device compare any new risk with Low "specialy in the case of returning from the background as we don't pass previous risks there".

The solution was to make the default risk level an optional and set it as unchanged if there is no previous risk level

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13184

## Screenshots
https://user-images.githubusercontent.com/15270737/171903543-02d77f88-54e2-4de2-a18a-bf987feebe2a.MP4


flxschmidt (approved):
LGTM

naveeddotio (approved):

Ein-Tim:
This PR is assigned to the wrong milestone. It is in the 2.24 milestone but should be in the 2.25 one.

cc @30mar, @nickguendling, @flxschmidt & @naveeddotio

Issue #3728, opened 2021-10-20
Translucent tab bar in contact journal


max-ae:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: iPhone 12 Pro Max
- iOS version: 15.0.2
- App version: 2.12.1

## Describe the bug

Tab bar in diary view is translucent when selecting person or location

## Steps to reproduce the issue

1. go to contact journal
2. tap on a day
3. see bug

![image](https://user-images.githubusercontent.com/4730112/138117130-64e15777-2b8b-49cc-9f70-bb01fe1e641a.jpeg)

---
Internal Tracking ID: [EXPOSUREAPP-9910](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9910)

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-10178](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10178) duplicate of
Internal Tracking ID: [EXPOSUREAPP-9910](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9910)

Ein-Tim:
PR: #3788 merging into `release/2.14.x`.

Ein-Tim:
@dsarkar Please, could you inform the team that PR #3788 did **not** sufficiently fix this issue? See https://github.com/corona-warn-app/cwa-app-ios/issues/3161#issuecomment-1032963516.

max-ae:
Addition from my side:

Bug also occurs if the following is done:
1. select a day in the journal
2. slightly swipe from the left hand side of the screen as if going back to the overview, but don't complete the swipe to stay on the day view
3. tab bar is transparent again

dsarkar:
@max-ae Thanks for reporting. In which iOS do you reproduce this issue?

max-ae:
> @max-ae Thanks for reporting. In which iOS do you reproduce this issue?

iOS 15.3.1 on iPhone 12 Pro Max, CWA version 2.18.1

Ein-Tim:
[...]

PR #4639, opened 2022-06-24
Fix/9404 transparent status bar

naveeddotio:
## Description
Fixing transparent status bar issue on two remaining screens

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9404


Ein-Tim:
I think this would be the optimal way to also fix https://github.com/corona-warn-app/cwa-app-ios/issues/3728#issuecomment-1050160505

naveeddotio:
> I think this would be the optimal possibility to also fix [#3728 (comment)](https://github.com/corona-warn-app/cwa-app-ios/issues/3728#issuecomment-1050160505)

This fix is only confined to the devices with iOS 12.5

flxschmidt (approved):

30mar (approved):

PR #4663, opened 2022-07-11
Fix/13476 delete exported certificates from temporary directory

naveeddotio:
## Description
Deleting exported certificates (pdf documents) from temporary directory
- When the certificate preview screen is closed
- When the app has freshly started (to cater the scenario when app crashes)
- When the app is killed by the user or the OS 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13476


30mar (approved):

src/xcode/ENA/ENA/Source/Extensions/FileManager+ClearTemporaryDirectory.swift
```diff
@@ -0,0 +1,22 @@
+//
+// 🦠 Corona-Warn-App
+//
+
+import Foundation
+
+ extension FileManager {
+	func removePDFsFromTemporaryDirectory() {
```
flxschmidt, line 8: Optional:
We could handle that stuff on a background thread, because it's independent from the sync stuff on main thread.

```swift
DispatchQueue.global(qos: .background).async { ... }
```

src/xcode/ENA/ENA/Source/Extensions/FileManager+ClearTemporaryDirectory.swift
```diff
@@ -0,0 +1,22 @@
...
+		do {
+			let temporaryDirectoryURL = FileManager.default.temporaryDirectory
+			let temporaryDirectory = try FileManager.default.contentsOfDirectory(atPath: temporaryDirectoryURL.path)
+			try temporaryDirectory.forEach { file in
+				if file.contains(".pdf") {
+					let fileUrl = temporaryDirectoryURL.appendingPathComponent(file)
+					try FileManager.default.removeItem(atPath: fileUrl.path)
```
flxschmidt, line 15: Do you think we should write a unit test with a dummy file for it, for the BSI I mean?

flxschmidt (commented):
LGTM! :) Two minor picks

flxschmidt (approved):
Nice :)

PR #4691, opened 2022-07-21
Fix/13462: Expired Certificates - "About to expire" needs to be removed on personal wallet on screen "Zertifikate"

flxschmidt:
## Description
Show caption for Test certificates that are invalid or blocked, or VC or RC certificates, that are not valid or will not expire soon.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13462


naveeddotio (changes requested):
Unit tests need to be fixed as well.

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Overview/Cells/HealthCertifiedPerson/HealthCertifiedPersonCellModel.swift
```diff
@@ -62,6 +62,19 @@ class HealthCertifiedPersonCellModel {
 		}
 
 		self.onTapToDelete = nil
+		
+		func shouldShowCaptionFor(initialCertificate certificate: HealthCertificate) -> Bool {
```
naveeddotio, line 66: Please move this function outside of the init and make it private.

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Overview/Cells/HealthCertifiedPerson/HealthCertifiedPersonCellModel.swift
```diff
@@ -62,6 +62,19 @@ class HealthCertifiedPersonCellModel {
...
+			// Test certificates that are invalid or blocked.
+			if certificate.type == .test {
```
naveeddotio, line 68: Would it make sense to cater both conditions with one if statement?

flxschmidt: That's a good question. I had that in your way, but then the code are harder to read.

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Overview/Cells/HealthCertifiedPerson/HealthCertifiedPersonCellModel.swift
flxschmidt: The reason I used an `inner func` is, that the logic I replaced with this PR get wrapped in a naming function, that encapsulate the code and make it more readable.
Unfortunately you cannot use a function (as a `private` one in a constructor), before the initialization of stored properties like `self.caption` is done.
You would get that compiler error:
```swift
'self' used in method call 'shouldShowCaptionFor' before all stored properties are initialized
```

As an alternative we could implement that function as an `extension` of `HealthCertificate`.
But that logic is view dependent and not globally valid. That's why I implement it in the regarding ViewModel class.

naveeddotio: I had with Omer and we recommend if you could just use the code with out the function in else block since it is just called once :)

30mar: exactly so just extract the code block from the function and use it directly if that is ok with you 👍

flxschmidt: okay, the inner func is now removed. 👍

naveeddotio (approved):

30mar (approved):

PR #4697, opened 2022-07-26
PPA submission: Error API_TOKEN_ALREADY_ISSUED shown

30mar:
## Description
the testers couldn't force submit the PPA as the forceApiHeader needed to be added to the headers

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13691

## Screenshots
https://user-images.githubusercontent.com/15270737/181000442-2a01c8c8-a1c4-48d3-be5a-9fde6d8e48bc.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
@@ -17,15 +17,24 @@ extension Locator {
 	) -> Locator {
 		let fake = String(isFake ? 1 : 0)
 		let forceApiHeader = String(forceApiTokenHeader ? 1 : 0)
+		
+		var defaultHeaders = [
+			"Content-Type": "application/x-protobuf",
+			"cwa-ppac-ios-accept-api-token": forceApiHeader,
+			"cwa-fake": fake
+		]
+		
+		#if !RELEASE
+		if forceApiTokenHeader {
+			defaultHeaders["cwa-ppac-ios-accept-api-token"] = "1"
+		}
+		#endif
+
 		return Locator(
 			endpoint: .dataDonation,
 			paths: ["version", "v1", "ios", "dat"],
 			method: .post,
-			defaultHeaders: [
-				"Content-Type": "application/x-protobuf",
-				"cwa-ppac-ios-accept-api-token": forceApiHeader,
-				"cwa-fake": fake
-			]
+			defaultHeaders: defaultHeaders
 		)
 	}
 
--- a/src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
@@ -12,21 +12,17 @@ extension Locator {
 	// comment:	Custom error handling required
 	static func submitPPA(
 		payload: SAP_Internal_Ppdd_PPADataIOS,
-		forceApiTokenHeader: Bool = false,
-		isFake: Bool
+		forceApiTokenHeader: Bool = false
 	) -> Locator {
-		let fake = String(isFake ? 1 : 0)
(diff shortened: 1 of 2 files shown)


naveeddotio (approved):

puneetmahali1 (approved):

src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
```diff
@@ -17,15 +17,24 @@ extension Locator {
 	) -> Locator {
 		let fake = String(isFake ? 1 : 0)
 		let forceApiHeader = String(forceApiTokenHeader ? 1 : 0)
+		
+		var defaultHeaders = [
+			"Content-Type": "application/x-protobuf",
+			"cwa-ppac-ios-accept-api-token": forceApiHeader,
+			"cwa-fake": fake
```
mlenkeit, line 24: I think before the refactoring, these were not set by default. Can you please double-check?

https://github.com/corona-warn-app/cwa-app-ios/compare/v2.24.1..v2.25.0-rc.5#diff-931f7490dd281c0b996106c696f4925213315ec8dbee5f55c7926af6d6054ab4

mlenkeit (changes requested):

src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
30mar: sure

30mar: for the second we do this for all the refactored requestsm for example check https://github.com/corona-warn-app/cwa-app-ios/blob/release/2.26.x/src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/KeySubmission.swift

So this was already being done before but directly in the session

30mar: for cwa-ppac-ios-accept-api-token we have **#if !RELEASE** before and we have it now so we are sure we never add this in production

30mar: for the **isFake** flag, what we normally did is to fire a request normally and then after we get the response we check the isFake flag, if it is set to true we return an error that it is a fakeRequest, this is for Plausible-deniability so it is done **also on production**,  here are some screenshot how we checked it before the new rest service and how we do it now after the refactoring:
Old HTTP Client:
Screenshot 2022-07-27 at 09 34 17
New Rest Service:
Screenshot 2022-07-27 at 09 50 44.
PS: We dont show a visual error to the user in this case

src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
```diff
@@ -12,21 +12,17 @@ extension Locator {
...
-		let forceApiHeader = String(forceApiTokenHeader ? 1 : 0)
 		
 		var defaultHeaders = [
-			"Content-Type": "application/x-protobuf",
-			"cwa-ppac-ios-accept-api-token": forceApiHeader,
-			"cwa-fake": fake
+			"Content-Type": "application/x-protobuf"
 		]
 		
+		let forceApiHeader = String(forceApiTokenHeader ? 1 : 0)
```
mlenkeit, line 22: @30mar this could be moved one line down into the `#if !RELEASE` block AFAIK

mlenkeit (approved):

src/xcode/ENA/ENA/Source/HTTPClient/LocatorsNotYetUsed/SubmitPPA.swift
naveeddotio: Exactly, if is it not built for RELEASE, variable forceApiHeader is useless.

PR #4704, opened 2022-08-02
Feature/support dcc schema 1.3.2

30mar:
## Description
support dcc schema 1.3.2 and update 1.3.2

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13663

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
@@ -45,9 +45,6 @@ extension Name {
 	}
 	
 	var reversedStandardizedName: String? {
-		guard var standardizedFamilyName = self.standardizedFamilyName else {
-			return nil
-		}
 		if self.standardizedFamilyName != nil && self.standardizedGivenName != nil {
 			standardizedFamilyName += "<<"
 		}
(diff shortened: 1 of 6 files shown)


src/xcode/ENA/ENA/Source/Services/HealthCertificate/Models/Extensions/Name+Extension.swift
```diff
@@ -39,12 +41,16 @@ extension Name {
...
+		return standardizedFamilyName?.groupingComponents() ?? []
 	}
 	
-	var reversedStandardizedName: String {
-		var standardizedFamilyName = self.standardizedFamilyName
-		standardizedFamilyName += "<<"
+	var reversedStandardizedName: String? {
+		guard var standardizedFamilyName = self.standardizedFamilyName else {
+			return nil
+		}
```
naveeddotio, line 50:
```suggestion
```

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Model/DigitalCovidCertificate.swift
```diff
@@ -294,13 +294,13 @@ public struct Name: Codable, Equatable {
 
     public let familyName: String?
     public let givenName: String?
-    public let standardizedFamilyName: String
+    public let standardizedFamilyName: String?
```
naveeddotio, line 297: Is it really optional? as it should contain either of the component fnt or gnt

naveeddotio (approved):

puneetmahali1 (approved):

src/xcode/ENA/HealthCertificateToolkit/Sources/HealthCertificateToolkit/CertificateAccess/Model/DigitalCovidCertificate.swift
30mar: Yes if we have only a given name and no family name this is a valid case so the given name should be assigned and the family name should be nil, this is why the tech spec says both technically can be nil, but logically we have to check that only one of them is nil

PR #4713, opened 2022-08-08
Fix/13213 Update DCCWalletInfo in foreground

puneetmahali1:
## Description
Certificate was not updating when the app comes in foreground from the background.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13213

## Screenshots
https://user-images.githubusercontent.com/72390277/183383328-a5389961-ec21-4685-8366-28f150995e78.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
@@ -207,7 +207,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		updateExposureState(state)
 		Analytics.triggerAnalyticsSubmission()
 		appUpdateChecker.checkAppVersionDialog(for: window?.rootViewController)
-		healthCertificateService.updateDCCWalletInfosIfNeeded()
+		healthCertificateService.updateValidityStatesAndNotifications(completion: { })
 	}
 	
 	func applicationWillTerminate(_ application: UIApplication) {


naveeddotio (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -207,7 +207,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 		updateExposureState(state)
 		Analytics.triggerAnalyticsSubmission()
 		appUpdateChecker.checkAppVersionDialog(for: window?.rootViewController)
-		healthCertificateService.updateDCCWalletInfosIfNeeded()
+		healthCertificateService.updateValidityStatesAndNotifications(completion: { })
```
flxschmidt, line 210: That technically correct, for sure. :)
However I would prefer an optional `completion` argument that can accept `nil`, instead an empty closure. ;)

puneetmahali1: Yes but previously it's defined like this and they call the empty closure in various point.
If I used nil then it will shows the error: "'nil' is not compatible with expected argument type '() -> Void'"

naveeddotio: If you make the completion argument optional, it will accept nil too.

flxschmidt: Okay, if empty closure were fine before, we're gonna keep that on your code.

flxschmidt (approved):

PR #4726, opened 2022-08-12
Fix/10070: Fix Xcode compiler warnings.

flxschmidt:
## Description
Fix or silence Xcode compiler warnings.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10070

## Screenshots
Screen Shot 2022-08-12 at 15 23 01

### DIFF
--- a/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
+++ b/src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
@@ -172,7 +172,7 @@ final class ENAExposureManager: NSObject, ExposureManager {
 		self.manager = manager
 		super.init()
 
-		// swiftlint:disable:next unavailable_condition
+		// swiftlint:disable:next superfluous_disable_command
 		if #available(iOS 13.5, *) {
 			// Do nothing since we can use BGTask in this case.
 		} else if NSClassFromString("ENManager") != nil {	// Make sure that ENManager is available.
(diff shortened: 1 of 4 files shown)


puneetmahali1 (approved):

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -171,6 +171,8 @@ final class ENAExposureManager: NSObject, ExposureManager {
 	) {
 		self.manager = manager
 		super.init()
+
+		// swiftlint:disable:next superfluous_disable_command
 		if #available(iOS 13.5, *) {
```
naveeddotio, line 176: Can this be refactored with a guard like above?

flxschmidt: That's a good question, and I had the same to me. Yes I think so. 
But then comes the compiler flag for UI testing before `guard` statement or twice (after guard and in `guard-else` block).

Do you know wether we could use `#unavailable(::)` condition in the future?

```swift
if #available(iOS 13.5, *) {
   // Do nothing since we can use BGTask in this case.
} else if NSClassFromString("ENManager") != nil { ... }

#if DEBUG
if isUITesting { self.enable { _ in } }
#endif
```

naveeddotio: Unfortunately as of now we can't use #unavailable preprocessor directive.

naveeddotio (approved):

PR #4731, opened 2022-08-17
fix/10429 Certificates tab formatting issue

puneetmahali1:
## Description
Formatting issue on tab "Certificates"

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10429

## Screenshots
10429-iPhone 12 Simulator

10429-iPhone Xs


naveeddotio (changes requested):
This bug is regarding the line wrapping and not about the height of the caption count.

puneetmahali1:
> This bug is regarding the line wrapping and not about the height of the caption count.

Yes but that was fixed in January I guess and after that this ticket has reopened because of the small badge icon or let say caption count.

naveeddotio:
> > This bug is regarding the line wrapping and not about the height of the caption count.
> 
> Yes, you are right @naveeddotio but that was fixed in January this year and after that this ticket has reopened because of the small badge icon or let say caption count. Therefore, height of the caption count allow to fix that.
> 
> Please see the last comment into the ticket for more clarification.

Okay, thanks for the clarification, don't you think that it could be a little shorter?

puneetmahali1:
> > > This bug is regarding the line wrapping and not about the height of the caption count.
> > 
> > 
> > Yes, you are right @naveeddotio but that was fixed in January this year and after that this ticket has reopened because of the small badge icon or let say caption count. Therefore, height of the caption count allow to fix that.
> > Please see the last comment into the ticket for more clarification.
> 
> Okay, thanks for the clarification, don't you think that it could be a little shorter?

You are welcome! I have make it shorter now, It looks fine now.

naveeddotio (approved):

30mar (approved):

PR #4741, opened 2022-08-23
Fix/12222 wrong badge count on status tab

puneetmahali1:
## Description
Badge Count "1" on "Status" tab does not increment correctly when replacing tests

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12222

## Screenshots
https://user-images.githubusercontent.com/72390277/186169769-f862ccc9-ab6a-4d35-baa3-43dfaf3e3ce4.MP4


src/xcode/ENA/ENA/Source/Scenes/Home/HomeBadgeWrapper.swift
```diff
@@ -46,6 +46,15 @@ class HomeBadgeWrapper {
...
 	}
+	
+	func decrease(_ badgeType: BadgeType, by value: Int) {
+		let oldValue = badgesCount[badgeType] ?? 0
+		guard oldValue ?? 0 - value > 0 else {
+			return
+		}
+		badgesCount[badgeType] = (oldValue ?? 0) - value
+		saveAndUpdate()
+	}
```
flxschmidt, line 57: An alternate approach to avoid those redundant `??` fallbacks could be:

```swift
func decrease(_ badgeType: BadgeType, by value: Int) {
	let oldValue = badgesCount[badgeType]?.map { $0 } ?? 0
	guard oldValue - value > 0 else { return }

	badgesCount[badgeType] = oldValue - value
	saveAndUpdate()
}
```

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
```diff
@@ -554,8 +554,12 @@ class CoronaTestService: CoronaTestServiceProviding {
 		switch coronaTestType {
 		case .pcr:
 			pcrTest.value = nil
+			// updating badge count for home tab
```
flxschmidt, line 557: What do you think about to remove the comments, and instead we rename the `badgeWrapper` property?

```swift
// Old
private let badgeWrapper: HomeBadgeWrapper

// New
private let homeBadgeWrapper: HomeBadgeWrapper
```

flxschmidt (commented):
LGTM! some minor additional thoughts :)

src/xcode/ENA/ENA/Source/Services/CoronaTestService/CoronaTestService.swift
puneetmahali1: I was also thinking but the property was declared previously and need to refactored a lot then because it's used in several classes.

flxschmidt: ok then keep you comments

flxschmidt (approved):

naveeddotio (approved):

PR #4774, opened 2022-09-15
Fix build for Xcode14

naveeddotio:
## Description
Small fix to support building the application for Xcode14


puneetmahali1 (approved):

src/xcode/ENA/ENA/Source/Scenes/App/AppNavigationController.swift
```diff
@@ -8,17 +8,25 @@ import UIKit
...
+	private(set) lazy var _defaultScrollEdgeAppearance: Any? = nil
 
+	@available(iOS 13.0, *)
+	fileprivate var defaultScrollEdgeAppearance: UINavigationBarAppearance {
+		if _defaultScrollEdgeAppearance == nil {
+			_defaultScrollEdgeAppearance = UINavigationBarAppearance()
+		}
+		// swiftlint:disable:next force_cast
+		return _defaultScrollEdgeAppearance as! UINavigationBarAppearance
+	}
```
flxschmidt, line 20: I completely agree with that, cause I see no other solution at the moment. Would be nice, when there is a better solution without `Any` and `Force Cast` next time ...

flxschmidt (approved):

PR #4779, opened 2022-09-19
fix/13923 No mask feature by default on version update

naveeddotio:
## Description
Fixing the issue of no mask feature on version update

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13923


flxschmidt (approved):
LIKE!

puneetmahali1 (approved):

Issue #1393, opened 2020-10-22
Code comments link to non-public repository


ir-fuel:
## Avoid duplicates
* [x] Question is not already answered in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Question has not already been asked in another issue

## Your Question

* Source File:
[`SymptomsOnset.swift`](https://github.com/corona-warn-app/cwa-app-ios/blob/a1b47b3e054110e68f2cf4e7ad3135fb2041c842/src/xcode/ENA/ENA/Source/Models/SymptomsOnset.swift#L62)
* Line(s):
62
* Question:
The link

https://github.com/corona-warn-app/cwa-app-tech-spec/blob/56521167b688f418127adde09a18a48f262af382/docs/spec/days-since-onset-of-symptoms.md

Points to `cwa-app-tech-spec` which is not public

---
Internal Tracking ID: EXPOSUREAPP-3428

heinezen:
Hey @ir-fuel ,

Thanks for the feedback. I've reported this to Jira with ticket ID EXPOSUREAPP-3428. It will now be picked up by a developer and be fixed.

Regards,
CH

----
Corona-Warn-App Open Source Team

heinezen:
According to the development team, this is not something that will be changed.

----
Corona-Warn-App Open Source Team

PR #4795, opened 2022-09-22
Fix/3428 Remove non public repo link

naveeddotio:
## Description
Removing non-public repo link from comment

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-3428


Ein-Tim:
In the end this will fix https://github.com/corona-warn-app/cwa-app-ios/issues/1393, however, if referencing internal repos/documents helps the team, I don't think that is an all too big issue.

flxschmidt (approved):

puneetmahali1 (approved):

PR #4801, opened 2022-09-27
Fix/13907: Mask Badge in Person Overview should be aligned with Toggle

flxschmidt:
## Description
Show mask status badge, admission status badge and preferred person switch toggle right aligned equally.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13907

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/PreferredPerson/PreferredPersonTableViewCell.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/PreferredPerson/PreferredPersonTableViewCell.swift
@@ -108,6 +108,7 @@ class PreferredPersonTableViewCell: UITableViewCell, ReuseIdentifierProviding {
 		preferredPersonSwitch.setContentHuggingPriority(.required, for: .horizontal)
 		preferredPersonSwitch.addTarget(self, action: #selector(didTogglePreferredPersonSwitch(sender:)), for: .valueChanged)
 		topContentStackView.addArrangedSubview(preferredPersonSwitch)
+		preferredPersonSwitch.transform = CGAffineTransform(translationX: -2, y: 0)
 
 		descriptionLabel.numberOfLines = 0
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertifiedPerson/Cells/PreferredPerson/PreferredPersonTableViewCell.swift
```diff
@@ -108,6 +108,7 @@ class PreferredPersonTableViewCell: UITableViewCell, ReuseIdentifierProviding {
 		preferredPersonSwitch.setContentHuggingPriority(.required, for: .horizontal)
 		preferredPersonSwitch.addTarget(self, action: #selector(didTogglePreferredPersonSwitch(sender:)), for: .valueChanged)
 		topContentStackView.addArrangedSubview(preferredPersonSwitch)
+		preferredPersonSwitch.transform = CGAffineTransform(translationX: -2, y: 0)
```
naveeddotio, line 111: Why don't we use constraints?

flxschmidt: Because of 3 Reasons:
1. Our layout definition is correct, but there is that minor displacement from the `UISwitch` as part of the `UIStackView`. It comes from auto layout calculation. 
2. As arranged subview of a stackview, there two way to fix it by constraints: wrap `UISwitch` into `UIView`, and declare a `.trailing` padding, or change `UIStackView` trailing constraint. Especially the last approach is not a good way, because the constraints are compliant with Figma.
4. Doesn't matter, but interesting: as I remember, `CGAffineTransform` calculations comes from GPU, constraints from CPU. Thatswhy you should use `CGAffineTransforms` for movements, instead of updating constraint constants. It works often, if calculations are not intensive, but it's not the best way.

naveeddotio (approved):

30mar (approved):

puneetmahali1 (approved):

PR #4809, opened 2022-09-29
Fix/6329: "Neue Funktionen" doesn't scroll back when you left the screen and open it again (in App Informationen)

flxschmidt:
## Description
This bug fix shows a fresh instance of `DeltaOnboardingNewVersionFeaturesViewController`, to fix the scroll behaviour issue.

For Reviewers:
This bug targets a common question, that some other developers also have. Unfortunately without a good solution.
While fixing, I found two approaches for that. 
I've documented both here: https://stackoverflow.com/a/73896786/7007818 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6329

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationViewController.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationViewController.swift
@@ -128,7 +128,6 @@ class AppInformationViewController: DynamicTableViewController, NavigationBarOpa
 		tableView.deselectRow(at: indexPath, animated: true)
 		
 		if let category = Category(rawValue: indexPath.row) {
-			
 			if category == .versionInfo {
 				execute(action: versionInfoAppInformationCellModel.action)
 			} else {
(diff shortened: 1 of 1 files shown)


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/AppInformation/AppInformationViewController.swift
```diff
@@ -126,10 +126,16 @@ class AppInformationViewController: DynamicTableViewController, NavigationBarOpa
 
 	override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
 		tableView.deselectRow(at: indexPath, animated: true)
-
-		if let category = Category(rawValue: indexPath.row),
-			let action = model[category]?.action {
-			self.execute(action: action)
+		
+		if let category = Category(rawValue: indexPath.row) {
+
```
naveeddotio, line 131:
```suggestion
```

puneetmahali1 (approved):

PR #4817, opened 2022-10-05
Fix/9412 date has wrong format on IP-History

puneetmahali1:
## Description
date has wrong format on IP-History

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9412

## Screenshots
IMG_8B858B783275-1

### DIFF
--- a/src/xcode/ENA/ENA/Source/Utils/ENAFormatter.swift
+++ b/src/xcode/ENA/ENA/Source/Utils/ENAFormatter.swift
@@ -17,6 +17,7 @@ enum ENAFormatter {
 		let dateFormatter = DateFormatter()
 		dateFormatter.dateStyle = .long
 		dateFormatter.timeStyle = .short
+		dateFormatter.dateFormat = "yyyy.MM.dd HH:mm"
 		let target = Locale.autoupdatingCurrent
 		dateFormatter.locale = target
 		


naveeddotio (approved):

src/xcode/ENA/ENA/Source/Utils/ENAFormatter.swift
```diff
@@ -17,6 +17,7 @@ enum ENAFormatter {
 		let dateFormatter = DateFormatter()
 		dateFormatter.dateStyle = .long
 		dateFormatter.timeStyle = .short
+		dateFormatter.dateFormat = "yyyy.MM.dd HH:mm"
```
naveeddotio, line 20: It will also affect other places, perhaps create a new one and use it.

naveeddotio (approved):

naveeddotio (changes requested):
Small comment

src/xcode/ENA/ENA/Source/Utils/ENAFormatter.swift
puneetmahali1: Yes, I was also thinking the same but this used only this place and should not affect others.

naveeddotio (approved):

flxschmidt (approved):

PR #4850, opened 2022-10-19
Fix/14149 Crash when scanning a redeemed code

30mar:
## Description
fix the crash when scanning a redeemed code, we shouldn't save the expired tests on the app even temporarily this causes a confusion in the number of rows, because there should be one cell for the risk state "in case of the redeemed test" but the app thinks there should be 2 "risk cell and test result cell", the introduced solution will prevent the app from adding this extra cell to the view model and hence prevent the crash 

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14149

## Screenshots
https://user-images.githubusercontent.com/15270737/196716875-2fcdfcf8-2c73-4f16-8ac2-60409f8eaf49.MP4


src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -216,7 +216,7 @@ class HomeTableViewModel {
 			riskAndTestResultsRows.append(.risk)
 		}
 
-		if let pcrTest = coronaTestService.pcrTest.value {
+		if  let pcrTest = coronaTestService.pcrTest.value, pcrTest.finalTestResultReceivedDate != nil {
```
naveeddotio, line 219:
```suggestion
		if let pcrTest = coronaTestService.pcrTest.value, pcrTest.finalTestResultReceivedDate != nil {
```

naveeddotio (approved):

flxschmidt (approved):

puneetmahali1 (approved):

PR #4855, opened 2022-10-24
Admin/SPM Update latest package versions.

flxschmidt:
## Description
SPM Update latest package versions.


puneetmahali1 (approved):

30mar (approved):

naveeddotio (changes requested):
We can not update without checking for side effects

PR #4866, opened 2022-10-27
Fix/14149 app crash scanning redeemed code

naveeddotio:
## Description
Fixing the crash when scanning a redeemed code, we shouldn't save the expired tests on the app even temporarily this causes a confusion in the number of rows, because there should be one cell for the risk state "in case of the redeemed test" but the app thinks there should be 2 "risk cell and test result cell", the introduced solution will prevent the app from adding this extra cell to the view model and hence prevent the crash

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14149


naveeddotio:
Not an appropriate solution

PR #4885, opened 2022-11-11
Protocol Buffer update for release 2.29.x at Fri Nov 11 12:45:23 UTC 2022

Corona-Warn-App-Technical-User:


mlenkeit:
FYI, this PR does not contain any changes that are relevant for iOS. Anyway, I would suggest to merge it anyway to keep future ProtoBuf replications in sync.

PR #4904, opened 2022-11-30
Fix/SRS submitting works own SRS model type (with string)

flxschmidt:
## Description
Fixes a decoding problem with expiration Date

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
@@ -241,7 +241,9 @@ final class OTPService: OTPServiceProviding {
 
 			switch result {
 			case .success(let otpForSRSResponsePropertiesReceiveModel):
-				guard let expirationDate = otpForSRSResponsePropertiesReceiveModel.expirationDate else {
+				guard let expirationDateString = otpForSRSResponsePropertiesReceiveModel.expirationDate,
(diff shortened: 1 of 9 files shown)


30mar (approved):

puneetmahali1 (approved):

src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
```diff
@@ -240,12 +240,17 @@ final class OTPService: OTPServiceProviding {
...
+					completion(.failure(.restServiceError(.receivedResourceError(.invalidResponseError))))
+					return
+				}
+				
 				// Success: We store the authorized otp with timestamp and return the token.
 				let verifiedToken = OTPToken(
 					token: otp,
 					timestamp: Date(),
-					expirationDate: expirationDate.expirationDate
+					expirationDate: ISO8601DateFormatter.justUTCDateFormatter.date(from: expirationDate)
```
naveeddotio: It could potentially crash if the formatter will not be able to process the string

naveeddotio (changes requested):
See comment

src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
flxschmidt: good point, is guarded

naveeddotio (approved):

PR #4910, opened 2022-12-02
Story/14157 Extend plausible deniability for SRS and PPA

naveeddotio:
## Description
Extend plausible deniability for SRS and PPA

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14157

### DIFF
--- a/src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
@@ -31,7 +31,7 @@ struct OTPAuthorizationForSRSResource: Resource {
 			}
 		)
 		
-		self.locator = .authorizeOtpSrs(isFake: false)
+		self.locator = .authorizeOtpSrs(isFake: isFake)
 		self.type = .default
 		self.receiveResource = JSONReceiveResource<OTPResponsePropertiesReceiveModel>()
 		self.trustEvaluation = trustEvaluation
--- a/src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
+++ b/src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
@@ -10,6 +10,7 @@ struct OTPAuthorizationForSRSResource: Resource {
 	
 	init(
 		otpSRS: String,
+		isFake: Bool = false,
 		ppacToken: PPACToken,
 		trustEvaluation: TrustEvaluating = DefaultTrustEvaluation(
 			publicKeyHash: Environments().currentEnvironment().pinningKeyHashData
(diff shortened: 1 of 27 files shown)


src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
```diff
@@ -22,14 +24,20 @@ struct OTPAuthorizationForSRSResource: Resource {
 		let payload = SAP_Internal_Ppdd_SRSOneTimePassword.with {
 			$0.otp = otpSRS
 		}
+		
 		self.sendResource = ProtobufSendResource(
 			SAP_Internal_Ppdd_SRSOneTimePasswordRequestIOS.with {
-				$0.payload = payload
+				if isFake {
+					// we set the requestPadding only in case of fakeService
```
30mar, line 31: 💯

src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
```diff
@@ -116,6 +120,8 @@ final class OTPService: OTPServiceProviding {
 		   expirationDate > Date(),
 		   store.otpElsAuthorizationDate == nil {
```
30mar:
```suggestion
		   store.otpSrsAuthorizationDate == nil {
```

src/xcode/ENA/ENA/Source/Services/OTP/OTPService.swift
```diff
@@ -116,6 +120,8 @@ final class OTPService: OTPServiceProviding {
...
 			Log.info("Existing OTP ELS was not consumed before and can be used for submission.", log: .otp)
```
30mar:
```suggestion
			Log.info("Existing OTP SRS was not consumed before and can be used for submission.", log: .otp)
```

30mar (changes requested):
The PR looks good, great Job! I noticed a bug outside your code, can you check my comments and update it?

30mar (approved):

src/xcode/ENA/ENA/Source/HTTPClient/Resources/OTPAuthorizationForSRS/OTPAuthorizationForSRSResource.swift
```diff
@@ -22,14 +24,20 @@ struct OTPAuthorizationForSRSResource: Resource {
...
+					$0.requestPadding = requestPadding ?? Data()
+				} else {
+					$0.payload = payload
```
nickguendling, line 34: Looks to me like you're setting a value for `otpSRS`, but it's not used as the payload is not added if `isFake` is true. Not sure what's the right behavior here, if the payload should always be set or the `otpSRS` string is superfluous in the fake scenario

naveeddotio: @nickguendling I confirmed with Max as well, it is the expected behaviour `otpSrs` is irrelevant in case of fakeSubmission. I would like to still passed the value as I would not like to change the type to optional just for fakeSubmission. What do you think?

puneetmahali1 (approved):
👍

PR #4914, opened 2022-12-05
Fix/7109 add alert view on error log

puneetmahali1:
## Description
Add Alert View on error log when the user would like to stop analysing.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7109

## Screenshots
stop analysing


KathrinLuecke (approved):
UA approved

30mar (approved):

PR #4935, opened 2022-12-13
Fix/10243 layout of warn on behalf

puneetmahali1:
## Description
on extreme a11y settings the layout of the page "Warn on behalf of others" is broken. The descriptive text at the top can grow so much that was push the table of QR codes below the visible screen.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10243

## Screenshots
https://user-images.githubusercontent.com/72390277/207376985-46da2b46-6ab9-403f-bf8a-aec1c7316f11.MP4


src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
```diff
@@ -42,16 +42,23 @@ class DynamicTypeTableViewCell: UITableViewCell, DynamicTableViewTextCell {
 	func configureDynamicType(size: CGFloat = 17, weight: UIFont.Weight = .regular, style: UIFont.TextStyle = .body) {
 		contentTextLabel.font = UIFont.preferredFont(forTextStyle: style).scaledFont(size: size, weight: weight)
 		contentTextLabel.adjustsFontForContentSizeCategory = true
-		contentTextLabel.numberOfLines = 0
+		contentTextLabel.lineBreakMode = .byTruncatingTail
+		contentTextLabel.numberOfLines = 4
```
flxschmidt: That's not the best place to change the properties, because `DynamicTypeTableViewCell.swift` is a common `class.` 
You want a specific behaviour ;) 
So, if the changes solves your problem, maybe the changes aren't fit to other `UITableView`s that use that `class` behaviour.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
```diff
@@ -42,16 +42,23 @@ class DynamicTypeTableViewCell: UITableViewCell, DynamicTableViewTextCell {
...
 	}
 
 	func configure(text: String, color: UIColor? = nil) {
 		contentTextLabel.text = text
 		contentTextLabel.textColor = color ?? .enaColor(for: .textPrimary1)
+		
+		let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(labelTapped))
+		contentTextLabel.addGestureRecognizer(tapGestureRecognizer)
+		contentTextLabel.isUserInteractionEnabled = true
+		contentTextLabel.isAccessibilityElement = true
```
flxschmidt, line 56: That `internal func` is declared to configure only a text and a color. 

To configure a special gesture + handler, you should better provide a separate configure function for that. 

Of if the gesture behaviour shall be generic / common valid, then use another private place in the class to configure it. :)

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
```diff
@@ -88,4 +95,10 @@ class DynamicTypeTableViewCell: UITableViewCell, DynamicTableViewTextCell {
 		contentView.layoutMargins = UIEdgeInsets(top: 8, left: 16, bottom: 8, right: 16)
 		contentView.insetsLayoutMarginsFromSafeArea = false
 	}
+	
+	@objc
+	private func labelTapped() {
+		contentTextLabel.adjustsFontSizeToFitWidth = true
```
flxschmidt, line 102: That property should alway be `true` on `UILabel`s, when we want to use Dynamic Type. Independent from a gesture handler.

flxschmidt (changes requested):

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
```diff
@@ -42,16 +42,23 @@ class DynamicTypeTableViewCell: UITableViewCell, DynamicTableViewTextCell {
...
 
 	func configure(text: String, color: UIColor? = nil) {
 		contentTextLabel.text = text
 		contentTextLabel.textColor = color ?? .enaColor(for: .textPrimary1)
+		
+		let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(labelTapped))
+		contentTextLabel.addGestureRecognizer(tapGestureRecognizer)
+		contentTextLabel.isUserInteractionEnabled = true
+		contentTextLabel.isAccessibilityElement = true
 	}
```
flxschmidt: That `internal func` is declared to configure only a text and a color. 

To configure a special gesture + handler, you should better provide a separate configure function for that. 

Of if the gesture behaviour shall be generic / common valid, then use another private place in the class to configure it. :)

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
puneetmahali1: fixed.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
puneetmahali1: Yes, I am also assuming the same but unfortunately without this property it doesn't work.

src/xcode/ENA/ENA/Source/Scenes/DynamicTableViewController/Views/DynamicTypeTableViewCell.swift
puneetmahali1: Wrap into a separate func()

naveeddotio (approved):

flxschmidt (approved):
🍓

PR #4951, opened 2022-12-19
Fix the overlap between submission cells in contact diary

30mar:
## Description
fix the overlap between submission cells

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14479

## Screenshots
before
Screenshot 2022-12-16 at 12 03 09
after
Screenshot 2022-12-19 at 06 58 03


Ein-Tim:
May I ask in which situation two entries are shown? If the user warns twice on the same day?

naveeddotio:
> May I ask in which situation two entries are shown? If the user warns twice on the same day?

Edge case, only relevant for testers. Should not happen in ideal case.

naveeddotio (approved):

puneetmahali1 (approved):

30mar:
> May I ask in which situation two entries are shown? If the user warns twice on the same day?

@Ein-Tim exactly like what Naveed said, normally we have a condition that a certain time has to pass between submissions and more that one submission in the same day is logically not possible but we can disable these checks sometimes for testing and this is where we found the bug

PR #4991, opened 2023-01-10
Fix/6511 Nothing happens when you check-in at an event

puneetmahali1:
## Description
Nothing happens when you check-in at an event which has the same start and end time. Now, the default event duration is at-least for 15 minutes.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6511

## Screenshots
https://user-images.githubusercontent.com/72390277/211600085-fc0fdb77-c795-4168-884f-1fcd137c6945.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
+++ b/src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
@@ -47,11 +47,10 @@ struct TraceLocation: Equatable {
 			let eventDuration = Calendar.current.dateComponents(
 				[.minute],
 				from: startDate,
-				// The event duration should be atleast 15 minutes
-				to: endDate.addingTimeInterval(15 * 60)
+				to: endDate
 			).minute
 
-			duration = eventDuration ?? fallback
+			duration = eventDuration.map { $0 == 0 ? fallback : $0 } ?? fallback
 		} else {
 			duration = fallback
 		}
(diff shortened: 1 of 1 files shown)


src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
```diff
@@ -47,7 +47,8 @@ struct TraceLocation: Equatable {
 			let eventDuration = Calendar.current.dateComponents(
 				[.minute],
 				from: startDate,
-				to: endDate
+				// The event duration should be atleast 15 minutes
+				to: endDate.addingTimeInterval(15 * 60)
 			).minute
 
 			duration = eventDuration ?? fallback
```
flxschmidt: At the moment you'll add 15 minutes everytime for `suggestedCheckoutLengthInMinutes: Int`. I guess that's not what you would like to have. 

Rather a reaction, when `startDate` and `endDate` are the same (`eventDuration == 0`). Then the fallback would be `15` minutes.
That's also what I've been implemented with [14148](https://github.com/corona-warn-app/cwa-app-ios/pull/4956/files) (you can compare with the commit history). But it seems, the the change was overwritten.

Could you please check the App behaviour with that adoption:

```swift

let eventDuration = Calendar.current.dateComponents(
	[.minute],
	from: startDate,
	to: endDate
).minute

duration = eventDuration.map { $0 == 0 ? fallback : $0 } ?? fallback // <-- HERE
```

flxschmidt (changes requested):

src/xcode/ENA/ENA/Source/Scenes/Events/Store/Model/TraceLocation.swift
puneetmahali1: fallback adjusted

naveeddotio (approved):

flxschmidt (approved):
LGTM

PR #4972, opened 2023-01-02
 Fix/14416 Bad-behavior-of-CWA-when-WLAN-has-no-connection

30mar:
## Description
the app is loading for one min in case of a poor network when connected to WiFi, the bug tickets requires making the timeout 10 sec max

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14416

PR #5002, opened 2023-01-16
Revert " Fix/14416 Bad-behavior-of-CWA-when-WLAN-has-no-connection"

30mar:
Reverts corona-warn-app/cwa-app-ios#4972


Ein-Tim:
Could you give a reason why the changes are reverted, please?

PR #5003, opened 2023-01-17
Fix/14577 first srs submission leads to api token already issued

30mar:
## Description

- merging the 3 api tokens we had previously "EDUS, ELS and SRS" into one created this bug, the fix was a migration logic to check the store for the old values
- add dev menu feature to see all stored api tokens
- add dev menu feature to delete the new api token from the store

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14577

## Screenshots
Simulator Screen Shot - iPhone 8 - 2023-01-17 at 11 41 48

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
+++ b/src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
@@ -90,6 +90,10 @@ final class MockTestStore: Store, PPAnalyticsData {
 	var previousPpacApiTokenSrs: TimestampedToken?
 	var otpTokenSrs: OTPToken?
     var otpSrsAuthorizationDate: Date?
+	/// DEPRECATED, Only used for dev menu testing for API tokens
+	var ppacApiTokenEdus: TimestampedToken?
+	/// DEPRECATED, Only used for dev menu testing for API tokens
+	var ppacApiTokenEls: TimestampedToken?
 
 	// MARK: - PrivacyPreservingProviding
 
--- a/src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
+++ b/src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
@@ -537,6 +537,14 @@ extension SecureStore: PrivacyPreservingProviding {
 		get { kvStore["previousAPITokenPPAC"] as TimestampedToken? }
 		set { kvStore["previousAPITokenPPAC"] = newValue }
 	}
+	
+	// FOR reading legacy values for DEVMENU ONLY!, dont save any new code to these 2 variables!
+	var ppacApiTokenEdus: TimestampedToken? {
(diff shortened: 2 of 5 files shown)


puneetmahali1 (approved):

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
```diff
@@ -521,16 +521,30 @@ extension SecureStore: PrivacyPreservingProviding {
...
 		set { kvStore["previousAPITokenPPAC"] = newValue }
 	}
+	
+	// FOR reading legacy values for DEVMENU ONLY!, dont save any new code to these 2 variables!
+	var ppacApiTokenEdus: TimestampedToken? {
+		kvStore["ppacApiToken"] as TimestampedToken?
+	}
+	var ppacApiTokenEls: TimestampedToken? {
+		kvStore["ppacApiTokenEls"] as TimestampedToken?
+	}
```
flxschmidt, line 547: What do you think, to make that only public/internal for Dev-Builds?. Cause the only Getter is the Dev Menu.

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
```diff
@@ -90,6 +90,10 @@ final class MockTestStore: Store, PPAnalyticsData {
 	var previousPpacApiTokenSrs: TimestampedToken?
 	var otpTokenSrs: OTPToken?
     var otpSrsAuthorizationDate: Date?
+	/// DEPRECATED, Only used for dev menu testing for API tokens
+	var ppacApiTokenEdus: TimestampedToken?
+	/// DEPRECATED, Only used for dev menu testing for API tokens
+	var ppacApiTokenEls: TimestampedToken?
```
flxschmidt, line 96:
```suggestion
	@available(*, deprecated, message: "Only used for dev menu testing for API tokens")
	var ppacApiTokenEdus: TimestampedToken?
	@available(*, deprecated, message: "Only used for dev menu testing for API tokens")
	var ppacApiTokenEls: TimestampedToken?
```

src/xcode/ENA/ENA/Source/Workers/Store/SecureStore.swift
30mar: there is no setter for the property anyway so there is no way to over write it

src/xcode/ENA/ENA/Source/Services/__tests__/Mocks/MockTestStore.swift
30mar: but is can still be accessed after using this ?

flxschmidt: Yes, that should leave a warning message. But it's not important, you can also keep it as it is.

flxschmidt (approved):

PR #5019, opened 2023-01-24
Fix/14578 no internet prevents clicking on srs

puneetmahali1:
## Description
No internet connections prevents clicking on Self test submission

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14578

## Screenshots
https://user-images.githubusercontent.com/72390277/214326263-249aa7be-f80d-4239-b4e2-1dd7447175d0.MP4

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
+++ b/src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
@@ -31,9 +31,10 @@ final class CachedAppConfiguration: AppConfigurationProviding {
 	// MARK: Protocol AppConfigurationProviding
 
 	func appConfiguration(forceFetch: Bool = false) -> AnyPublisher<SAP_Internal_V2_ApplicationConfigurationIOS, Never> {
+		let isOnline = InternetConnectivityMonitor.shared.isDeviceOnline
 		let force = shouldFetch() || forceFetch
 
-		if let cachedVersion = store.appConfigMetadata?.appConfig, !force {
+		if let cachedVersion = store.appConfigMetadata?.appConfig, (!force || !isOnline) {
 			Log.debug("fetching cached app configuration", log: .appConfig)
 			// use the cached version
 			return Just(cachedVersion)


Ein-Tim:
So IIUC when you are not connected to the internet and click on the tile nothing happens.  Like nothing at all? I don't think that's a good idea! Some sort of error should be shown to the user, otherwise this could lead to very confused users.

thomasaugsten:
@Ein-Tim This is under certain circumstances the situation with 3.0 and this PR will fix the issue.

Ein-Tim:
@thomasaugsten So there will actually be some sort of UI information presented to the user why this doesn't work?

thomasaugsten:
Depends if he never has an internet connection he will receive the error that we cannot check his phone time.
If its a temporary no internet he will receive an no internet error when he tries to submit

30mar (approved):

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
```diff
@@ -31,9 +31,10 @@ final class CachedAppConfiguration: AppConfigurationProviding {
 	// MARK: Protocol AppConfigurationProviding
 
 	func appConfiguration(forceFetch: Bool = false) -> AnyPublisher<SAP_Internal_V2_ApplicationConfigurationIOS, Never> {
+		let isOnline = InternetConnectivityMonitor.shared.isDeviceOnline
```
30mar, line 34: aren't we doing a shared class now as discussed?

30mar (changes requested):
aren't we doing a shared class for checking the network now as discussed?

src/xcode/ENA/ENA/Source/Services/Risk/Provider/Helper/CachedAppConfiguration.swift
puneetmahali1: InternetConnectivityMonitor is the shared class which I got from Felix after merged his reopened PR.

30mar (approved):

naveeddotio (approved):

PR #5047, opened 2023-02-13
Fix/10829 Duration Picker Auto Scroll

puneetmahali1:
## Description
Auto Scroll was missing on header tap on event duration picker and increase the auto scroll length

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10829

## Screenshots
https://user-images.githubusercontent.com/72390277/218560809-c7a2c1e1-91ff-49ba-942f-911037bce6f4.mov

## Recording for Bigger Screen
https://user-images.githubusercontent.com/72390277/219011808-e6c22123-4036-4986-a7f4-5daac84d0949.mov


naveeddotio:
Hey @puneetmahali1 since there are many hard coded values, could you also verify it for bigger sized devices and share the recording too? Thanks

puneetmahali1:
> Hey @puneetmahali1 since there are many hard coded values, could you also verify it for bigger sized devices and share the recording too? Thanks

@naveeddotio: Sure, I have attached the bigger screen recording(iPhone 14 Pro max) there we don't need the auto scroll as we can see the picker view without it.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/Events/TraceLocations/Configuration/TraceLocationConfigurationViewController.swift
```diff
@@ -229,12 +229,32 @@ class TraceLocationConfigurationViewController: UIViewController, FooterViewHand
 	private func temporaryDefaultLengthHeaderTapped() {
 		viewModel.temporaryDefaultLengthHeaderTapped()
 		view.endEditing(true)
+		
+		// If iPhone resolution height is less than or equal to 844 then the screen will auto scroll to see the event duration picker
+		// if the height is more than 844 then user can see the event duration picker without auto scroll.
+		// For screen resolution please see more details on: https://www.ios-resolution.com/
+		if !viewModel.temporaryDefaultLengthPickerIsHidden && UIScreen.main.bounds.size.height <= 667 {
```
flxschmidt, line 236: Your solution is fixing the bug.

From the code base, I would prefer a common Helper class for device bounds (device resolutions).
Similar to that https://github.com/anatoliyv/AssistantKit

But at the moment your solution is fine for me

flxschmidt (approved):

PR #5051, opened 2023-02-17
Task/14811: Preparation - global variable

flxschmidt:
## Description
Determine if the CWA hibernation state shall be given.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14811

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/Globals.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/Globals.swift
@@ -19,32 +19,6 @@ var isScreenshotMode: Bool {
 
 #endif
 
-/// Determines the CWA Hibernation State, beginning on 01.06.2023.
-var isHibernationState: Bool {
-	#if RELEASE
-
-	var hibernationStartDateComponents = DateComponents()
-	hibernationStartDateComponents.year = 2023
-	hibernationStartDateComponents.month = 6
-	hibernationStartDateComponents.day = 1
-	hibernationStartDateComponents.hour = 0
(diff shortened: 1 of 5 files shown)


30mar (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/Globals.swift
```diff
@@ -19,6 +19,32 @@ var isScreenshotMode: Bool {
...
+	hibernationStartDateComponents.year = 2023
+	hibernationStartDateComponents.month = 6
+	hibernationStartDateComponents.day = 1
+	hibernationStartDateComponents.hour = 0
+	hibernationStartDateComponents.minute = 0
+	hibernationStartDateComponents.second = 0
+	hibernationStartDateComponents.timeZone = .utcTimeZone
+	
+	guard let hibernationStartDate = Calendar.current.date(from: hibernationStartDateComponents) else {
+		fatalError("The hibernation start date couldn't be created.")
```
Ein-Tim: Will this fatalError lead the app to crash if the hibernation start date couldn't be created?

flxschmidt: Yes, because that is just a theoretically `guard` `else` case. The `date(from:)` method from `Calendar` returns an `optional` value. The `DateComponents` above are correct, so that case will never be happen.

Ein-Tim: Ok, sounds good! Thank you, Felix!

naveeddotio (approved):

PR #5061, opened 2023-02-23
Task/14819 adapt changes to background tasks

flxschmidt:
## Description
Don't register / schedule background tasks, when CWA Hibernation is active.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14819


src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
```diff
@@ -186,7 +186,13 @@ final class ENAExposureManager: NSObject, ExposureManager {
 							Log.error("AppDelegate is not of Type AppDelegate, this should not happen!", log: .background)
 							return
 						}
-						Log.info("Starting backgroundTask via Bluetooth", log: .background)
+						
+						guard !CWAHibernationProvider.shared.isHibernationState else {
```
naveeddotio: Do we really need this code? 

We are already calling appDelegate.taskExecutionDelegate.executeENABackgroundTask below and you have the same code there.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
```diff
@@ -65,6 +65,12 @@ class TaskExecutionHandler: ENATaskExecutionDelegate {
 			return
 		}
 		
+		guard !CWAHibernationProvider.shared.isHibernationState else {
```
naveeddotio: Would suggest to move above the guard before, it will not make any difference but logically it make sense as this check is priority 1.

naveeddotio (changes requested):
Please see comments

src/xcode/ENA/ENA/Source/Models/Exposure/ExposureManager.swift
flxschmidt: The initial reason was that `init` and `executeENABackgroundTask` are both `internal` accessible. But you're right, after `executeENABackgroundTask` isn't any further code. So I remove the `guard` in the `init` block.

src/xcode/ENA/ENA/Source/AppDelegate & Globals/ENATaskExecutionDelegate.swift
flxschmidt: good point, agree.

src/xcode/ENA/ENA/Source/Models/Task Scheduling/ENATaskScheduler.swift
```diff
@@ -48,6 +48,11 @@ final class ENATaskScheduler {
 	// MARK: - Task registration.
 	@available(iOS 13.0, *)
 	private func registerTask(with taskIdentifier: ENATaskIdentifier, execute: @escaping ((BGTask) -> Void)) {
+		guard !CWAHibernationProvider.shared.isHibernationState else {
```
naveeddotio: Instead of having this logic in the function doesn’t it make sense that we check the state before calling the function? It will be more cleaner I believe.

flxschmidt: As long as the function is `private` -- yes, I would agree to your approach. 👍

naveeddotio (approved):

30mar (approved):

PR #5077, opened 2023-03-09
Fix/14900 dev menu hibernation not working properly

30mar:
## Description
update dev menu logic to change the start of EOL
We should be able to change the start of eol not the current date as per testers requirement

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14900

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
@@ -29,10 +29,10 @@ class CWAHibernationProvider: RequiresAppDependencies {
 			return LaunchArguments.endOfLife.isHibernationStateEnabled.boolValue
 		}
 		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
-		return secureStore.hibernationStartDate >= hibernationStartDate
+		return secureStore.hibernationStartDate >= Date()
 		#elseif !RELEASE
 		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
-		return secureStore.hibernationStartDate >= hibernationStartDate
+		return secureStore.hibernationStartDate >= Date()
 		#else
 		return Date() >= hibernationStartDate
 		#endif
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
@@ -28,7 +28,11 @@ class CWAHibernationProvider: RequiresAppDependencies {
 		if isUITesting {
 			return LaunchArguments.endOfLife.isHibernationStateEnabled.boolValue
 		}
-		return secureStore.hibernationComparisonDate >= hibernationStartDate
+		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
+		return secureStore.hibernationStartDate >= hibernationStartDate
+		#elseif !RELEASE
+		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
+		return secureStore.hibernationStartDate >= hibernationStartDate
 		#else
 		return Date() >= hibernationStartDate
 		#endif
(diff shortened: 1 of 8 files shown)


OLVI5 (approved):

src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
```diff
@@ -28,7 +28,11 @@ class CWAHibernationProvider: RequiresAppDependencies {
 		if isUITesting {
 			return LaunchArguments.endOfLife.isHibernationStateEnabled.boolValue
 		}
-		return secureStore.hibernationComparisonDate >= hibernationStartDate
+		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
+		return secureStore.hibernationStartDate >= Date()
+		#elseif !RELEASE
+		Log.debug("current hibernationStartDate \(secureStore.hibernationStartDate)")
+		return secureStore.hibernationStartDate >= Date()
```
flxschmidt, line 35: I don't understand the intension of that code. 
Example: 
Today = 09.03.2023
secureStore.hibernationStartDate = 10.03.2023

`return secureStore.hibernationStartDate >= Date() // hibernation would be true`

Hibernation would be true every time the set date in Dev Menu is later than today.
Hibernation would be false every time the set date in Dev Menu is earlier than today

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/CWAHibernationProviderTests.swift
```diff
@@ -24,7 +24,7 @@ final class CWAHibernationProviderTests: CWATestCase {
 		guard let hibernationComparisonDate = Calendar.current.date(from: beforeHibernationStartDateComponents) else {
 			return XCTFail("Expect the hibernation comparison date from the corresponding date components.")
 		}
-		mockTestStore.hibernationComparisonDate = hibernationComparisonDate
+		mockTestStore.hibernationStartDate = hibernationComparisonDate
```
flxschmidt, line 27: We should better use a separate item in the store.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
```diff
@@ -825,8 +825,6 @@ class HealthCertificateService: HealthCertificateServiceServable {
 							dispatchGroup.leave()
 						}
 					)
-				}
-				if !CWAHibernationProvider.shared.isHibernationState {
```
flxschmidt, line 829: Why you outcomment this?

src/xcode/ENA/ENA/Source/AppDelegate & Globals/__tests__/CWAHibernationProviderTests.swift
30mar: we dont need the old flag i renamed it and changed the logic

src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
30mar: yes that is what the testers are expecting

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
30mar: it makes some of the failing unit tests pass

src/xcode/ENA/ENA/Source/AppDelegate & Globals/CWAHibernationProvider.swift
flxschmidt: Then we should have this requirements in a ticket description. Nobody could comprehend the changes.
This would be also a change request and not a bug.

src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
flxschmidt: Tests have to adapt to the latest code base. Not the code base to the latest tests.

flxschmidt (approved):
> yes that is what the testers are expecting

Please still add the business logic change requests in the ticket as discussed in the chat. It's important that everybody can understand the changes and why we made. Approved.

PR #5096, opened 2023-03-15
Fix/14950 Update homescreen when app becomes active

30mar:
## Description
make sure home screen is updated by subscribing to notification didBecomeActive
Note we are now reloading the whole table view because we are refreshing almost all the sections not just one or 2 like the past.
## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14950

##Video
https://user-images.githubusercontent.com/15270737/225344504-94bb80f8-bd97-49a6-acc8-63c1ff5f0472.MP4


src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -102,28 +102,6 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 				)
 			}
 			.store(in: &subscriptions)
-		
-		viewModel.cclService.shouldShowNoticeTile
```
flxschmidt, line 106: With this deletion we would ignore the CCL config for App Closure Notice. I know it makes sense, because we show either the notice or the thank-you tile. But is it aligned with the Tech Spec?

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -1134,6 +1112,12 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 	@objc
 	private func onDidBecomeActiveNotification() {
 		setupRightBarButtonItem()
+		
+		let isHibernationState = CWAHibernationProvider.shared.isHibernationState
+		viewModel.isHibernationState = isHibernationState
```
flxschmidt:
```suggestion
		viewModel.isHibernationState = CWAHibernationProvider.shared.isHibernationState
```

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -1134,6 +1112,12 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
...
+		DispatchQueue.main.async {
+			self.tableView.reloadData()
+		}
```
flxschmidt:
```suggestion
		DispatchQueue.main.async { [weak self] in
			self?.tableView.reloadData()
		}
```

src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewModel.swift
```diff
@@ -366,7 +365,7 @@ class HomeTableViewModel {
 			return 1
 		}
 		#endif
-		if shouldShowAppClosureNotice, !isHibernationState {
```
flxschmidt: As mentioned above, that's a static decision, without CCL config.

flxschmidt (changes requested):
Good approach! let's clarify the open questions. Please still add the PR Title prefix from the ticket.

PR #5112, opened 2023-03-22
Fix/14975 Set hibernation flag for the notice period

30mar:
## Description
When we return from background in EOL state we are updating 2 things separatly:
1- the notice tile  "as it depends on the ccl flag"
2- the rest of the tableView sections.
I could reproduce the crash and i think the reason is the race condition if the notice tile is trying to update before setting the viewModel.isHibernationState which is currenly only being set in point 2.
so we should set it in both points.

After adding this the crash is gone for me

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14975


flxschmidt (approved):
🍓 🍒

PR #5122, opened 2023-03-28
Fix/14992 deadman notification still comes when eol is already reached

30mar:
## Description
make sure notifications are canceled as well as other EOL events

## Link to Jira


flxschmidt (approved):
LGTM

puneetmahali1 (approved):

PR #5131, opened 2023-03-30
Fix/14981 Blocked Vaccination and Recovery Certificates stay in Status BLOCKED after RampDown

flxschmidt:
## Description
In hibernation we set the validity state of blocked health certificate to valid.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14981

## Screenshots
Screenshot 2023-03-30 at 16 16 27

### DIFF
--- a/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
+++ b/src/xcode/ENA/ENA/Source/Services/HealthCertificate/HealthCertificateService.swift
@@ -864,10 +864,8 @@ class HealthCertificateService: HealthCertificateServiceServable {
 		// Hibernation
 		guard !CWAHibernationProvider.shared.isHibernationState else {
 			
-			// In hibernation we set the validity state of any health certificate to expired or valid.
-			if Date() >= healthCertificate.expirationDate {
-				healthCertificate.validityState = .expired
-			} else {
+			// In hibernation we set the validity state of blocked health certificate to valid.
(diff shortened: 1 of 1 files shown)


Ein-Tim:
I don't fully understand, the PR title says: "Blocked Vaccination and Recovery Certificates stay in Status BLOCKED after RampDown" and the PR description says: "In hibernation we set the validity state of blocked health certificate to valid.". This is contradictory, what's actually the intended behavior?

flxschmidt:
> I don't fully understand, the PR title says: "Blocked Vaccination and Recovery Certificates stay in Status BLOCKED after RampDown" and the PR description says: "In hibernation we set the validity state of blocked health certificate to valid.". This is contradictory, what's actually the intended behavior?

You're right, that's a bit historical grown. That's the title of the bug ticket, but the latest decision is as follows:

- Certificates REVOKED before EOL should stay REVOKED as the info comes from the revocation list, and the last list known to the mobile is used to evaluate the status (in the best case the information is still available in the app).

- Certificates INVALID before EOL should stay INVALID as the info comes from the DSC list, and the last list known to the mobile is used to evaluate the status (in the best case the information is still available in the app, otherwise the fallback list should be used).

- Certificates BLOCKED before EOL are unblocked again after EOL, as this should be part of the Wallet Info, and the Wallet is reset after EOL.

30mar (approved):

Ein-Tim:
Thanks for the explanation, Felix!

PR #5140, opened 2023-04-03
Fix/14992 EOL - Remove notifications syncronously

30mar:
## Description

- When we reach EOL and try to close the app we try to cancel notifications in an async method, this wont work properly as the app will be removed from memory before the async block is reached and the notifications are canceled
- the solutions is to call `UNUserNotificationCenter.current().removeAllPendingNotificationRequests()` which will do the same but synchronously.
The downside of reverting the previous  code:
We wont be able to stop notifications in the future if the EOL is not reached yet. e.g today is 3/4/2023 Eol is 4/4/2023 and we have a notification scheduled in 5/4/2023. this notification will not be canceled unless the user tries to open, close, of move the app to background **DURING** End of life.

a redundant call was removed from `applicationDidEnterBackground()` as we will cancel anyway in `applicationWillResignActive()` which has to be called before  `applicationDidEnterBackground`, so need to cancel twice

## Link to Jira


flxschmidt (approved):
For me that's completely fine, as discussed.

PR #5145, opened 2023-04-04
Fix/14992 deadman notification still comes when eol is already reached

30mar:
## Description
- Deadman notification is scheduled "if not already scheduled" to fire in 36 hours. based on that we should cancel  scheduled deadman notifications if EOL will be reached in 36 hours or less, because there is no point of showing them in EOL.

NOTE: this is just an improvement to the proposed approached, we still have some limitations from the OS as we cant cancel we check pending notification if the app is terminated "i.e closed"
## Link to Jira


flxschmidt (approved):
For me your approach and suggestion is fine.

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/Overview/HealthCertificateOverviewViewModel.swift
```diff
@@ -27,6 +27,7 @@ class HealthCertificateOverviewViewModel {
 					.filter { !$0.healthCertificates.isEmpty }
 				self.decodingFailedHealthCertificates = $0
 					.flatMap { $0.decodingFailedHealthCertificates }
+
```
30mar:
```suggestion
```

PR #5156, opened 2023-04-12
EOL-14954-offline-onboarding-shows-popup-for-risk-calculation

30mar:
## Description
Missing return from the hibernation check makes the alert shows in some cases

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14954

## Screenshots
https://user-images.githubusercontent.com/15270737/231428733-1d681a38-00bb-4cbf-842f-205da389c76b.MP4


src/xcode/ENA/ENA/Source/Scenes/Home/HomeTableViewController.swift
```diff
@@ -923,6 +923,7 @@ class HomeTableViewController: UITableViewController, NavigationBarOpacityDelega
 	private func showBackgroundFetchAlertIfNeeded(completion: @escaping () -> Void = {}) {
 		if CWAHibernationProvider.shared.isHibernationState {
 			completion()
+			return
```
flxschmidt, line 926: As we need a return here, I would prefer a `guard.` That's the same result, but makes the code more clearly. :)

flxschmidt (approved):

PR #5169, opened 2023-05-03
Fix/15131 Quick actions are not reduced correctly in hibernation state

flxschmidt:
## Description
The quick actions "QR-Codes sannen" & "Check-ins anzeigen" should not be shown in hibernation state.

## Link to Jira
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-15131

## Screenshots
Screenshot 2023-05-03 at 11 53 32

### DIFF
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
@@ -87,7 +87,7 @@ enum QuickAction: String {
 			
 			let status = AVCaptureDevice.authorizationStatus(for: .video)
 			if status == .authorized || status == .notDetermined {
-				// dont show camera related actions if no camera access is granted
+				// don't show the camera related actions if no camera access is granted
 				shortcutItems.append(
 					contentsOf: [
 						UIApplicationShortcutItem(
--- a/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
+++ b/src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
@@ -51,21 +51,40 @@ enum QuickAction: String {
 			application.shortcutItems = nil
 			return
 		}
-
-		var shortcutItems = [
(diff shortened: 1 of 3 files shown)


src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate+QuickActions.swift
```diff
@@ -51,21 +51,60 @@ enum QuickAction: String {
...
+					type: QuickAction.showCertificates.rawValue,
+					localizedTitle: AppStrings.QuickActions.showCertificates,
+					localizedSubtitle: nil,
+					icon: UIApplicationShortcutIcon(templateImageName: "Icons_Tabbar_Certificates")
+				)
+			]
+			
+			let status = AVCaptureDevice.authorizationStatus(for: .video)
+			if status == .authorized || status == .notDetermined {
+				// dont show camera related actions if no camera access is granted
```
puneetmahali1:
```suggestion
				// don't show the camera related actions if no camera access is granted
```

puneetmahali1 (approved):
LGTM!

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertificate/HealthCertificateCoordinator.swift
```diff
@@ -108,7 +108,13 @@ final class HealthCertificateCoordinator {
 	private var validationCoordinator: HealthCertificateValidationCoordinator?
 	
 	private var infoScreenShown: Bool {
-		get { store.healthCertificateInfoScreenShown }
+		get {
```
naveeddotio, line 111: what's the usage of this change?

flxschmidt: Good question! 
The usage is, that the info screen has texts like "Add certificates" and that would be not a use case in hibernation, because we only show certificates. Thatswhy I change the getter for hibernation and skip the info screen at this point.

naveeddotio (approved):

src/xcode/ENA/ENA/Source/Scenes/HealthCertificates/HealthCertificate/HealthCertificateCoordinator.swift
Ein-Tim: I guess that's a first (and maybe the only) move reg. https://github.com/corona-warn-app/cwa-documentation/issues/950?

Issue #5168, opened 2023-05-02
App doesn't correctly deactivate the ENF in some circumstances


Ein-Tim:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for iOS only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: Not important
- iOS version: Not important
- App version: 3.2.1

## Describe the bug

Based on Twitter reports, I was able to reproduce the following problem: The CWA iOS app (3.2.1) doesn't correctly turn off the ENF, if the version 3.2.0 was installed before and was opened and immediately crashed. 
**Important:**
- The problem doesn't occur if version 3.2.0 started up correctly at least once after 00:00h on May 1, 2023
- I'm unsure if the problem occurs when version 3.2.0 wasn't opened at all, but only version 3.2.1 is opened. Please check that. 

## Steps to reproduce the issue

1. Make sure you have CWA version 3.2.0 installed and that the app is not running in the background
2. Open the app
3. Notice the crash
4. Check in the iOS settings whether the ENF was deactivated and notice that it wasn't
5. Update to version 3.2.1 of the CWA
6. Open the CWA
7.  Check in the iOS settings whether the ENF was deactivated and notice that it's still active

## Expected behaviour

The CWA should deactivate the ENF after May 1, 2023. 

---
Internal Tracking ID: [EXPOSUREAPP-15162](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-15162)

Ein-Tim:
@naveeddotio 

> I'm unsure if the problem occurs when version 3.2.0 wasn't opened at all, but only version 3.2.1 is opened. Please check that.

Just curious: Did you check that and if yes, what's the behavior?

Ein-Tim:
- PR fixing this #5171

Ein-Tim:
@dsarkar You can also apply the https://github.com/corona-warn-app/cwa-app-ios/labels/Fix%203.2.2 label to this issue.

Ein-Tim:
This was fixed with version 3.2.2.

PR #5170, opened 2023-05-03
Changing the location of disabling ENF

naveeddotio:
## Description
Change in the location of disabling ENF

https://github.com/corona-warn-app/cwa-app-ios/issues/5168


src/xcode/ENA/ENA/Source/AppDelegate & Globals/AppDelegate.swift
```diff
@@ -237,10 +241,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, CoronaWarnAppDelegate, Re
 
 	func applicationDidBecomeActive(_ application: UIApplication) {
 		Log.info("Application did become active.", log: .appLifecycle)
-		
-		if CWAHibernationProvider.shared.isHibernationState {
-			applyEndOfLifeChanges()
-		}
```
flxschmidt, line 243: Did you checked, that there will be no side effect if we remove `applyEndOfLifeChanges()` in `applicationDidBecomeActive(:)`?

naveeddotio:
Doesn't solve the issue

